Issue #235, opened 2013-11-22
Add F# Compiler Helpers


panesofglass:
I would often like to avoid the fsproj file for various reasons, including a desire to organize my files in folders as I see fit. The fsproj format has too many other things going on, and I would rather use the lighter-weight `build.fsx` to declare my build order and dependencies.

As an example, let's use the following folder and build declaration:

```
- Project A
    - Module 1
        - Module1.fs
        - Type2.fs
    - Build.fsx
    - Type1.fs
```

Inside `Project A\Build.fsx`:
```
Target "Build" (fun _ ->
    [ "Type1.fs"
      "Module1\Type2.fs"
      "Module1\Module1.fs"
    ]
    |> Project.dependencies [ "Project B" ] // Look in NuGet or other repo
    |> Fsc.clean
    |> Fsc.build // these are just API ideas and not at all intended to inform a design
)
```

I see the implementation following a progression of the following steps (or skipping unnecessary ones):

1. Host FSC in a process similar to what is done for FSI today.
2. Reference the [FSharp.Compiler.Service](https://github.com/fsharp/FSharp.Compiler.Service) to compile interactively. The long term vision of that library should allow us to move the FSI helpers to use the library reference, as well.
3. Solve issue #101 so that builds could run interactively, outside the FAKE.exe.
4. Leverage interactive builds to gen assemblies as they are requested by `Assembly.Load`.
5. Again leverage interactive builds to run them on `FileSystemWatcher` events, i.e. something similar to my [reload hack](https://gist.github.com/panesofglass/2908284).

I'm not sure if this is a fit for FAKE. I do know we will need to address these problems soon. Unfortunately, I don't have any concrete interfaces to reference other than those used in [scriptcs modules](https://github.com/scriptcs/scriptcs/tree/dev/src/ScriptCs.Hosting), as seen in [scriptcs-fsharp](https://github.com/scriptcs-contrib/scriptcs-fsharp). That's probably a good enough starting point to frame the discussion.

Again, this may not be a fit for FAKE, as FAKE doesn't currently address project structure or dependencies.

panesofglass:
@forki, I think we can close this, right?

PR #497, opened 2014-07-06
ReFake: an incremental build logic DSL

yawaramin:
My effort at creating an incremental build DSL on top of FAKE. For several reasons, as explained in the commit message, I did this by creating a new type of build 'target' that stores more information about the target and its dependencies. The implementation is as simple as possible; there are no lookups of target names to objects happening in the background. The build script writer directly creates target objects using the helper functions, and arranges them into a dependency network at the same time. They then run the topmost target in the network to lazily build whatever needs to be built.

This also allows a pretty elegant way of setting up multiple build configurations in the same script using what I call 'virtual targets'. A simple example:

    #r "/path/to/FakeLib.dll"
    open Fake.FscHelper
    open Fake.ReFake
    
    // Debug configuration is default.
    let debug = ref true
    let outputPath = ref "bin/Debug/"
    let optimize = ref "--optimize-"
    
    // A virtual target that sets up a release configuration and doesn't have any dependencies.
    let releaseConfig = vx "Release Configuration" [] (fun _ _ ->
      debug := false
      outputPath := "../../../build/"
      optimize := "--optimize+"
      0) // Need to return an exit code from build function

    // A file target that depends on 'Main.fs'.
    let Main_exe = fx' "Main.exe" [fx "Main.fs"] (fun name deps ->
      fxs deps // Gets list of names of input files.
      |> fsc (fun ps -> // 'fsc' automatically returns exit code.
        { ps with Output = !outputPath + name
                  Debug = !debug
                  OtherParams = [!optimize] }))

    // A virtual target that depends on 'Main.exe'.
    let debugBuild = vx "Debug Build" [Main_exe] nop
    // A virtual target that depends on 'releaseConfig' _and_ 'Main.exe' (list order matters).
    let releaseBuild = vx "Release Build" [releaseConfig; Main_exe] nop

    reRunAndExit debugBuild
    //reRunAndExit releaseBuild

Because the `releaseConfig` target is a dependency of the `releaseBuild` target, it will always be run before `releaseBuild`; and will thus set up the release parameters.

I'm currently trying to convert the `FakeLib.fsproj` file into a ReFake build script, and am about 80% there. Just ironing out a few referencing issues now.

This is a first effort, so I'd welcome feedback.

Relates to issue #235.

### DIFF
--- a/help/refake.md
+++ b/help/refake.md
@@ -1,0 +1,151 @@
+# ReFake - a DSL for Specifying Incremental Builds in FAKE
+
+ReFake is a domain-specific language (DSL) for telling FAKE how to do
+incremental builds natively, without using any other build tool (like
+MSBuild or `xbuild`).
+
+ReFake will build only those parts of the project that are absolutely
+necessary. All the logic for this is built into the FAKE library, so you
+can specify _everything_ in F# instead of XML configuration files or
+whatever.
+
+The idea behind ReFake is exactly the same as the one behind `make`,
+with which you're probably already familiar. To illustrate, let me use a
+simple example 'makefile':
+
+    myapp.dll: myapp.fs
+      fsc -o myapp.dll -a myapp.fs
+
+    myapp.exe: myapp.dll main.fs
+      fsc -o myapp.exe -r myapp.dll main.fs
+
+Here's a direct translation of this simple makefile into ReFake format:
+
+    #r "/path/to/FakeLib.dll"
+    open Fake.FscHelper
+    open Fake.ReFake
+
+    let myapp_dll =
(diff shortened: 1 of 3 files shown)


panesofglass:
This is pretty nice! One thing I would mention is that this may be better as a separate library (as mentioned in #235) that builds on FAKE rather than extensions built into FAKE itself. Also, how do you take care of file ordering?

yawaramin:
@panesofglass agh, I forgot this was still open :-) I've since moved more towards writing MSBuild scripts in F#: https://github.com/yawaramin/FsBuild/blob/master/Build.fsx

PR #617, opened 2015-01-01
Add space after appendWithoutQuotes to fix XUnit include/exclude traits issue

kimsk:
When adding `IncludeTraits` and `ExcludeTraits` together, `XUnitHelper` doesn't add a space before `/-trait` which cause issue with `xunit.console.clr4.exe`.


forki:
thanks

PR #618, opened 2015-01-07
fixes for dotCover support

isakiev:
1) Filters parameter wasn't actually passed to dotCover command line
2) There was no way to perform any additional dotCover configuration (e.g. enable logs, specify verbosity and ~10 less frequently used arguments)
3) cosmetic naming improvements
- dotCover report is often confused with dotCover snapshot
- "dotCover.xml" as a report name is often confused with "dotCover.xml" as a console runner configuration file name

source request:
https://devnet.jetbrains.com/message/5532079


forki:
thanks

PR #620, opened 2015-01-09
fix jenkins detection and improve teamcity

john7doe:


forki:
I thought we already normalize to lowercase?!

PR #621, opened 2015-01-12
Added a reset step before starting a deployment

### LINKED ISSUE #619: Multiple deployments in the same session with Fake.Deploy causes an exception
I am attempting to use Fake.Deploy.exe (version 3.13.3)  to perform a remote deployment of a NuGet package. Currently I have the exe running in a command window and when sending the command over from the client I can see the agent come to life. 

The first run completes without any issues and I can see the output from the install script in my command window. If I then attempt to perform another deployment without restarting the agent process I get the following exception:

```
System.ArgumentException: An item with the same key h
as already been added.
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boo
lean add)
   at Fake.TargetHelper.targetFromTemplate[a](TargetTemplate`1 template, String
name, a parameters) in D:\code\fake\src\app\FakeLib\TargetHelper.fs:line 149
   at Fake.TargetHelper.TargetTemplateWithDependecies@154.Invoke(String name, a
parameters) in D:\code\fake\src\app\FakeLib\TargetHelper.fs:line 154
   at <StartupCode$FSI_0001>.$FSI_0001_Install$fsx.main@() in [path_to_deployment]\active\install.fsx:line 11
```

I am a complete F# novice but could the issue be to do with the fact that TargetHelper.TargetDict is static and therefore retained across deployment sessions?

aembury:
fixes #619

### DIFF
--- a/src/app/Fake.Deploy/DeploymentHelper.fs
+++ b/src/app/Fake.Deploy/DeploymentHelper.fs
@@ -74,6 +74,7 @@ let unpack workDir isRollback packageBytes =
 /// Runs a deployment script from the given package
 let doDeployment packageName scriptFileName scriptArgs = 
     try 
+        TargetHelper.reset()
         let workingDirectory = DirectoryName scriptFileName
         let (result, messages) = 
             FSIHelper.executeBuildScriptWithArgsAndReturnMessages workingDirectory (FullName scriptFileName) scriptArgs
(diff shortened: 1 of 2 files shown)


forki:
that looks like a solution.

forki:
Thanks for this fix. Let's see how this works in production.

test

PR #631, opened 2015-01-20
Adding SpecificMachines option to OctoTools

tkeyser:
Adding support for the SpecificMachines option of Octo.exe


forki:
very cool thanks

PR #632, opened 2015-01-20
Another happy user adding to users.md

Mpdreamz:
Added NEST and Elasticsearch.NET to list of happy users. 

Been using FAKE for a year and I don't think I cursed once, :scream: :smile: 

Thank you for all your work on it.


forki:
cool thanks

PR #640, opened 2015-01-28
Fix jarsigner executing on Windows environment

rflechner:
There is a small bug when using XamarinHelper on Windows.
Simple quotes caused a build fail on Windows environment.


forki:
that was unrelated error.
Thanks for contributing

Issue #643, opened 2015-02-03
Update broke our NuGetPack process


jamescrowley:
Can anyone explain the purpose of this commit:

https://github.com/fsharp/FAKE/commit/2ddc4d4c57693f974cd49feeb0d2adcd3d631d2f

? It's broken our build script when we just updated our FAKE version. 

We previously did:

        Target "Package" (fun _ ->
            !! (currentDirectory @@ "db" @@ "Database.nuspec")
            |>  Seq.iter (NuGetPack (fun param -> { param with 
                                                    ToolPath = nuGetPath
                                                    OutputPath = artifactsDir
                                                    WorkingDir = currentDirectory
                                                    IncludeReferencedProjects = true
                                                    Version = buildVersion
                                                    Properties = [("Configuration", buildMode)] }))
        )

but now Database.nuspec is written to the root of our project (currentDirectory), and then packages up our entire source code tree. Previously it just ran nuget pack on the nuspec file in the db directory and packaged that up. 

I'd fix this but I don't understand the new behavior caused by 
https://github.com/fsharp/FAKE/commit/2ddc4d4c57693f974cd49feeb0d2adcd3d631d2f enough - why is it always writing (and deleting) a nuspec file even if you're giving it one that exists?

forki:
I think you need to set the working dir to currentDirectory @@ "db"

forki:
@ilkerde do you remember what changed?

jamescrowley:
@forki that's what I've been trying as a fix, but this then breaks subsequent items: 

    Target "Package" (fun _ ->
        !! (currentDirectory @@ "db" @@ "Database.nuspec")
        ++ (currentDirectory @@ "rules" @@ "Rules.nuspec")
        ++ (currentDirectory @@ "src" @@ "TestApi" @@ "TestApi.csproj")
        |>  Seq.iter (fun path -> 
            path |> NuGetPack (fun param -> { param with 
                                                ToolPath = nuGetPath
                                                OutputPath = artifactsDir
                                                WorkingDir = Path.GetDirectoryName(path)
                                                IncludeReferencedProjects = true
                                                Version = buildVersion
                                                Properties = [("Configuration", buildMode)] }))
    )

jamescrowley:
The db nuget package gets created okay with the above fix, but I then get 

       Creating .nuspec file at src\TestApi\TestApi.csproj0.0.0.0-beta.nuspec
[...]

PR #644, opened 2015-02-03
Attempting to allow csproj being passed as a NuSpec file

jamescrowley:
First crack at fixing https://github.com/fsharp/FAKE/issues/643


forki:
Thanks for the fast pull request. 

@ilkerde  could you please verify that it also works for you?

ilkerde:
@forki Thanks for letting me know. @jamescrowley Thanks for your fast contribution. I like your changes since the case of accompanying .nuspec is quite often.

However, I checked this PR and have to say that it breaks a test. After running CI I receive

    Failures:
    when packing with nuspec template
    » should create nupkg file (FAIL)
    System.Exception: Error during NuGet package creation. C:\Code\ilkerde.fake\test\TestData\nuget.exe pack -Version 0.0.1 -OutputDirectory "C:\Users\ilker.cetinkaya\AppData\Local\Temp" "fake.nuspec"

This is the test for `NugetPack` to be able to transform templates. Seems that your change breaks this. I suppose that you are changing something along the line of `createNuspecFile`, most probably the return value around [L227](https://github.com/jamescrowley/FAKE/blob/fix.NuGetPack/src/app/FakeLib/NuGet/NugetHelper.fs#L227). I would recommend to have tests for your case (nuspec alongside the csproj) as well.

Any chance to get this fixed (and probably even have a test for it)?

forki:
@jamescrowley can I help you to get this in?

jamescrowley:
@forki This isn't good to merge btw, the whole approach doesn't work - requires a bigger refactor to deal with issues like the current code just tries to delete whatever file you pass in as an argument. And also doesn't work if you pass in a .csproj with a .nuspec alongside that's a template.

jamescrowley:
I think this at least works now @forki @ilkerde. it's ugly as anything, but has a test and will have to defer to others in terms of how we tidy up

src/test/Test.FAKECore/Test.FAKECore.csproj
```diff
@@ -225,6 +225,12 @@
     <Content Include="TestData\AllObjects_2.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+    <Content Include="TestData\bin\Debug\TestNuGetPack.dll">
```
forki: I think these files are gitignored. You need to add them "forced"

jamescrowley:
https://bugzilla.xamarin.com/show_bug.cgi?id=17430 Hmm. Nice. So you can't actually nuget pack on a csproj file on mono. Thoughts @forki ?

forki:
mhm. then maybe just use isMono and throw a good exception.
In the test try to catch that exception in the mono case. does this sound OK?

BTW: we are actively working on http://fsprojects.github.io/Paket/paket-pack.html

jamescrowley:
@forki I think I'd probably make the test check for isMono rather than the nuget helper? given we don't actually know whether nuget.exe in the future would support, and at least you get a decent error output when it fails ow?

forki:
yes that's ok, too

forki:
thanks for being so patient ;-)

PR #645, opened 2015-02-03
Add Verbose to argument list for NuGet update

jamescrowley:


forki:
thanks. I'll try to locate the build error

PR #647, opened 2015-02-07
Fixes backslash escaping in the MSTest module

zsims:
Not currently broken as they're not valid escape characters. Looking to fix tool warnings + avoid future confusion.


forki:
LGTM. thanks.

PR #648, opened 2015-02-07
Adds a VSTest module for working with VSTest.Console

zsims:
Similar to the existing MSTest module. Handy as VSTest.Console.exe is preferred to MSTest for VS2012+, see https://msdn.microsoft.com/en-us/library/ms182486.aspx

This will only work for >= VS2012 (where VSTest is available).

Working example:

```fsharp
#r "FakeLib.dll"
open Fake
open Fake.VSTest

Target "Test" (fun _ ->
    VSTest (fun p -> 
        { p with
            EnableCodeCoverage = true
            Logger = "trx"
            Framework = "Framework45"
            Platform = "x86"
    }) ["UnitTestProject1.dll"]
)

Run "Test"
```

Not sure how to tie this in with the FAKE automated tests, suggestions welcome.


forki:
LGTM

Issue #622, opened 2015-01-15
Replacements in ReplaceInFiles should accept regex patterns


ryanewtaylor:
Currently, it appears the patterns used in ReplaceInFiles must match the string you want to replace exactly. However, it would be very helpful if instead of an exact match the regular expressions could be used in the pattern matching. This would enable, for instance, replacing version strings in arbitrary files with an expression akin to:

````
    !! ".\**\AssemblyInfo.cs"
        |> ReplaceInFiles [("[\d]\.[\d]\.[\d]\.[\d]", version)]

    // replace Major.minor.revision.build with the desired version
````

forki:
This would be a breaking change, right?
How about a new function which does this?

I mark this as up-for-grabs if you are interested please send a PR.

ryanewtaylor:
Yeah. That  might be a breaking change. Perhaps an additional argument with a default value could be added that defaulted to the current behavior but when provided would use regex to find matches? Then again, perhaps a separate ReplaceInFilesRegex would be cleaner.

ryanewtaylor:
@forki, I am not sure of the proper protocol for closing F# - FAKE issues. I believe this issue was initially met by PR #649. May I close this issue or would you prefer to do so?

forki:
You can just close issues when things are fixed. Thanks

PR #649, opened 2015-02-09
Implement functions to replace text in one or more files using regular expressions for the search pattern

ryanewtaylor:
This is my attempt at addressing #622 - Replacements in ReplaceInFiles should access regex patterns. 

This implementation requires that the caller specify the encoding for the files that they are processing. I attempted a solution whereby I did not specify the encoding in the ReadAlltext and WriteAlltext methods but I found that .NET was not always guessing the encoding correctly. As a result this process modified the files from their original encoding (ANSI) to UTF-8. As far as I know there isn't a simple way to detect file encoding so in this implementation I put the responsibility on the caller to specify the encoding.

### DIFF
--- a/src/app/FakeLib/FileHelper.fs
+++ b/src/app/FakeLib/FileHelper.fs
@@ -425,10 +425,10 @@ let ReplaceInFiles replacements files = processTemplates replacements files
 /// Replace all occurences of the regex pattern with the given replacement in the specified file
 /// ## Parameters
 ///
-/// - `file` - The path of the file to process
 /// - `pattern` - The string to search for a match
 /// - `replacement` - The replacement string
 /// - `encoding` - The encoding to use when reading and writing the file
+/// - `file` - The path of the file to process
 let RegexReplaceInFileWithEncoding pattern (replacement:string) encoding file =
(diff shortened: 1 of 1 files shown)


forki:
mhm. cool contribution. 

I think it should be splitted in a function which handles the encoding and one that covers the replacement.
maybe also create a overload which doesn't use the encoding overload.

ryanewtaylor:
My apologizes. I do not follow the suggestion with regards to splitting the function to handle the encoding vs. handling the replacement as the built in RegEx.Replace method handles the replacement. Could you elaborate more on this?

Regarding the second suggestion. I attempted to do just this but being _very_ new to F# I was unable to determine how to make an overloaded function. From what I could gather default parameters are allowed in function but not when the function is a part of a let binding. I was also unsure what a good default would. The "safest" option would probably be either:

a. To not pass any encoding to ReadAllText and WriteAllText and let those methods attempt to figure out the encoding.
b. To pass in Encoding.Default which is the system default.

Neither option seemed particularly "right". For instance, in my project (a mix of .NET and Flex) we have UTF-8 files mixed with ANSI. Any suggestions on a good default?

forki:
Thanks for contributing this.

PR #650, opened 2015-02-09
Fixes #634, added option for MergeByHash

### LINKED ISSUE #634: OpenCover -mergebyhash option
From what I could find it is only possible to supply OpenCover with the typed parameters is that the case? Do I need to create a pull request to get the option for -mergebyhash?

Mharlin:
Also added option to add any additional parameters so that it is possible to use all available commands for OpenCover without re-compiling.


forki:
Thanks for contributiong this.

PR #651, opened 2015-02-11
Team city bug and additions

brianary:
The first commit is a couple of fixes for the TeamCity/NUnit import integration.

The following commit adds all available TeamCity imports.


forki:
LGTM. thanks

PR #652, opened 2015-02-11
Fix SpecFlow MSTest integration

brianary:
SpecFlow uses different parameters for MSTest than it does for NUnit
(see "specflow help mstestexecutionreport").


forki:
thanks for fixing this.

Issue #642, opened 2015-01-30
Fake.Deploy: FsiEvaluationSession locks active deploy


xkrt:
Our script.fsx inside deploy package uses assemblies inside package with #r "libs/SomeLib.dll".

First deploy just fine, but second with same package fails on step "Deleting contents of e:\FAKE\src\app\Fake.Deploy\bin\Debug\.\deployments/PRJ\active".
Libs inside active/libs dir cant be moved, Fake.Deploy locks them (thru FsiEvaluationSession, I guess).

Adding /shadowcopyreferences to commonOptions (FSIHelper.fs) doesnt solve that problem.

xkrt:
Looks like this FSharp.Compiler.Service issue https://github.com/fsharp/FSharp.Compiler.Service/issues/292

vasily-kirichenko:
fixed in https://github.com/fsharp/FAKE/pull/653

PR #653, opened 2015-02-12
Revert fsi exe in fake deploy

vasily-kirichenko:
This is a fix for https://github.com/fsharp/FAKE/issues/642


forki:
Thanks for fixing this

PR #656, opened 2015-02-18
Create a helper for uploading mobile apps to HockeyApp.

bretkoppel:
An initial stab at an upload helper for HockeyApp(http://hockeyapp.net) since TestFlight is shutting down. Follows relatively closely to the original TestFlightHelper, but returns a typed result for usage upstream instead of returning the process exit code.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -32,3 +32,4 @@ nuget Machine.Specifications.Should
 nuget Machine.Specifications.Runner.Console
 nuget Nancy.Testing
 nuget xunit.extensions
+nuget Newtonsoft.Json
--- a/src/app/FakeLib/FakeLib.fsproj
+++ b/src/app/FakeLib/FakeLib.fsproj
@@ -4,9 +4,7 @@
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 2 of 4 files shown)


paket.dependencies
```diff
@@ -31,4 +31,5 @@ nuget FsCheck.Xunit
 nuget Machine.Specifications.Should
 nuget Machine.Specifications.Runner.Console
 nuget Nancy.Testing
-nuget xunit.extensions
\ No newline at end of file
+nuget xunit.extensions
+nuget Newtonsoft.Json
```
forki: I was so happy that I got rid of NewtonSoft.Json ;-)

Seems it coming back over and over again.

forki: I assume we have to fix the NuGet task in FAKE's build.fsx to include the dll.

bretkoppel: Hey, we asked whether you wanted a different serializer up front! Keeping Json.Net makes sense, though, given that it's included elsewhere. =)

I'm actually able to build from both mac and windows without any additional modifications. This might be a side effect of NewtonSoft.Json being included in the Deploy and Deploy.Lib projects... Or I might be totally misunderstanding your comment!

forki: sorry I forgot about that ;-)
All is well.

forki:
ok

forki:
Are you interested in writing a small tutorial like http://fsharp.github.io/FAKE/canopy.html ?

bretkoppel:
Sure thing! Will try to get to it this weekend.

PR #659, opened 2015-02-20
Typed exception when nunit tests fail will not reported to teamcity anymore

brase:
I am a f# newbie. So please tell me if this is the way you would do it.

I think it would be better if the exception is analysed in sendTeamCityError but there the exception was not availlable.


forki:
let's see what happens...

PR #662, opened 2015-02-23
Adding ability to run nuget package restore on a visual studio solution

amcguier:
Also cleaning up some repeated code in RestorePackageHelper


forki:
thanks for contributing

PR #673, opened 2015-03-04
Remvoved internal class generated in AssemblyInfo.Vb

Vidarls:
Working on adding vb.net support in ProjectScaffold, and trying to run the assemblyinfo generate task caused build of vb.net project to fail.

Removing the generated class fixed the problem.

(I know the generated class actually had a syntax error, missing the class name, but even fixing this failed the build.)

It seems adding classes to AssemblyInfo.vb may not be supported, though I could not find any info on it.

Suggesting to remove the generated class from AssemblyInfo.vb for now, until more info is found.

(Also I might look into this part of the build process as part of figuring out what to do with fsprojects/ProjectScaffold#107 )


forki:
I think @simonhdickson, @pblasucci, @tpetricek and @petersondrew worked on this file.
Any objections?

petersondrew:
None here :thumbsup:

PR #675, opened 2015-03-06
Added metadata string property for AssemblyInfoReplacementParams

jchannon:
Added metadata string property for AssemblyInfoReplacementParams

### DIFF
--- a/src/app/FakeLib/AssemblyInfoHelper.fs
+++ b/src/app/FakeLib/AssemblyInfoHelper.fs
@@ -216,7 +216,8 @@ let AssemblyInfoReplacementDefaults =
       AssemblyConfiguration = null
       AssemblyVersion = null
       AssemblyFileVersion = null
-      AssemblyInformationalVersion = null }
+      AssemblyInformationalVersion = null 
+      AssemblyMetadata = null}
 
 let ReplaceAssemblyInfoVersions param = 
     let (parameters : AssemblyInfoReplacementParams) = param AssemblyInfoReplacementDefaults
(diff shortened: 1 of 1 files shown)


forki:
You added a parameter but I can't see where you are using it.

forki:
Ah now I see. Lgtm. The only issue is I'm on vacation so I can only merge it Monday

forki:
And you still need to give the default parameter a value.

forki:
Ok the error now is only related to github Api limits

jchannon:
Actually this might need to be an array rather than a string looking at the screenshot here http://fsharp.github.io/FAKE/assemblyinfo.html

forki:
That thing from the Screenshot is a different api

jchannon:
But Metadata needs to populate multiple values?  eg `"gitcommit","23445fs"`

forki:
Ah that's what you mean.

jchannon:
guess I could just do the below for now
```
fun f -> 
            {f with 
                  AssemblyVersion = fullBuildNumber
                  AssemblyInformationalVersion = fullBuildNumber
                  AssemblyMetadata = "\"githash\"" + "," + "\"6868767gf\""
                  })
```

forki:
I'd prefer to have it an array instead

jchannon:
Probably best you add that, I've spent the last couple of hours trying to get it to work and failed :smile:

forki:
Will do

forki:
65338a7 - is this what you meant?

jchannon:
Yup I tried with a seq and a string concat but couldn't get it to work!

Nice work. Will pull the latest version tomorrow :)

PR #676, opened 2015-03-06
Parallel Builds

krauthaufen:
Added support for parallel builds by using an environment-variable %fake-parallel-jobs%

Note that this of course only works when target-functions can be executed concurrently.
The feature can be enabled by passing for example "fake-parallel-jobs=4" to FAKE.exe

Since parallel build cause problems in TraceHelpers the current "stack" of opened tags is now thread-local.

### DIFF
--- a/src/app/FakeLib/TargetHelper.fs
+++ b/src/app/FakeLib/TargetHelper.fs
@@ -202,8 +202,10 @@ let targetError targetName (exn:System.Exception) =
         sendTeamCityError <| error exn
 
 let addExecutedTarget target time =
-    ExecutedTargets.Add (toLower target) |> ignore
-    ExecutedTargetTimes.Add(toLower target,time) |> ignore
+    lock ExecutedTargets (fun () ->
+        ExecutedTargets.Add (toLower target) |> ignore
+        ExecutedTargetTimes.Add(toLower target,time) |> ignore
+    )
 
 /// Runs all activated final targets (in alphabetically order).
 /// [omit]
@@ -322,36 +324,93 @@ let listTargets() =
 // Instead of the target can be used the list dependencies graph parameter.
 let doesTargetMeansListTargets target = target = "--listTargets"  || target = "-lt"
 
+
+let determineBuildOrder (d : TargetTemplate<unit>) =
+    let levels = Dictionary<string, int>()
+        
+    let found (d : TargetTemplate<unit>) (level : int) =
+        match levels.TryGetValue d.Name with
+            | (true, old) ->
+                if old < level then
+                    levels.[d.Name] <- level
+            | _ ->
+                levels.[d.Name] <- level
+
+    let rec traverse (graphs : seq<TargetTemplate<unit>>) (level : int) =
+        for g in graphs do
+            found g level
+
+        let allDependencies = graphs |> Seq.collect (fun g -> g.Dependencies) |> Set.ofSeq
+
+        if not <| Set.isEmpty allDependencies then
+            traverse (allDependencies |> Seq.map getTarget) (level + 1)
+        
+
+    traverse [d] 0
+
+    let result = 
+        levels |> Seq.map (fun (KeyValue(l,p)) -> (l,p))
+                |> Seq.groupBy snd
+                |> Seq.sortBy (fun (l,_) -> -l)
+                |> Seq.map snd
+                |> Seq.map (fun v -> v |> Seq.map fst |> Set.ofSeq |> Seq.map getTarget |> Seq.toArray)
+                |> Seq.toList
+
+    result
+
 /// Runs a target and its dependencies.
 let run targetName =            
     if doesTargetMeansListTargets targetName then listTargets() else
     if LastDescription <> null then failwithf "You set a task description (%A) but didn't specify a task." LastDescription
-    let rec runTarget targetName =
-        try      
-            if errors = [] && ExecutedTargets.Contains (toLower targetName) |> not then
-                let target = getTarget targetName      
-      
-                if hasBuildParam "single-target" then
-                    traceImportant "Single target mode ==> Skipping dependencies."
-                else
-                    List.iter runTarget target.Dependencies
-      
-                if errors = [] then
-                    traceStartTarget target.Name target.Description (dependencyString target)
-                    let watch = new System.Diagnostics.Stopwatch()
-                    watch.Start()
-                    target.Function()
-                    addExecutedTarget targetName watch.Elapsed
-                    traceEndTarget target.Name                
-        with
-        | exn -> targetError targetName exn
-      
+//    let rec runTarget targetName =
+//        try      
+//            if errors = [] && ExecutedTargets.Contains (toLower targetName) |> not then
+//                let target = getTarget targetName      
+//      
+//                if hasBuildParam "single-target" then
+//                    traceImportant "Single target mode ==> Skipping dependencies."
+//                else
+//                    List.iter runTarget target.Dependencies
+//      
+//                if errors = [] then
+//                    traceStartTarget target.Name target.Description (dependencyString target)
+//                    let watch = new System.Diagnostics.Stopwatch()
+//                    watch.Start()
+//                    target.Function()
+//                    addExecutedTarget targetName watch.Elapsed
+//                    traceEndTarget target.Name                
+//        with
+//        | exn -> targetError targetName exn
+//      
     let watch = new System.Diagnostics.Stopwatch()
     watch.Start()        
     try
         tracefn "Building project with version: %s" buildVersion
-        PrintDependencyGraph false targetName
-        runTarget targetName
+
+        let order = determineBuildOrder (getTarget targetName)
+
+        for par in order do
+            let sem = new System.Threading.SemaphoreSlim(0)
+
+            for target in par do
+                System.Threading.Tasks.Task.Factory.StartNew(fun () ->
+                    try
+                        traceStartTarget target.Name target.Description (dependencyString target)
+                        let watch = new System.Diagnostics.Stopwatch()
+                        watch.Start()
+                        target.Function()
+                        addExecutedTarget target.Name watch.Elapsed
+                        traceEndTarget target.Name       
+                    finally
+                        sem.Release() |> ignore
+                ) |> ignore
+
+            for _ in par do
+                sem.Wait()
+            sem.Dispose()
+
+        //PrintDependencyGraph false targetName
+        //runTarget targetName
     finally
         if errors <> [] then
             runBuildFailureTargets()    
--- a/src/app/FakeLib/TargetHelper.fs
+++ b/src/app/FakeLib/TargetHelper.fs
@@ -378,16 +378,17 @@ let runTargetsParallel (count : int) (targets : Target[]) =
     // since Async.Parallel may or may not create Tasks
     // plain TPL is used here (ensuring that all targets
     // are executed in their own task)
+    let startAndAwait = Async.StartAsTask >> Async.AwaitTask
+
     targets |> Seq.map (fun t -> 
-                System.Threading.Tasks.Task.Factory.StartNew<unit> (fun () ->
-                    sem.Wait()
+                async {
                     try
+                        sem.Wait()
                         runSingleTarget t
                     finally
                         sem.Release() |> ignore
-                )
+                } |> startAndAwait
             )
-            |> Seq.map Async.AwaitTask
             |> Async.Parallel
             |> Async.RunSynchronously
             |> ignore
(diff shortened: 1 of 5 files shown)


src/app/FakeLib/TargetHelper.fs
```diff
@@ -322,6 +324,41 @@ let listTargets() =
...
+
+    traverse [d] 0
+
+
+    let result = 
+        levels |> Seq.map (fun (KeyValue(l,p)) -> (l,p))
+                |> Seq.groupBy snd
+                |> Seq.sortBy (fun (l,_) -> -l)
+                |> Seq.map snd
+                |> Seq.map (fun v -> v |> Seq.map fst |> Set.ofSeq |> Seq.map getTarget |> Seq.toArray)
```
vasily-kirichenko: Maybe `Seq.distinct` is more clear here?

src/app/FakeLib/TargetHelper.fs
```diff
@@ -345,13 +382,72 @@ let run targetName =
                     traceEndTarget target.Name                
         with
         | exn -> targetError targetName exn
+
+    let parallelJobs = environVarOrDefault "fake-parallel-jobs" "1" |> Int32.Parse
```
vasily-kirichenko: `Int32.Parse` -> `int`?

src/app/FakeLib/TargetHelper.fs
```diff
@@ -345,13 +382,72 @@ let run targetName =
...
       
     let watch = new System.Diagnostics.Stopwatch()
     watch.Start()        
     try
         tracefn "Building project with version: %s" buildVersion
-        PrintDependencyGraph false targetName
-        runTarget targetName
+
+        if parallelJobs > 1 then
```
vasily-kirichenko: The body of this `if` is very long and hard to read. What about extracting a function or two?

src/app/FakeLib/TargetHelper.fs
```diff
@@ -345,13 +382,72 @@ let run targetName =
...
+                                finally
+                                    finished.Signal() |> ignore
+                            | _ ->
+                                System.Threading.Thread.Sleep(100)
+                with :? System.OperationCanceledException ->
+                    ()
+
+            let workers =
+                Array.init parallelJobs (fun i -> 
+                    let thread = new System.Threading.Thread(System.Threading.ThreadStart(worker), IsBackground = true)
```
vasily-kirichenko: As you use unordered `ConcurrentBag` for targets, I think it's better to use the thread pool instead of dedicated threads. Just create an `async` for each `Target` and start all of them in parallel (`Async.Parallel >> Async.RunSynchronously`). To set degree of parallelism you could use a `SemaphoreSlim`. Another advantage is that cancellation is handled automatically. As an alternative, you could use raw TPL (`seq of targets -> AsParallel -> WithDegreeOfParallelism -> ToArray`).

src/app/FakeLib/TargetHelper.fs
```diff
@@ -322,6 +324,41 @@ let listTargets() =
 // Instead of the target can be used the list dependencies graph parameter.
 let doesTargetMeansListTargets target = target = "--listTargets"  || target = "-lt"
 
+
+let determineBuildOrder (d : TargetTemplate<unit>) =
+    let levels = Dictionary<string, int>()
```
vasily-kirichenko: Maybe use `Seq.fold` and a `Map` instead of mutable dic and imperative loop?

src/app/FakeLib/TargetHelper.fs
```diff
@@ -345,13 +382,72 @@ let run targetName =
...
-        runTarget targetName
+
+        if parallelJobs > 1 then
+            // multi threaded build
+
+            tracefn "Running parallel build with %d workers" parallelJobs
+
+            let queue = System.Collections.Concurrent.ConcurrentBag()
+            let finished = new System.Threading.CountdownEvent(0)
+            let cancel = new System.Threading.CancellationTokenSource()
```
vasily-kirichenko: let -> use

krauthaufen:
Hi there, I reworked the entire thing a little according to your review.
Sorry for the long code-segments but I'm used to the F# Outlining extension collapsing those functions/etc.

krauthaufen:
Hi vasily, I finally managed to change everything according to your review.
I decided to use the TPL directly since I'm not quite sure about the semantics of Async.Parallel

However all the other points should be adressed now. Please let me know if you have any further suggestions.

src/app/FakeLib/TargetHelper.fs
```diff
@@ -322,36 +324,114 @@ let listTargets() =
...
+            let watch = new System.Diagnostics.Stopwatch()
+            watch.Start()
+            target.Function()
+            addExecutedTarget target.Name watch.Elapsed
+            traceEndTarget target.Name     
+    with exn ->
+        targetError target.Name exn
+
+/// Runs the given array of targets in parallel using count tasks
+let runTargetsParallel (count : int) (targets : Target[]) =
```
vasily-kirichenko: I think we can do better :) What about
```fsharp
let runTargetsParallel (count : int) (targets : Target[]) =
    targets.AsParallel()
        .WithDegreeOfParallelism(count)
        .Select(runSingleTarget)
        .ToArray()
    |> ignore
```
or
```fsharp
let runTargetsParallel (count : int) (targets : Target[]) =
    Parallel.ForEach(targets, ParallelOptions(MaxDegreeOfParallelism = count), runSingleTarget)
```
?

krauthaufen: Yes this is obviously more concise but I didn't want to use "C# style Linq stuff" here.
I liked your first fragment better for some reason :)

src/app/FakeLib/TargetHelper.fs
```diff
@@ -322,36 +324,114 @@ let listTargets() =
 // Instead of the target can be used the list dependencies graph parameter.
 let doesTargetMeansListTargets target = target = "--listTargets"  || target = "-lt"
```
vasily-kirichenko: means -> mean :)

src/app/FakeLib/TargetHelper.fs
```diff
@@ -322,36 +324,114 @@ let listTargets() =
...
 
+/// Determines a parallel build order for the given set of targets
+let determineBuildOrder (d : TargetTemplate<unit>) =
```
vasily-kirichenko: This function is quite complex. I think it's a good idea to write some unit tests for it.

src/app/FakeLib/TargetHelper.fs
vasily-kirichenko: It turns out that the both LINQ solutions uses the thread from which they are executed as one of the worker threads. This is what it's like if it's run from main thread (on a 4 cores box):

image

I think it's totally ok for this case, but you should understand that not all targets will be executed in Thread Pool.

src/app/FakeLib/TargetHelper.fs
krauthaufen: not my function

src/app/FakeLib/TargetHelper.fs
krauthaufen: unit test for traversing an acyclic graph? really? :-D

src/app/FakeLib/TargetHelper.fs
krauthaufen: I'm familiar with .NETs TPL and thats why i said everything will be running on its own Task (not thread) which of course means that also the main thread can be used by the scheduler.
But conceptually (when considering Tasks as abstract threads) everything runs on its own Task...

src/app/FakeLib/TargetHelper.fs
forki: please add unit tests. Sure, they don't prove anything about the function, but experience shows they work very well in practice. Don't forget we have more than 100 contributors (and much more users) from very different fields (a lot don't have any computer science background). Having a set of tests which checks basic edge cases will make it much easier for future contributors to maintain this function. 

If you still think unit tests are useless here, then please look at [FsCheck](https://fsharp.github.io/FsCheck/) and try to prove some important properties. We already have a FsCheck project at https://github.com/fsharp/FAKE/tree/master/src/test/FsCheck.Fake.

krauthaufen: Okay I'm convinced that Tests could help refactoring this function appropriately.
I'm just not sure if the test code will be mich easier to unterstand than the function itself...

forki: it's not about understanding one side alone. It's more like [double-entry bookkeeping](http://en.wikipedia.org/wiki/Double-entry_bookkeeping_system). Both sides play together.

krauthaufen: The problem I'm having is that generators for arbitrary depdendency graphs do not compose with the global mutable state (namely TargetDict).
Nevertheless it would be cool to have arbitrary dependency graphs for testing...

The only solution I came up with is having a different dependency graph and augmenting determineBuildOrder with functions for extracting dependencies..

Any thoughts on that?

krauthaufen: Hi there,
I implemented some very basic tests for the parallel build order checking if it maintains dependencies correctly.
However I did generate the graphs by hand and cleared the TargetDict manually before running the test.
Please let me know if you have any further suggestions.

krauthaufen:
Hi there,
I implemented some very basic tests for the parallel build order checking if it maintains dependencies correctly.
However I did generate the graphs by hand and cleared the TargetDict manually before running the test.
Please let me know if you have any further suggestions.

src/test/FsCheck.Fake/TestParallelBuildOrder.fs
```diff
@@ -0,0 +1,143 @@
...
+    "a" ==> "dep" |> ignore
+    "b" ==> "dep" |> ignore
+    "c" ==> "dep" |> ignore
+
+    let order = determineBuildOrder "dep"
+
+    validateBuildOrder order "dep"
+
+    match order with
+        | [f; [|d|]] ->
```
vasily-kirichenko: Maybe it's worth to introduce an active pattern `Target`, which'd enable syntax like

```fsharp
match order with
| [Target "dep"; [| Target "a"; Target "b"; Target "c" |] -> ...
| _ -> failwith ...
```

src/test/FsCheck.Fake/TestParallelBuildOrder.fs
```diff
@@ -0,0 +1,143 @@
...
+    "a" ==> "b" ==> "d" |> ignore
+    "a" ==> "c" ==> "d" |> ignore
+
+    let order = determineBuildOrder "d"
+    validateBuildOrder order "d"
+
+    match order with
+        | [[|a|];[|b;c|];[|d|]] ->
+            let bcNamesCorrect =
+                if b.Name = "b" then c.Name = "c"
```
vasily-kirichenko: Same as above.

src/test/FsCheck.Fake/TestParallelBuildOrder.fs
krauthaufen: Hi Vasily, I like the idea but the problem with active patterns is that the set of targets on each level is basically unordered which means that 
``
[| Target "a"; Target "b" |]
``

would need to be wrapped into a set-pattern like:
``
TargetSet ["a"; "b"; "c"]
``
where all target-names are (for instance) ordered alphabetically...
I'm going to work on that in the evening...

Let me know if you have any problems with that solution...

krauthaufen:
Hi folks, the tests are now using active patterns (as described above)

vasily-kirichenko:
Looks good for me, thanks! :)

forki:
Awesome! :heart:  

So the only thing that's left is a small tutorial.
Could you please come up with something similar to 
https://github.com/fsharp/FAKE/blob/master/help/specifictargets.md

krauthaufen:
added just a little documentation and a list of issues:
https://github.com/krauthaufen/FAKE/blob/master/help/parallelbuild.md

forki:
very cool. I'm getting excited.

alk

Only one small thing: Can we use `parallel-jobs` instead of `fake-parallel-jobs`?

krauthaufen:
done

forki:
released in 3.21

let's see how it goes.

parallel

thanks for all the effort.

Thorium:
How do I configure dependency tree to take advantage of parallel run?
Is this ok:

    Task_1
      ==> Task_2_A
      ==> Task_3
    
    Task_1
      ==> Task_2_B
      ==> Task_3

Will it run Task_2_A and Task_2_B in parallel if I have e.g. parallel-jobs=2 ?

krauthaufen:
Yes it should indeed find a build-plan where those tasks are executed in parallel...
The graph is basically traversed in level-order which causes some tasks to be started not at the earliest possible time but should give a a reasonable approximation for the build order...

The algorithm starts with the final target and assigns all targets a maximal distance to that...
In the end everything having the same level will be executed in parallel...

Hope that helps

PR #677, opened 2015-03-08
Added basic support for creating Azure Cloud Services.

isaacabraham:
Using this locally, all works fine for a single role cloud service. Feedback appreciated.

### DIFF
--- a/src/app/FakeLib/AzureCloudServices.fs
+++ b/src/app/FakeLib/AzureCloudServices.fs
@@ -1,0 +1,69 @@
+﻿/// Contains tasks to package Azure Cloud Services.
+module Fake.Azure.CloudServices
+
+open System.IO
+open Fake
+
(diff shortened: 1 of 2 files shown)


forki:
very cool. thanks

TWith2Sugars:
Nice, this will come in very useful for me.

forki:
Would be awesome if we would have a tutorial for this as well.

isaacabraham:
@TWith2Sugars it's pretty basic at the moment - it assumes you've already done an MSBuild, and assumes that the svcdef is called ServiceDefinition etc.., but it's working for me :-)

@forki ok - if I were to do one would you just want me to add to the docs in the source (that pushes out to the github website)?

forki:
you could copy and edit https://github.com/fsharp/FAKE/blob/master/help/azurewebjobs.md

PR #685, opened 2015-03-10
AppVeyorEnvironment.RepoTag & .RepoTagName

ctaggart:
We use these to make stable builds based on the tag. I do this in SourceLink too. AppVeyor used to put the tag name in `APPVEYOR_REPO_BRANCH`, but changed it.


forki:
very cool.

PR #691, opened 2015-03-12
Fixes for Globbing in Paket and LongFileName support

colinbull:
Fixes for globbing with trialing slashes and also __very__ experimental long file name support, (I wouldn't use this yet!!).


src/test/Test.FAKECore/Globbing/TestSample1/FileSetSample1Specs.cs
```diff
@@ -147,4 +147,26 @@ public class when_scanning_for_any_dll_and_starting_in_a_subfolder_and_going_up
...
     }
+
+    public class when_scanning_folder_with_trialing_slash : when_extracting_zip
+    {
+        Because of = () => Files =
+            FileSystem
+                .Include("temptest/SampleApp/obj/")
+                .ToArray();
+
+     //   It should_find_dir = () => Files.First().ShouldEndWith(@"temptest\SampleApp\obj");
```
forki: could you please remove that commented code?

colinbull:
W.r.t to the globbing everything should still be the same. And I ended up removing the long file name support.

forki:
> will it break our "normal" stuff?
I meant the long name stuff. But since it's gone...

Issue #670, opened 2015-03-02
@@ path.combine strips root markers from path2, why?


sillyotter:
In dealing with a fake script that I want to run on both Windows and Linux I discovered that the @@ operator runs differently on the two different platforms.  

The code in question is the [combinePaths](https://github.com/fsharp/FAKE/blob/99bf805630805455031684c9369b1872d14a4fdd/src/app/FakeLib/EnvironmentHelper.fs#L21) function.

The combinePaths function is mostly a simple wrapper around [Path.Combine](https://msdn.microsoft.com/en-us/library/fyy7a5kt%28v=vs.110%29.aspx) but adds some extra stripping operations in to path2 that causes it to behave differently on the two different platforms.  Specifically, if you look at the docs for Path.Combine, it says:

> If path2 does not include a root (for example, if path2 does not start with a separator character or a drive specification), the result is a concatenation of the two paths, with an intervening separator character. If path2 includes a root, path2 is returned.

if path2 includes a root, path2 is returned.  In my case, I've got a bit of code that says "src" @@ folderName (its actually in the genFSAssemblyInfo of a FSharpProjectScaffold build.fsx file.)  On a windows box, folderName is a full path, "C:\\Projects\\Something".  "src" @@ "C:\\Projects\\Something" returns "C:\\Projects\Something" as expected.  Calling Path.Combine in mono with "src" and "/home/projects/something" returns "/home/projects/something" just as it is expect to from the docs.  But, the combinePaths function in FAKE is stripping off the linux root / from the second path, resulting in the local path "src/home/projects/something"

The underlying Path.Combine methods seem to do what they are supposed to, why does Fake change that behavior with the extra trim operations?  And can it be fixed to make both environments work the same?

forki:
could you please add test cases and fix the behaviour?

sillyotter:
I believe so.  Ill have a pull request shortly.

sillyotter:
Pull request #695 adds a new non trimming route to Path.Combine and an operator </> to access it.

PR #692, opened 2015-03-12
Modified combinePaths to work like Path.Combine to fix fsharp/FAKE#670

sillyotter:
The original combinePaths insisted on stripping leading characters off of the second path, causing it to not behave like Path.Combine.  /a/b/c joined with /d/e/f is supposed to come out as /d/e/f, but since the / was stripped on before we got /a/b/c/d/e/f, which caused certain operations to fail.  Since on windows roots like C:\ weren't being stripped, on windows the combinePaths worked like Path.Combine.  But on Linux it would result in different results.  This should cause combinePaths and @@ to work the same on both systems.

This only broke some deployment helper methods that used leading /'s, but those have been fixed.  Tests pass on Windows and Linux.  Just make sure not to use leading /'s in second paths if you expect those to join up like they used to.

### DIFF
--- a/src/app/Fake.Deploy/DeploymentHelper.fs
+++ b/src/app/Fake.Deploy/DeploymentHelper.fs
@@ -29,7 +29,7 @@ let getActiveReleases dir = !!(dir @@ deploymentRootDir @@ "**/active/*.nupkg") 
 
 /// Retrieves the NuSpec information for the active release of the given app.
 let getActiveReleaseFor dir (app : string) = 
-    !!(dir @@ deploymentRootDir @@ app @@ "/active/*.nupkg")
+    !!(dir @@ deploymentRootDir @@ app @@ "active/*.nupkg")
     |> Seq.map GetMetaDataFromPackageFile
     |> Seq.head
 
@@ -38,7 +38,7 @@ let getAllReleases dir = !!(dir @@ deploymentRootDir @@ "**/*.nupkg") |> Seq.map
 
 /// Retrieves the NuSpec information for all releases of the given app.
 let getAllReleasesFor dir (app : string) = 
-    !!(dir @@ deploymentRootDir @@ app @@ "/**/*.nupkg") |> Seq.map GetMetaDataFromPackageFile
+    !!(dir @@ deploymentRootDir @@ app @@ "**/*.nupkg") |> Seq.map GetMetaDataFromPackageFile
 
 /// Returns statistics about the machine environment.
 let getStatistics() = getMachineEnvironment()
@@ -102,7 +102,7 @@ let runDeploymentFromPackageFile workDir packageFileName scriptArgs =
 /// Rolls the given app back to the specified version
 let rollback workDir (app : string) (version : string) = 
     try 
-        let currentPackageFileName = !!(workDir @@ deploymentRootDir @@ app @@ "/active/*.nupkg") |> Seq.head
+        let currentPackageFileName = !!(workDir @@ deploymentRootDir @@ app @@ "active/*.nupkg") |> Seq.head
         let backupPackageFileName = getBackupFor workDir app version
         if currentPackageFileName = backupPackageFileName then 
             Failure { Messages = Seq.empty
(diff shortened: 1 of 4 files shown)


src/app/Fake.Deploy/DeploymentHelper.fs
```diff
@@ -29,7 +29,7 @@ let getActiveReleases dir = !!(dir @@ deploymentRootDir @@ "**/active/*.nupkg")
 
 /// Retrieves the NuSpec information for the active release of the given app.
 let getActiveReleaseFor dir (app : string) = 
-    !!(dir @@ deploymentRootDir @@ app @@ "/active/*.nupkg")
+    !!(dir @@ deploymentRootDir @@ app @@ "active/*.nupkg")
```
forki: is this change needed?

sillyotter: Without that change several of the DeploymentHelper tests failed.  With the new combinePath using just a straight call to Path.Combine, app @@ "/active/*.nupkg" just returns "/active/*.nupkg".  And the tests will fail as there isn't anything on the filesystem at /active/whatever.  The leading forward slash implies its a root directory, and Path.Combine says if path2 is a root, just return path2. 

In effect, saying Path.Combine("something", "/active/*.nupkg") is the same as saying Path.Combine("something", "C:\\active\\*.nupkg") and right or not, that is supposed to return C:\active\*.nupkg.    There is no need to add the separator character to the beginning of path2 (the whole point of Path.Combine is to add that for you), and on linux and very likely osx, its also true that that separator can also be interpreted as a root directory indicator.  Its true that its a bit confusing when thinking of @@ as +, but the original stripping of the leading / caused things to work differently on windows and linux.

The original issue I raised was presented as a question because I was unsure about fixing it as I expected that such a change as this would break more than it did.  I'm guessing most everyone uses @@ as if it were a + operator, and don't fully realize its a Path.Combine under the hood and what that means, but it did cause different behavior cross platforms.

forki: mhm yeah. I assume this is a correct fix, but will break tons of stuff.

sillyotter: Well, it didn't break any of the unit tests, but I certainly admit it is a big change as its used quite a lot and will possibly break scripts in the wild.  The issue originally raised is still present though.  Passing in a full path as path2 on windows just returns the full path, doing so on Linux results in a really long nested wrong path.  It is of course your decision to make.  Accepting the change would at least require some important wording in the next release notes.  On windows, it shouldn't make any difference, but it might break some things on linux that are working around the current behavior, or that expected this behavior.

forki: how about introducing the `(/)` operator with correct behaviour and slowly making the @@ obsolete?

sillyotter: I don't see any thing wrong with that.  I can close this pull request and make a new one that adds such an operator and its appropriate tests if you'd like.

forki: yes please

sillyotter: (/) wont work, but (@/) will, or (@@+)  Anything but (/) really.  Any opinions?

forki: `(//)` ?

sillyotter: That's a comment isn't it?

forki: damn it. I remember I saw an operator somewhere which did the same as @@. I'd like to use the same. But can't remember where.

sillyotter: I thought I saw one in a previous issue thread, but I couldn't find it either.  From when I was first searching for why @@ behaved the way it did.  

I was thinking @@+ as it implies its like the old @@ but with more, but its ugly.  Or something more like ++ but that smells of plain concatenation.  <+> smells of haskell monad combinators but isn't bad.  :: isn't overrideable.  Nor is anything \ related.  |+ might work.  In any event, ive got it set to @@+ in my code, but if you come up with something better, let me know and I'll swap it out. I'm still working on a unit test that takes into consideration the different behaviors of Path.Combine under different platforms.

forki: @mausch found it: http://stackoverflow.com/questions/2812084/overload-operator-in-f

forki: there strange implementations of `(/)` and also `(</>)`

PR #694, opened 2015-03-12
Better support for AssemblyMetadata in AssemblyInfoHelper - closes #683

### LINKED ISSUE #683: AssemblyInfoHelper.ReplaceAssemblyInfoVersions should not support AssemblyMetadata
There are two points I see here:
* If one sets the value to something like `["githash"; "abcd0123"; "build-date"; "2015-03-09"]`, then the generated code for C# would be `[assembly: AssemblyMetadata("githash", "abcd0123", "build-date", "2015-03-09")]`, which wouldn't compile.
* Since the `AssemblyMetadata` attribute is allowed to be declared multiple times for an assembly, the current function would replace each line with this attribute by the one above.

So I think the `AssemblyMetadata` attribute shouldn't be supported by this function at all, and `AssemblyInfoFile.CreateCSharpAssemblyInfo` should be used instead.

Changing the type of the `AssemblyMetadata` parameter to `string * string` would solve the first problem. The second one would need some more sophisticated parsing logic, but `CreateCSharpAssemblyInfo` can get the job done as well.

inosik:
Introduced a new function for handling AssemblyMetadata attributes.

Didn't write tests yet, just played around with that in FSI.

What do you say @jchannon?

### DIFF
--- a/src/app/FakeLib/AssemblyInfoHelper.fs
+++ b/src/app/FakeLib/AssemblyInfoHelper.fs
@@ -225,8 +225,8 @@ let ReplaceAssemblyInfoVersions param =
         if isNullOrEmpty value then line
         else regex_replace (sprintf "%s\\s*[(][^)]*[)]" attributeName) (sprintf "%s(\"%s\")" attributeName value) line
 
-    let rec replaceMetadataAttribute metadata line =
-        let replaceMetadataAttribute' key value line =
+    let rec replaceMetadataAttributes metadata line =
+        let replaceSingleMetadataAttribute key value line =
             if isNullOrEmpty key then line
             else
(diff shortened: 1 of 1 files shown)


forki:
That's breaking, but I think @jchannon is the only one using it right now ;-)

jchannon:
:+1:

forki:
please tell me when it's ready to merge

inosik:
Alright, I think I'm done here.

Maybe we should merge `AssemblyInfoHelper.fs` and `AssemblyInfoFile.fs` for the next major version of FAKE, since they provide kind of similar functionality.

forki:
thanks.

PR #698, opened 2015-03-16
Added Recognos Romania to list of users

bogdangaliceanu:
We're using FAKE to build a rather large application so I thought I'd add the company


forki:
thanks

PR #704, opened 2015-03-24
Update AppVeyor.fs

TWith2Sugars:
More appveyor environment vars added


forki:
How about "IsForcedBuild"?

TWith2Sugars:
Good to me, I was just removing the AppVeyor part of the environment name to get the member name.

PR #706, opened 2015-03-24
Use Environment.SystemDirectory to retrieve the system's root directory

bbenoist:
Thus, does not exclusively works on Windows.


forki:
nice.

PR #709, opened 2015-03-24
Code style

bbenoist:
To improve the readability of the source code contributions, I have added:
* A [`.editorconfig`](http://EditorConfig.org) file which can be loaded by [most editors](http://editorconfig.org/#download) if they have loaded the correct plugin.
* A code style section which refers to EditorConfig, some generic F# guidelines and the [CodeMaid](http://www.codemaid.net/) VS extension.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig is awesome:
+http://EditorConfig.org
+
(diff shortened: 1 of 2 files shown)


forki:
thanks.

PR #718, opened 2015-03-30
Find NuGet.exe in current folder (and sub-folders) first, then look in PATH

### LINKED ISSUE #717: Local nuget.exe should take precedence over NuGet on PATH (follow-up #702)
I absolutely agree with #702 that NuGet.exe should be searched in the PATH, but the way how this has been implemented in f83255dd488b4b55148f615eeb80cd44e35e626c, it has now broken several of my CI builds (had the latest NuGet in "./packages", but an older one installed globally), and I have a feeling this might be the case for others, too. 

I think NuGet should be looked for locally *first*, and only in the PATH as a fall-back. Other tools like npm work the same way, so most users come with that expectation, and would find the new behavior counter-intuitive.

mthierba:
Fixes #717. Also bumped priority of the "defaultPath" parameter - treating it as the "default" path (to look in) now, rather than merely as a fall-back.

### DIFF
--- a/src/app/FakeLib/RestorePackageHelper.fs
+++ b/src/app/FakeLib/RestorePackageHelper.fs
@@ -5,30 +5,39 @@ module Fake.RestorePackageHelper
 
 open System
 
-/// Looks for a tool in all subfolders - returns the tool file name.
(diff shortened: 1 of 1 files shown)


forki:
cool. thanks.

PR #721, opened 2015-03-31
Added frameworkAssemblies to nuspec templating

khellang:
japbcvpavbzau9dbuaxf


src/app/FakeLib/NuGet/NugetHelper.fs
```diff
@@ -19,6 +19,10 @@ type NugetFrameworkReferences =
     { FrameworkVersion : string
       References : NugetReferences }
 
+type NugetFrameworkAssemblyReferences =
```
khellang: I guess I could've reused the existing `NugetFrameworkReferences` type and grouped framework assembly references on `FrameworkVersion` instead. What do you think?

src/app/FakeLib/NuGet/NugetHelper.fs
```diff
@@ -153,9 +159,18 @@ let private createNuSpecFromTemplate parameters (templateNuSpec:FileInfo) =
...
     let referencesXml = sprintf "<references>%s</references>" (references + referencesByFramework)
     
+    let getFrameworkAssemblyTags references =
+        references
+        |> Seq.map (fun x -> sprintf "<frameworkAssembly assemblyName=\"%s\" targetFramework=\"%s\" />" x.AssemblyName x.FrameworkVersion)
+        |> toLines
+
+    let frameworkAssemblies = getFrameworkAssemblyTags parameters.FrameworkAssemblies
+
+    let frameworkAssembliesXml = sprintf "<frameworkAssemblies>%s</frameworkAssemblies>" frameworkAssemblies
```
forki: if frameworkAssemblies = [] then "" else
    sprintf "<frameworkAssemblies>%s</frameworkAssemblies>" frameworkAssemblies

src/app/FakeLib/NuGet/NugetHelper.fs
forki: I think it's ok to have both types. We also separate these in paket.

forki:
thanks. only one small issue.

src/app/FakeLib/NuGet/NugetHelper.fs
khellang: `frameworkAssemblies` is a `string`. Solved it another way :smile:

src/app/FakeLib/NuGet/NugetHelper.fs
forki: I wonder if we should make the FrameworkVersion a string list

PR #722, opened 2015-03-31
Changed FrameworkAssembly (string) -> FrameworkAssemblies (string list)

khellang:
According to [this](https://github.com/octokit/octokit.net/blob/master/Octokit.nuspec#L29), comma-separated `targetFramework` should work.


forki:
:+1:

PR #726, opened 2015-04-01
Hide nuget integration options in Fake.Deploy.Web setup

ashic:
* Hides the nuget integration options as the feature isn't ready.
* Adds Knockout to paket file.
* References correct knockout file in layout.

Also, projects and solutions work with VS2015 CTP6+.


forki:
very cool

forki:
thanks

Issue #719, opened 2015-03-30
Building Documentation throws UnresolvedPathReferenceNoRange


neoeinstein:
While in the process of preparing a possible new module for a pull request, I went to build the documentation using the `GenerateDocs` target in the build, but it failed while trying to format the metadata for `FakeLib.dll`. It appears to have been looking for the `ICSharpCode.SharpZipLib` assembly, but, not finding it, choked.
```
C:\dev\neoeinstein\FAKE\packages\FSharp.Formatting.CommandTool\tools\fsformatting.exe "metadataformat" "--generate" "--outdir" "./docs/apidocs/" "--layoutroots" "./help/templates/" "./help/templates/reference/" "--sourceRepo" "https://github.com/fsharp/FAKE/blob/master" "--sourceFolder" "C:\dev\neoeinstein\FAKE" "--parameters" "page-description" "FAKE - F# Make" "page-author" "Steffen Forkmann, Mauricio Scheffer, Colin Bull" "project-author" "Steffen Forkmann, Mauricio Scheffer, Colin Bull" "github-link" "https://github.com/fsharp/FAKE" "project-github" "http://github.com/fsharp/fake" "project-nuget" "https://www.nuget.org/packages/FAKE" "root" "http://fsharp.github.io/FAKE" "project-name" "FAKE - F# Make" --dllfiles "C:\dev\neoeinstein\FAKE\build\FakeLib.dll"
[0 sec] Resolving assembly: FSharp.Compiler.Service.resources, Version=0.0.82.0, Culture=en-US, PublicKeyToken=null
[0 sec] Resolving assembly: FSharp.Compiler.Service.resources, Version=0.0.82.0, Culture=en-US, PublicKeyToken=null
[0 sec] Resolving assembly: FSharp.Compiler.Service.resources, Version=0.0.82.0, Culture=en, PublicKeyToken=null
[0 sec] Resolving assembly: FSharp.Compiler.Service.resources, Version=0.0.82.0, Culture=en, PublicKeyToken=null
[3 sec] Reading assembly: C:\dev\neoeinstein\FAKE\build\FakeLib.dll
[3 sec] Parsing assembly (with documentation file 'C:\dev\neoeinstein\FAKE\build\FakeLib.XML')
[3 sec] Error while building member-name for Fake.AdditionalSyntax.( ? ) because: not a named type
[3 sec] Error while building member-name for Fake.AdditionalSyntax.( ?<- ) because: not a named type
[...]

PR #727, opened 2015-04-01
Generalize archive helpers

neoeinstein:
Provide a more comprehensive suite of archive and compression helpers, including zip, gzip, bzip2, tar, and tar.gz/tar.bz2.

This module isn't `AutoOpen`, and should be considered as a general case replacement for the `ZipHelper` module (which is `AutoOpen`). The modules in `ArchiveHelper` could be expanded to support some of the use cases provided in `ZipHelper`, but this is a good base from which to start.

Note that this will fail to build documentation until #719 is resolved. (`.\build GenerateDocs` fails)

### DIFF
--- a/src/app/FakeLib/ArchiveHelper.fs
+++ b/src/app/FakeLib/ArchiveHelper.fs
@@ -1,9 +1,7 @@
 ﻿/// Provides utility tasks for storing and compressing files in archives.
 module Fake.ArchiveHelper
 
-open System
 open System.IO
-open ICSharpCode.SharpZipLib.Core
 
 [<Literal>]
 let private DefaultBufferSize = 32768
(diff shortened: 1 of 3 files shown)


forki:
cool.

is it possible to let the ziphelper call these functions?

neoeinstein:
I can modify ZipHelper to delegate to these functions. I'll make sure all the PascalCased functions do the right delegation. Do we particularly care about keeping the camelCased functions, or are these generally considered open to removal?

neoeinstein:
With these changes, would there be any interest in marking `ZipHelper` as `Obsolete`, encouraging people to move over to the analogues in `ArchiveHelper` without breaking existing consumers?

forki:
thanks. for adding this.

PR #728, opened 2015-04-01
sample usage for xunit2 fixed

robertmuehsig:
I'm not sure, but I guess the function in the sample should be xunit2.


forki:
thanks

PR #736, opened 2015-04-03
looks in current nuspec for namespace

ashic:
When Fake.Deploy is running a deployment, it looks up package details from the nuspec file. Currently, only a couple of namespaces are used, and depending on the version of nuget used, other namespaces may be in the nuspec file. For example, if the nuspec file used to generate a package doesn't have a namespace specified, any of the following may get used:

"http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd"
"http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd"
"http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd"
...

This pull request will try the (currently hardcoded) values, but will also use the document's namespace. This provides more robust handling of versions.


forki:
cool. maybe we should "paket github link" the nuspec file from Paket.

PR #744, opened 2015-04-07
fixes #740

### LINKED ISSUE #740: FAKE fails to give file-paths for NUnit on OS X
The NUnit task invokes mono. If you want to get the file paths of test failures, you need to pass the `--debug` flag, or the output will look like this:

```
Running build failed.
Error:
Fake.UnitTestCommon+FailedTestsException: NUnit test failed (30).
  at Fake.NUnitSequential.NUnit (Microsoft.FSharp.Core.FSharpFunc`2 setParams, IEnumerable`1 assemblies) [0x00000] in <filename unknown>:0
  at FSI_0001+clo@62-4.Invoke (Microsoft.FSharp.Core.Unit _arg3) [0x00000] in <filename unknown>:0
  at Fake.TargetHelper+targetFromTemplate@155[Microsoft.FSharp.Core.Unit].Invoke (Microsoft.FSharp.Core.Unit unitVar0) [0x00000] in <filename unknown>:0
  at Fake.TargetHelper.runSingleTarget (Fake.TargetTemplate`1 target) [0x00000] in <filename unknown>:0
```

haf:
@forki Like this?

### DIFF
--- a/src/app/FakeLib/ProcessHelper.fs
+++ b/src/app/FakeLib/ProcessHelper.fs
@@ -16,7 +16,7 @@ let startedProcesses = HashSet()
 /// [omit]
 let start (proc : Process) = 
     if isMono && proc.StartInfo.FileName.ToLowerInvariant().EndsWith(".exe") then
-        proc.StartInfo.Arguments <- "\"" + proc.StartInfo.FileName + "\" " + proc.StartInfo.Arguments
+        proc.StartInfo.Arguments <- "--debug \"" + proc.StartInfo.FileName + "\" " + proc.StartInfo.Arguments
         proc.StartInfo.FileName <- "mono"
 
     proc.Start() |> ignore


forki:
this will always set the flag. for all tasks. is this a good idea?

haf:
I think so?

This is the docs:

```
              Turns on the debugging mode in the runtime.  If an assembly was compiled with debugging information, it will produce line number information for stack traces.

              The optional OPTIONS argument is a comma separated list of debugging options.  These options are turned off by default since they generate much larger and slower code at
              runtime.

              The following options are supported:

              casts  Produces a detailed error when throwing a InvalidCastException.   This option needs to be enabled as this generates more verbose code at execution time.

              mdb-optimizations
                     Disable some JIT optimizations which are usually only disabled when running inside the debugger.  This can be helpful if you want to attach to the running process
                     with mdb.

              gdb    Generate and register debugging information with gdb. This is only supported on some platforms, and only when using gdb 7.0 or later.

```

It seems that just debug without flags means it's fairly fast. I'd rather have a build system that tells me what's wrong (perhaps even the *casts* flag then), than gives me bad output.

forki:
> I'd rather have a build system that tells me what's wrong (perhaps even the casts flag then), than gives me bad output.

totally agree.

PR #749, opened 2015-04-08
Force build server output if the user wishes to

eggapauli:
The parameters have probably been ignored by accident, because the docs already say "forces [TeamCity|AppVeyor CI] mode (normally auto-detected)"


forki:
thanks.

PR #750, opened 2015-04-09
Added support for Mage's CertHash parameter

WallaceKelly:
When signing a ClickOnce manifest, mage.exe can be called with either the -CertFile or the -CertHash options. The -CertFile option signs the manifest using a file from the file system. The -CertHash option signs the manifest using a certificate from the certificate store.


forki:
cool. thanks.

PR #758, opened 2015-04-15
Ref xunit 2 and use xunit.runner.console.

philderbeast:
Stop using retired xunit.runners and use xunit.runner.console instead.
Had to upgrade two projects to dotNet 4.5 from 4.0 to build with the
updated xunit references. Fix 758.


forki:
did you see that we have http://fsharp.github.io/FAKE/apidocs/fake-xunit2helper.html?

We wil make the xUnit helper obsolete.

PR #766, opened 2015-04-24
WIP: Implement file system change watching.

voiceofwisdom:
Created a wrapper around `FileSystemWatcher` that makes it easy to watch the disk for changes. Uses a glob pattern to filter the changes to a subset of the files. I had to reimplement globbing using a regex, the current implementation requires walking the file system every time, the regex can do a simple compare to see if a path is matched.

Example :
```f#
Target "Watch" (fun _ ->
    let stop = !! "c:/projects/watchDir/*.txt" |> WatchChanges (fun changes -> 
        RunTarget "Tests"
    )

    System.Console.ReadLine() |> ignore

    stop() //if you need to cleanup the watches.
)
```

There are two reasons I marked this as WIP. 

First, is I dont know if the documentation I added is sufficient. This seems to be cool enough to justify its own article in the docs, but I wanted to check before writing that.

Second, if you dont do something to keep FAKE from exiting, it will exit immediately after setting up the watcher. Hence the `ReadLine` at the end of the function. Personally I dont find this to be pleasant. To make this not required we would have to keep track of all running `WatchChanges` and prevent FAKE from exiting while they are running. What are others thoughts on this?

### DIFF
--- a/src/app/FAKE/FAKE.fsproj
+++ b/src/app/FAKE/FAKE.fsproj
@@ -24,8 +24,8 @@
     <WarningLevel>3</WarningLevel>
     <StartAction>Project</StartAction>
     <StartProgram>C:\sources\FAKE\tools\FAKE\FAKE.exe</StartProgram>
-    <StartWorkingDirectory>d:\Code\Paket</StartWorkingDirectory>
-    <StartArguments>GenerateHelp -st</StartArguments>
+    <StartWorkingDirectory>C:\Projects\FAKE\src\test\testscripts</StartWorkingDirectory>
+    <StartArguments>test1.fsx</StartArguments>
     <PlatformTarget>x86</PlatformTarget>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
--- a/src/app/FakeLib/ChangeWatcher.fs
+++ b/src/app/FakeLib/ChangeWatcher.fs
@@ -1,0 +1,101 @@
+﻿[<AutoOpen>]
+/// This module contains helper function to create and extract zip archives.
+module Fake.ChangeWatcher
+open System.IO
+
+type FileStatus =
+| Deleted
+| Created
+| Changed
+
+type FileChange = {
+    FullPath : string
+    Name : string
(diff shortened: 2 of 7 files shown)


forki:
hey cool. I will take a deeper look soon.

regarding the globbing change: cc @colinbull

forki:
regarding your sample. Maybe WatchChanges could return IDisposable!? Then we could do:

     use stop = !! "c:/projects/watchDir/*.txt" |> WatchChanges (fun changes -> 
            RunTarget "Tests"
    )

   ...
   stop.Dispose() // or rely on default dispose

colinbull:
This all looks good. Doesn't actually change the globbing implementation used in (!!) etc. It is in addition to I believe, unless I have missed something on my quick check. If this is the case happy to see this merged, as I have a potential use for it :+1:

forki:
What happens if I add a new file which would match the pattern?

colinbull:
Looks as if the even will simply get raised. If you want it in your file
set you would have to run (!!) pattern again. I think.

forki:
my main use case at the moment is documentation / website creation. When i add a new *.md file I want to rerun the FSharp.Formatting.

colinbull:
In that case, it is just a call to `Literate.ProcessDirectory` on the event? Isn't it, that will then consume the new file.

forki:
no the same thing is needed to copy the new pictures and stuff like that.

voiceofwisdom:
If you add a new file the standard `!!` glob will pick it up again when rerun. `WatchChanges` will raise a change event for new files that match the glob.

@colinbull  you are correct, this is another implementation of the glob. It should match the existing glob syntax exactly.

forki:
@agross what do you think?

agross:
This looks like a nice addition to FAKE. I'd say docs and tests are the main use cases.

forki:
I remember you creating a similar issue, but can't find it

forki:
thanks a lot.

forki:
I changed it to IDisposable and merged it in order to try it out downstream.
I hope you can a bit of of docs in a new PR

voiceofwisdom:
Created a pr for a small tutorial, #774.

PR #767, opened 2015-04-24
Treat non csproj files as nuspec files

MaximilianLackaw:
Currently nuspec files must have a .nuspec file extension to be treated as valid nuspec templates. We have nuspec files without this extension which no longer work with up to date versions. The only reason why this distinction exists, is to cover .csproj files. So our suggestion is to just look for the .csproj file extension.

### DIFF
--- a/src/app/FakeLib/NuGet/NugetHelper.fs
+++ b/src/app/FakeLib/NuGet/NugetHelper.fs
@@ -233,9 +233,9 @@ let private createNuSpecFromTemplate parameters (templateNuSpec:FileInfo) =
 
 let private createNuSpecFromTemplateIfNuSpecFile parameters nuSpecOrProjFile = 
     let nuSpecOrProjFileInfo = fileInfo nuSpecOrProjFile
-    match nuSpecOrProjFileInfo.Extension.ToLower() = ".nuspec" with
-    | true -> Some (createNuSpecFromTemplate parameters nuSpecOrProjFileInfo)
-    | false -> None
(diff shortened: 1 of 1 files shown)


forki:
I'd like to accept this change, but I assume yoo need to consider vb, F# and other project types.

Question: why don't you go with the nuget default?

ilkerde:
@forki Look at the change. It's just reversing the defaults. @MaximilianLackaw did not care about whatsoever project files are going to be supported. I think it was @jamescrowley only going for `.csproj` in #644.

forki:
I was talking about https://github.com/fsharp/FAKE/pull/767/files#diff-88b397b8b3d4d92a120c4eb5fce2b2d5R236

it introduces the new constant "csproj".

forki:
nice.

ilkerde:
I was talking about that as well. Maybe I can't read F#, but the match expression reads to me "If file ends with csproj, please do some magic, otherwise treat it as a nuspec template file". Isn't this exactly what #644 was attempting to but went for the inverted logic "if it ends with nuspec, then treat it as a nuspec template file, otherwise do some magic".

ilkerde:
So @MaximilianLackaw fixed it anyway. Forget about my nitpicking.

forki:
yep. but my point was: we should do the magic for all project types.

forki:
Just out of curiosity: why don't you go for *.nuspec?

ilkerde:
Well you know it might be that template files are going to be packaged within nugets but may not be suffixed `.nuspec` because they would be omitted by the pack command.

PR #776, opened 2015-04-29
Fix package id parsing and avoid NPE when feed is missing some propertie...

chillitom:
...s

Nuget feed's <m:properties> doesn't contain a <d:Id>, the id is instead found in the <entry>'s <title> element.

The <id> element in a nuget feed contains something like "http://www.nuget.org/api/v2/Packages(Id='_Atrico.Lib.CommonAssemblyInfo',Version='1.0.0.0')" which is not what we want here.

Some nuget package sources omit some elements, for example ProGet doesn't report the Language element for handle missing elements by returning an empty string.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -36,5 +36,6 @@ nuget xunit.extensions
 nuget Newtonsoft.Json
 nuget Microsoft.AspNet.Razor 2.0.30506
 nuget Microsoft.AspNet.WebPages 2.0.30506
+nuget Unquote
 
 github matthid/Yaaf.FSharp.Scripting src/source/Yaaf.FSharp.Scripting/YaafFSharpScripting.fs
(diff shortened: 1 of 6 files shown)


forki:
Hey that sounds good. But we should use id tag whenever it is there

chillitom:
I haven't seen any feeds that contain an d:Id tag, the <id> atom element is not what we want.. the nuget.org stream doesn't appear to have them either..  am I missing something though?

chillitom:
Fyi, just in the process of putting together a testcase

chillitom:
There isn't an F# tests project from what I can see, am I okay to add one?

forki:
Yep.

chillitom:
Had difficulty getting a new test project to run as part of the main build so have stuck the new test in the `FSCheck.Fake` project for now.

I'm remain a little confused as to how the code that was there before worked.  Have I done something dumb like reference the wrong stream? It appears to line up with the example in `getLatestPackage`

forki:
thanks for this.

PR #783, opened 2015-05-03
simplified fsharp.targets location

ctaggart:
* makes it work on VS 2015 too
* looks compatible with Mono 4, I can test, but haven't yet
* let's see if the Travis build passes :)
* this is the only line needed for VS 2013 & above

``` xml
<Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\FSharp\Microsoft.FSharp.Targets" />
```


ctaggart:
Yay! The reason this works on Mono now with `VisualStudioVersion` not set is that they have a `VisualStudio/v/` directory in addition to the versioned ones. :)

image

ctaggart:
Documented here. :)
http://blog.ctaggart.com/2015/05/importing-fsharptargets-compatible-with.html

forki:
very cool.

PR #784, opened 2015-05-04
Fix FAKE not working on machines with only F# 4.0 installed

ovatsus:


forki:
thanks

forki:
image

If I build locally on system with VS 2013 + VS 2015 RC I now get method not found errors.

ovatsus:
:( Maybe the binding redirect in app.config needs updating? Or maybe we can just reorder so 4.0 comes after 3.1

forki:
ok the issue is not this, but @ctaggart's https://github.com/fsharp/FAKE/pull/783.
I had to revert that (otherwise nice change) and everything works again.

ctaggart:
:(

ctaggart:
What's the FAKE version with the problem? I'll take a look and find the root cause when I get a chance.

forki:
There is no published version with a bug. It's just that I can't build (due
to test errors) on the machine where I do the release. So I reverted your
changes and it builds again. Only tests in FAKE.deploy Web stuff are
affected.

ctaggart:
The looks like a binding redirect problem. How do I reproduce?

forki:
I have VS2015 rc and f# 3.1 installed. Then try to build your commit.

PR #785, opened 2015-05-04
Add basic support for CMake configuration and build

bbenoist:
Description of the wrapped tool:

> CMake, is a cross-platform, open-source build system. CMake is a family of tools designed to build, test and package software. CMake is used to control the software compilation process using simple platform and compiler independent configuration files. CMake generates native makefiles and workspaces that can be used in the compiler environment of your choice.

See the [official website](http://www.cmake.org/) for more information about CMake.

Features:
* Configure and build any CMake project with FakeLib's CMakeHelper.
* Supports the most used CMake option switches.
* Strong CMake variable declaration to avoid conversion issues.
* Executable search algorithm working on both Windows and Unix.
* Toolchain file support.
* Setting unwrapped CMake arguments manually remains possible.
* Setting CMake arguments manually remains possible.

It has been tested on both Windows (with VS 2012 and 2013) and the Ubuntu/Vagrant box (g++) using a simple C++ project.

### DIFF
--- a/Samples/CMakeSupport/Advanced/Communicator/CMakeLists.txt
+++ b/Samples/CMakeSupport/Advanced/Communicator/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required (VERSION 2.6)
+
+project (Communicator CXX)
+
+include_directories ("${PROJECT_SOURCE_DIR}")
+add_library (Communicator Communicator.cxx)
+
+install (TARGETS Communicator
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin
+    COMPONENT library
+)
+install (FILES Communicator.h DESTINATION include)
--- a/Samples/CMakeSupport/Advanced/Communicator/Communicator.cxx
+++ b/Samples/CMakeSupport/Advanced/Communicator/Communicator.cxx
@@ -1,0 +1,13 @@
+#include "Communicator.h"
+
+using namespace std;
+
+Communicator::Communicator(const string & interlocutorName) :
+  _interlocutorName(interlocutorName)
+{
+}
+
+string Communicator::hello() const
+{
+  return "Hello " + _interlocutorName + "!";
+}
--- a/Samples/CMakeSupport/Advanced/Communicator/Communicator.h
+++ b/Samples/CMakeSupport/Advanced/Communicator/Communicator.h
@@ -1,0 +1,17 @@
+#ifndef COMMUNICATE_H
+#define COMMUNICATE_H
+
+#include <string>
+
+// Provides helpers for communicating with an end-user.
+class Communicator
+{
+  private:
+    std::string _interlocutorName;
+
+  public:
+    explicit Communicator(const std::string & interlocutorName);
+    std::string hello() const;
+};
+
+#endif // COMMUNICATE_H
--- a/Samples/CMakeSupport/Advanced/Hello/CMakeLists.txt
+++ b/Samples/CMakeSupport/Advanced/Hello/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required (VERSION 2.6)
+
+project (Hello CXX)
+
+# Not using FindLibrary to keep the example simple.
+if (NOT DEFINED COMMUNICATOR_INCLUDE_DIR OR NOT DEFINED COMMUNICATOR_LIBRARY)
+  message (FATAL_ERROR "Please define both COMMUNICATOR_INCLUDE_DIR and COMMUNICATOR_LIBRARY.")
+endif ()
+
+include_directories ("${COMMUNICATOR_INCLUDE_DIR}")
+
+add_executable (hello main.cxx)
+
+target_link_libraries (hello "${COMMUNICATOR_LIBRARY}")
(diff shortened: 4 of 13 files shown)


bbenoist:
@forki The `CMakeHelper` generated documentation includes two F# examples which does not seems to be colored by FSharp.Formatting. I did not find a way to fix it :cry:

bbenoist:
@forki I have tried to add a specific unit-test for `CMakeHelper` but strugle at determining what is the best way to do so.

Here is a brief desciption of what's expected:

#### Requiremed file system at runtime
Here is the file system which should be present at runtime
````
|-- <optional-directory-structure> // CMake does not requires a structured relation between the build and source directories.
| |-- CMakeTest // A directory containing some source code to be built with CMake.
| | |-- CMakeLists.txt
| | |-- main.cxx
|-- <optional-directory-structure> // CMake does not requires a structured relation between the build and source directories.
| |-- CMakeTest-Build // Created by `ensureDirectory`. This is the directory containing every file generated by CMake and the build tools it manages.
| | |- <generated-files>
````

#### Required build tools
In order to execute the routines of `CMakeHelper`, the following tools should be available:
* CMake: Not available on NuGet but a portable package is available on Chocolatey under the name [`cmake.portable`](https://chocolatey.org/packages/cmake.portable).
* A native build instructor: Visual Studio 2013 (the same version as FAKE) on Windows and GNU Make on *nix.
* A C++ compiler: Detected by CMake, can be MSVC on Windows and g++ on *nix

#### Example test
At best, the test should be something like the following:

```fsharp
let sourceDir = "./CMakeTest"
let buildDir = "./CMakeTest-Build"
CleanDir sourceDir
ensureDirectory buildDir
CMakeGenerate (fun p ->
    { p with
        SourceDirectory = sourceDir
        BinaryDirectory = buildDir
        Generator = if isUnix then "Unix Makefiles" else "Visual Studio 12 2013"
    }
)
CMakeBuild (fun p-> {p with BinaryDirectory = buildDir})
````

Any opinion about this?

**Comment by forki** [comment]
Hi,

thanks for contributing. 

Integration tests are very hard to do. If you are going to use this stuff in production then I don't think you need to provide such tests. I  will always ask you before I merge changes on this.

File: src/app/FakeLib/CMakeHelper.fs [hunk 1]
```diff
@@ -0,0 +1,286 @@
+﻿[<AutoOpen>]
```
forki: please remove the AutoOpen

src/app/FakeLib/CMakeHelper.fs
```diff
@@ -0,0 +1,286 @@
...
+        [ generator; toolchain; toolset; platform; caches; variables; cacheEntriesToRemove
+          [parameters.AdditionalArgs; "\"" + parameters.SourceDirectory + "\""]
+        ] |> List.concat |> String.concat " "
+    callCMake parameters.ToolPath parameters.BinaryDirectory args parameters.Timeout
+
+/// Calls `cmake --build` to build a project.
+/// ## Parameters
+///  - `setParams` - Function used to manipulate the default CMake parameters. See `CMakeBuildParams`.
+let CMakeBuild setParams =
+    let parameters = setParams CMakeBuildDefaults
```
forki: please add startTask and endTask tracing functions as in https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/DocuHelper.fs#L34

src/app/FakeLib/CMakeHelper.fs
bbenoist: OK, I was not aware that some modules are not auto-opened because of their low-usage... As FAKE scripts tends to hide as most F# specifics as possible (_"It is using an easy domain-specific language (DSL) so that you can start using it without learning F#."_), I believed that removing the need for specifying any module name was a design choice.

As not auto-opening `CMakeHelper` forces end-users to explicitly use the module name at each function call, I would like to rename:
* `CMakeHelper` into `CMake`
* `CMakeGenerate` into `Generate`
* `CMakeBuild` into `Build`

Usages of the helper would be cleaner because of their `<Noun>.<Verb>` form.

Are you OK with this?

forki: Yes that's much better. We try to minimize the autoopens since had name
conflicts.

src/app/FakeLib/CMakeHelper.fs
bbenoist: OK.

Is there a need for tracing the whole `generate` and `build` tasks or would using them from `callCMake` be sufficient (it would simplify the build of the description string)?

More globally, what is the point of these functions?

forki: yes try to make it as simple as possible.

2015-05-05 13:17 GMT+02:00 bbenoist <notifications@github.com>:

bbenoist:
> Integration tests are very hard to do. If you are going to use this stuff in production then I don't think you need to provide such tests. I will always ask you before I merge changes on this.

Yes, I intend to use it on production code but only on Windows as the .NET software I am working on cannot work with Mono.
In fact, I already had to make such testing to ensure a working Linux support.
Making it publicly available just takes a me a few seconds but would make people reusing `CMakeHelper` more autonomous.

Are you OK with adding a dedicated example in the `Samples` directory instead?

forki:
> Are you OK with adding a dedicated example in the Samples directory instead

yes

forki:
is this ready to merge?

bbenoist:
Not yet, French May is filled with holydays this year :sweat_smile:
It should be OK during the week...

bbenoist:
@forki I have finished my improvements. You can now merge it if you are OK with its content.

forki:
very cool. thanks a lot

PR #786, opened 2015-05-05
close stdin in asyncShellExec to avoid hangs

chillitom:
Nothing is written to stdin so we can safely close it, this will allow child processes that try to read from stdin to avoid hanging the build


forki:
cool. thanks

PR #793, opened 2015-05-13
Add test adapter path to vs test params

richardjharding:
It would be nice to be able to use the vstest console runner to run tests that are written with other testing frameworks such as xUnit - this would allow developers to make use of the vstest coverage and trx logger - for instance when used in conjunction with SonarQube
This pull request adds the optional TestAdapterPath to the vstest params type


forki:
cool thanks a lot.

PR #798, opened 2015-05-15
xUnit use relative path for output dir

neoeinstein:
A rooted *nix path (beginning with /) is interpreted by the xUnit runner as a command line directive. This causes the xUnit runner to fail when report directories are specified when building on Mono.


neoeinstein:
I used the FAKE build as a guinea pig to test this since it already has an xunit task as part of the build. When copying the built `FakeLib.dll` into the FAKE tools directory and re-running the build, it appropriately uses the relative paths.

I have a separate commit which I can also add that will cause FAKE to begin generating xUnit HTML reports into the reports directory, but as that breaks the Travis-CI build (at least until FAKE uses a version of FAKE which includes #798), I've withheld that commit.

neoeinstein:
Superceded by #800

PR #799, opened 2015-05-16
Fix WatchChanges on Mac, fix Dispose, improve Timer usage

tpetricek:
This does a couple of improvements to the watcher:

 - Make it work on Mac/Unix (if you specified full path before, the `getRoot` function would drop the `/` from the beginning and treat the path as not rooted - and you'd end up with wrong path)
 - Make it possible to specify just file names without `*` pattern (e.g. `!! "app.fsx"` if I want to watch for just a single file)
 - Rather than running the timer all the time, we now only start it when something happens. I also increased the delay a bit (to give the system a bit more time to collect all events related to one action)
 - Also, it now ignores events that happen when the user is reacting to another event (if the user-specified function touches the file - we do not want to get into an infinite loop).

This is mostly triggered by the Suave Dojo (using WatchChanges to live-reload web server), but I think all the changes are useful for the other scenarios I'm aware of.


forki:
cool. thanks

PR #800, opened 2015-05-16
Rewrite xunit tasks

neoeinstein:
This pull request supersedes #798 and proposes a fix for #797.

Changes in the `xUnit` and `xUnit2` tasks are breaking. There is an addition of an `xUnitSingle` task which explicitly indicates that it takes a single assembly for xUnit testing to allow a consumer to have more control over how the target is consumed since xUnit v1 only accepts one assembly at a time.

This PR also includes changes to allow specifying the reporting file names for HTML, XML, and NUnit-style reports as well as additional documentation and some refactoring to limit the exposed interface. Changes relevant to each task have been staged as separate commits.

Specifications have been added to the test project for all of these changes. Once merged, consumers who used the following xUnit parameters will break:
* `XUnitParams`
 * `ConfigFile` - Removed: Was not being used anyway
 * `HtmlOutput` - Removed: Pass in report file name path as `HtmlOutputPath`
 * `XmlOutput` - Removed: Pass in report file name path as `XmlOutputPath`
 * `NUnitXmlOutput` - Removed: Pass in report file name path as `NUnitXmlOutputPath`
 * `WorkingDir` - Type change: Now accepts a `string option`
 * `Verbose` - Removed: Pass in inverted value as `Silent`
 * `OutputDir` - Removed: No longer used
 * `IncludeTraits` - Type change: Now accepts a list of name-value tuples
 * `ExcludeTraits` - Type change: Now accepts a list of name-value tuples
* `XUnit2Params`
 * `ConfigFile` - Removed: Was not being used anyway
 * `Parallel` - Type change: Now accepts a discriminated union `ParallelOption`
 * `MaxThreads` - Type change: Now accepts a discriminated union `MaxThreadsOption`
 * `HtmlOutput` - Removed: Pass in report file name path as `HtmlOutputPath`
 * `XmlOutput` - Removed: Pass in report file name path as `XmlOutputPath`
 * `XmlV1Output` - Removed: Pass in report file name path as `XmlV1OutputPath`
 * `WorkingDir` - Type change: Now accepts a `string option`
 * `OutputDir` - Removed: No longer used
 * `IncludeTraits` - Type change: Now accepts a list of name-value tuples
 * `ExcludeTraits` - Type change: Now accepts a list of name-value tuples
 * `Teamcity` - Renamed: `ForceTeamCity`
 * `Appveyor` - Renamed: `ForceAppVeyor`

PR #802, opened 2015-05-18
NUnit documentation improvements

LimoWanKenobi:
This PR contains small improvements to documentation related to NUnit.

- Fixed link to `NUnitSequential` task on getting started page. Both links on getting started were pointing to `NUnitParallel` documentation.

- Linked `NUnitParams` apidoc on `NUnitSequential` and `NUnitParallel` documentation. I am not sure if this is the correct way to add a link to another file in the API docs.

- Unified NUnit documentation links to documentation for NUnit v2.6.4 (current stable version).

- Documented default values for `NUnitParams` in `NUnitDefaults` docs.

- Added docs for the `NUnitParams` record fields with text from NUnit documentation.

### DIFF
--- a/src/app/FakeLib/UnitTest/NUnit/Common.fs
+++ b/src/app/FakeLib/UnitTest/NUnit/Common.fs
@@ -40,24 +40,44 @@ type NUnitDomainModel =
 ///
 /// For reference, see: [NUnit-Console Command Line Options](http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6.4)
 type NUnitParams = 
-    { IncludeCategory : string
+    { 
+      /// Default: ""
+      IncludeCategory : string
+      /// Default: ""
       ExcludeCategory : string
+      /// Default FAKE will try to locate nunit-console.exe in any subfolder of the current directory.
       ToolPath : string
+      /// Default:"nunit-console.exe"
       ToolName : string
+      /// Default: false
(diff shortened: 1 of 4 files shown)


src/app/FakeLib/UnitTest/NUnit/Common.fs
```diff
@@ -34,29 +34,53 @@ type NUnitDomainModel =
         | DefaultDomainModel -> ""
         | SingleDomainModel -> "Single"
         | MultipleDomainModel -> "Multiple"
-/// Parameter type for NUnit.
+
+/// The [NUnit](http://www.nunit.org/) Console Parameters type.
+/// FAKE will use [NUnitDefaults](fake-nunitcommon.html) for values not provided.
+///
+/// For reference, see: [NUnit-Console Command Line Options](http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6.4)
 type NUnitParams =
```
neoeinstein: In my PR, #800, making changes and documentation improvements to the xUnit tasks, I put the default values as documentation for the defaults value. Having it here seems wrong as a record doesn't actually have any defaults. So, that would be my recommendation.

LimoWanKenobi: Yes, it looks wrong. I will change it to the same style as your comments to xUnit then.

forki:
very very cool. Thx

PR #804, opened 2015-05-19
Added new function to WiX helper for retrieving file id from wixDir output

DigitalFlow:
This is the first implementation of my function for retrieving a file id from a wixDir string.
It takes the wixDir string and a regex to match the executable name, which you are trying to get the id from.
It returns the first match for the given regex out of the wixDir string.
I'll add a test case soon.


src/app/FakeLib/WiXHelper.fs
```diff
@@ -77,6 +78,24 @@ let rec wixDir fileFilter asSubDir (directoryInfo : DirectoryInfo) =
             directoryInfo.Name dirs compo
     else sprintf "%s%s" dirs compo
 
+/// Retrieves the file id of the first file in WiXString, which name matches fileRegex
+/// ## Parameters
+///  - `wiXString` - The directory string which was generated by wixDir
+///  - `fileRegex` - Regex which matches the file name
+///
+/// ## Sample
+/// let directoryString = wixDir (fun file -> true) true (DirectoryInfo directoryWithFilesForSetup)
```
forki: I think the sample code needs to be indented like https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/ZipHelper.fs#L165

forki:
very cool. only a small suggestion for the sample

DigitalFlow:
I fixed the line indentation in the sample and added a test case, not sure if it is at the right location, because I can't open FAKE correctly as it seems...
Any suggestions where to put the tests?

forki:
You can't open?

DigitalFlow:
When trying to run the "getFileIdFromWiXString" function in the test, I get a "getFileIdFromWiXString is undefined" error.

DigitalFlow:
Ok, fixed the build, my tests went into the Test.Fake.Deploy.Web project, I had to add Fake as a reference, forgot that.

DigitalFlow:
I wanted to get rid of the need to manually create a WiX XML.
So now there is a function which creates a standard WiX XML and various types for creating features, custom actions and install executions.
Just use the function for creating a WiXScript, create features, custom actions and so on as you wish and then pass them to the fillInWixScript function.
You can then run the WiX function without having to ever write a WiX XML yourself.

forki:
cool thanks.

PR #811, opened 2015-05-23
Add SemVerHelper.isValidSemVer(string)

nathanaeljones:
Allows build scripts to check if a git tag or input string is a semver.

### DIFF
--- a/src/app/FakeLib/SemVerHelper.fs
+++ b/src/app/FakeLib/SemVerHelper.fs
@@ -76,6 +76,15 @@ type SemVerInfo =
                     0
             | _ -> invalidArg "yobj" "cannot compare values of different types"
 
+
+let private SemVerPattern = "^(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)(?:-[\da-zA-Z\-]+(?:\.[\da-zA-Z\-]+)*)?(?:\+[\da-zA-Z\-]+(?:\.[\da-zA-Z\-]+)*)?$"
+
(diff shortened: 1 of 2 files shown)


src/app/FakeLib/SemVerHelper.fs
```diff
@@ -76,6 +76,15 @@ type SemVerInfo =
                     0
             | _ -> invalidArg "yobj" "cannot compare values of different types"
 
+
+let private SemVerPattern = "^(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)\.(?:0|[1-9][0-9]*)(?:-[\da-zA-Z\-]+(?:\.[\da-zA-Z\-]+)*)?(?:\+[\da-zA-Z\-]+(?:\.[\da-zA-Z\-]+)*)?$"
```
forki: shouldn't this be part of our normal Semver parser?

nathanaeljones: You're saying we should throw an exception during `parse` for parsable but technically invalid semver strings? Int32.Parse doesn't align with the spec, particularly if we think about culture. While we can improve parse (and possibly make it more strict), some things (like prohibiting leading zeroes) are more complex without a regex.

forki: Ok let's ask differently. Why do you need this new function? What's the use
case?

nathanaeljones: I use it to determine if a git tag is a semver. If it's a semver, then the build can take additional actions. Very useful for continuous delivery.

PR #812, opened 2015-05-23
Add AssemblyInfoFile.CreateCppCliAssemblyInfo

nathanaeljones:
Does not create static class like C# and VB counterparts.
Add test for CreateCSharpAssemblyInfo as well as CppCli


forki:
cool thanks

PR #817, opened 2015-05-26
WIP: Fix WiX API page

DigitalFlow:
I just discovered that the WiX API page has some missing type definitions and that samples are often rendered incorrectly.
I'm currently fixing this, I'll report when I'm done.


DigitalFlow:
Ok, I fixed the sample code.
Would you leave the visibility of the functions which contain the default parameters as they are, or would you like them to be internal / private?
I saw that some similar functions are public in NuGet helper, what's your best practice in here?

forki:
I make most functions public since you never know. But it's OK for now

DigitalFlow:
Ok, in this case I'm done, thanks.

PR #821, opened 2015-05-28
CMake support improvements

bbenoist:
Here are two new commits which includes improvements on the CMake support in FAKE:

* Commit f35c76d contains a fix for a typo I made.
* Commit b7193ed mostly aims to lower the CMake types in the HTML documentation. You can ignore it if you do not agree with the namespace changes.


forki:
Good idea.

PR #823, opened 2015-05-31
Clarify backward dependency operator

Vextasy:
I think that the meaning of the backward dependency operator was the wrong way round in the code comment. This change reverses the meaning.


Vextasy:
Apologies about the two merge pull requests that got included in this pull request. My intention was to just include the final 'Clarify backward dependency operator comment' commit. It's only a 2 line change - it might be simplest for you to take a look at it and manually process it (or not).

forki:
Thx

PR #824, opened 2015-06-01
Allow test nUnit test assemblies containing SetupFixture attributes be compatible with NUnitParallel

flashcurd:
nUnit unit test assemblies that contain a SetupFixture attribute will be flagged in the test report as a "SetUpFixture" test-suite type. e.g.

&lt; test-suite type="SetUpFixture" name="assembly_name_here" executed="True" result="Success" success="True" time="1.00" asserts="0" &gt;

This change will update the NUnitParallel test report merging so that "SetUpFixture" types will be included in the merged test report xml.


forki:
thx

PR #827, opened 2015-06-09
Re-added internal class generated for AssemblyInfo.vb

amazingant:
Was removed in 63185e7cded200f73e75dedcf5e0afe21f20b953 for #673 due to errors building. Contrary to the prior report, classes _can_ be added to `AssemblyInfo.vb`.

Two issues resolved here are:

1. In a VB.NET project named `TestApp`, using `Namespace System` in user-code results in a namespace named `TestApp.System` being created, which causes the System namespace to be hidden. Removed the namespace declaration around the class to resolve this.

2. As pointed out in #673, the class definition had no name. Added the name used in the C#/F# code.


forki:
@Vidarls could you please review? Thx to you both

Vidarls:
Looks good to me :)

forki:
cool thx

forki:
released in 3.35

amazingant:
Thanks, @forki :)

PR #836, opened 2015-06-23
Fix for single * glob not working.

voiceofwisdom:
Fixes [Paket issue #884](https://github.com/fsprojects/Paket/issues/884)


forki:
very cool. thx

PR #838, opened 2015-06-23
Added caching of compile regex for globs.

voiceofwisdom:
Should speed up handling large amount of file changes. Uses a concurrent dictionary, so there shouldn't be any issue with concurrency issues.


forki:
nice

PR #854, opened 2015-07-06
Add BuildConfig/TemplateFile options to PaketHelper's Pack command

amazingant:
The `MSBuildHelper` module allows building configurations other than Release, and the `NuGetHelper` module allows consuming them by specifying a build configuration in the `NuGetParams.Properties` field. Since the `PaketHelper` module currently doesn't offer any way to make this work, I've added paket's `pack buildconfig` option in as a new field on `PaketPackParams`. Added paket's `pack templatefile` option in as well, since I'm already under-the-hood and it may help someone else.


forki:
thx

PR #857, opened 2015-07-08
VB6 dependency updater

Vidarls:
Added task to automatically update VB6 project files to match
actual version of referred assemblies.

Makes managing .net dependencies for a VB6 app using something like Paket
much simpler.


forki:
cool. thx

PR #858, opened 2015-07-08
Add a UserInputHelper to allow interactive input

rneatherway:
The intended use case is to prompt for github credentials when doing a release, rather than storing them in non-version controlled file on disk.


forki:
cool

PR #859, opened 2015-07-08
Added caching of compiled build scripts.

### LINKED ISSUE #646: Add optional precompiling of build script
I've tried to precompile build script into executable and found that running executable saves me  about 5-7 seconds. Build script isn't being changed often, so i'd like to see option `--precompile-build-script=<path_to_executable>` with approximately following behavior (in some pseudocode):
```
if (not dirname(path_to_executable).exists()) {
  mkdir_p(path_to_executable)
}
if (not path_to_executable.exists()) {
  exec("fsharpc", [buildScriptPath, "--out:"+path_to_executable])
}
if (modification_time(path_to_executable) < modification_time(buildScriptPath)) {
  exec("fsharpc", [buildScriptPath, "--out:"+path_to_executable])
}

setPassedFakeOptionsInExecEnvironment()

try {
  exec(path_to_executable)
} catch (CanNotExecute) {
  exec("mono", [path])
}
```
---

Right now I'm using external script for precompiling.

It's necessary to use `#I` in build script to let `fsharpc` find FAKE libs without passing any option to compiler:
```fsharp
#I @"packages/FAKE/tools"
#r @"FakeLib.dll"
```
Then this script should work (create it in the same dir as build script named `build.fsx`, compiled executable is placed in `build/` subdirectory):
```sh
#!/bin/sh

THIS_SCRIPT_DIR="$(realpath "$(dirname "$0")")"

# Define invocation of command binaries.
FSHARPC="${FAKE_FSHARPC:-fsharpc}"
MONO="${FAKE_MONO:-mono}"
STAT_MOD_TIME="stat --format=%Y"

# Define path to build script and resulting executable.
BUILD_SCRIPT="${THIS_SCRIPT_DIR}/build.fsx"
COMPILED_BUILD_SCRIPT_DIR="${THIS_SCRIPT_DIR}/build"
COMPILED_BUILD_SCRIPT="${COMPILED_BUILD_SCRIPT_DIR}/build.exe"

# Invoke mono with including FAKE libs in MONO_PATH.
mono_with_fake_libs () {
  MONO_PATH="${THIS_SCRIPT_DIR}/packages/FAKE/tools:${MONO_PATH}" \
  "${MONO}" "$@"
}

# Get modification time (in seconds) of passed path.
modtime () {
  local STAT_OUTPUT="$(${STAT_MOD_TIME} "${1}")"
  echo "${STAT_OUTPUT#=}"
}

recompile () {
  ${FSHARPC} "${BUILD_SCRIPT}" --out:"${COMPILED_BUILD_SCRIPT}"
  if [ $? -gt 0 ]; then exit 1; fi
  mono_with_fake_libs --aot -O=all,-shared "${COMPILED_BUILD_SCRIPT}"
}

# Create output directory.
[ -d "${COMPILED_BUILD_SCRIPT_DIR}" ] || mkdir "${COMPILED_BUILD_SCRIPT_DIR}"

if [ ! -f "${COMPILED_BUILD_SCRIPT}" ]; then
  recompile
elif [ "$(modtime "${BUILD_SCRIPT}")" \
       -gt "$(modtime "${COMPILED_BUILD_SCRIPT}")" ]; then
  recompile
fi

mono_with_fake_libs "${COMPILED_BUILD_SCRIPT}"
```

When build script includes `RunTargetOrDefault "TargetName"`, running executable will invoke default target. FAKE options are passed to executable as environment variables:
```sh
$ target=Build build/build.exe
```
```sh
# env is required
# because `single-target` is not directly accepted by shell as a variable name
$ env target=RunTests single-target=true build
```

voiceofwisdom:
Caches the script file to a compiled dll when first run. Uses crc32 to hash the script, prevents using old dll when the script file has changed. Also recursively follows `#load`.

Takes 3312 milliseconds to interpret script without saving cache.
Takes 3326 milliseconds to interpret script and save cache.
Takes 260 milliseconds to run cached dll.

Fixes #646, and makes #645 less needed.

### DIFF
--- a/src/test/Test.FAKECore/FSIHelperSpecs.cs
+++ b/src/test/Test.FAKECore/FSIHelperSpecs.cs
@@ -167,64 +167,6 @@ namespace Test.FAKECore
 
                 included.ShouldEqual(new string[] { "justname", "./relative/path", "C:/absolute/path" });
             };
-        //It should_load_assembly =
-        //    () =>
-        //    {
-        //        var scriptPath = (Path.GetTempFileName() + ".fsx").Replace("\\", "/");
-
-        //        var script = 
-        //                "#r \"HashLib\"\nlet " +
-        //                "hasher = HashLib.HashFactory.Checksum.CreateCRC32a()\n" +
-        //                "printf \"%s\" (hasher.ComputeString(\"foo\").ToString())";
-        //        File.WriteAllText(scriptPath, script);
-
-        //        RunExplicit(scriptPath, "", true).ShouldEqual(
-        //            "Cache doesnt exist" + nl + 
-        //            "D9DD4FEE" + nl + 
-        //            "Saved cache" + nl);
-
-        //        RunExplicit(scriptPath, "", true).ShouldEqual(
-        //            "Using cache" + nl +
-        //            "D9DD4FEE");
-        //    };
     }
 
-    //public class when_running_the_fake_cli
-    //{
-    //    static string RunExplicit(string scriptFilePath, string arguments)
-    //    {
-            
-    //        var process = new System.Diagnostics.Process();
-    //        process.StartInfo.FileName = ".\\FAKE.exe";
-    //        process.StartInfo.Arguments = "\"" + scriptFilePath + "\" " + arguments;
-    //        process.StartInfo.UseShellExecute = false;
-    //        process.StartInfo.RedirectStandardOutput = true;
-    //        process.StartInfo.WorkingDirectory = Directory.GetCurrentDirectory();
-    //        process.StartInfo.RedirectStandardError = true;
-    //        process.Start();
-    //        //* Read the output (or the error)
-    //        string output = process.StandardOutput.ReadToEnd();
-    //        Console.WriteLine(output);
-    //        string err = process.StandardError.ReadToEnd();
-    //        //Console.WriteLine(err);
-    //        process.WaitForExit();
-    //        if (process.ExitCode != 0)
-    //        {
-    //            throw new Exception("Process exited with code " + process.ExitCode + ". Error: \n" + err);
-    //        }
-    //        return output;
-    //    }
-
-    //    static string Run(string script, string arguments) {
-    //        var scriptFilePath = Path.GetTempFileName() + ".fsx";
-    //        File.WriteAllText(scriptFilePath, script);
-    //        var result = RunExplicit(scriptFilePath, arguments);
-    //        File.Delete(scriptFilePath);
-    //        return result;
-    //    }
-
-        
-    //    It should_print_foobar =
-    //        () => Run("printfn \"foobar\"", "").ShouldEqual("foobar");
-    //}
 }
--- a/src/test/Test.FAKECore/FSIHelperSpecs.cs
+++ b/src/test/Test.FAKECore/FSIHelperSpecs.cs
@@ -1,0 +1,188 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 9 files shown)


voiceofwisdom:
Requires fsharp/FSharp.Compiler.Service#365. It has been merged, don't know when it will be released.

voiceofwisdom:
Implementing support for #r'd DLLs looks to be all that's missing. Only thing that's keeping my build scripts from running. Will work on fixing that.

voiceofwisdom:
Added support for build scripts with `#r` and `#I` directives. This now runs build scripts generated by FSharp.ProjectScaffold.

forki:
very very cool. is this ready?`whyt about the appveyor build?

src/test/Test.FAKECore/FSIHelperSpecs.cs
```diff
@@ -0,0 +1,230 @@
...
+                var script = 
+                    "#r \"justname\"\n" + 
+                    "#r \"./relative/path\"\n" +
+                    "#r \"C:/absolute/path\"";
+
+                var included = FSIHelper.getIncludedAssembly(script);
+
+                included.ShouldEqual(new string[] { "justname", "./relative/path", "C:/absolute/path" });
+            };
+        //It should_load_assembly =
```
forki: why is this commented?

voiceofwisdom: I was originally parsing out the included assemblies in the script files, that was the test for it. I switched to saving a list of all loaded assemblies paths and loading those on next run. Made the test not needed.

forki: Ok so could you please remove these?

voiceofwisdom:
Removed that dead code.

voiceofwisdom:
The appveyor build seems to be swallowing the exception that occured. Any help with that?

forki:
image

if you scroll up a bit you can see the test failure

forki:
I fixed the tests, but now I have issues with FAKE in fsreveal:

    #I @"packages/FsReveal/fsreveal/"
    #I @"packages/FAKE/tools/"
    #I @"packages/Suave/lib/net40"
    
    #r "FakeLib.dll"
    #r "suave.dll"
    
    #load "fsreveal.fsx"

image

voiceofwisdom:
Ah, didn't realize that `#load` took into account `#I`. I can add handling for that.

forki:
yes please send pull request. I can then update the FAKE 4.0 alpha package. I think this feature deserves it's own major version.

voiceofwisdom:
Agreed, it has enough potential weirdness hiding in it. Are the `trace` calls to output the state of the cache acceptable? I could not decide if they were meaningless noise when running FAKE in production.

forki:
yeah after thinking about it. 
The error is probably that we didn't increased the cache format version (JSON file format)

forki:
I think we will put these to verbose mode before we release 4.0

voiceofwisdom:
#859 has the fixes. Now that I have it working though, FSReveal is not going to save the cache until you stop watching changes. Felt a little odd.

forki:
mhm. I don't think that's an issue. or is it?

voiceofwisdom:
Not a huge issue, if you hit a key it finishes and caches it.

Thanks for the help getting this pulled in.

forki:
fake

forki:
@xavierzwirtz it seems we need a page in the docs which describes what happens here. Could you please create such a page?

xavierzwirtz:
Created in #872.

PR #863, opened 2015-07-14
Added MSBuild properties to AndroidPackageParams

MichalDepta:
Additional parameters passed to MSBuild when using AndroidPackage function can be useful e.g. for disabling StyleCop.


forki:
thx

PR #865, opened 2015-07-14
Expose Globbing.isMatch for use by external code.

mavnn:


forki:
but I can't release right now since the Caching functionality is broken. We need to fix this first

mavnn:
No problem!

PR #870, opened 2015-07-20
Add support for outputting NUnit style test result XML to Fake.Testing.XUnit2

andersosthus:
The xUnit2 Console runner supports outputting test results in NUnit style XML, but this option is not included in Fake.Testing.XUnit2.

This PR adds in the param ```NUnitXmlOutputPath ``` to ```XUnit2Params```, and works the same way as the other test output params.

This is also similar to Fake.Testing.XUnit, which already has this option:
https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/UnitTest/XUnit/XUnit.fs#L17


forki:
sounds good. thx

PR #877, opened 2015-07-23
Add support for VS2015 VSTest executable

satsuper:
Allows using the VS2015 VSTest console executable from the VSTest helper.


forki:
thx

satsuper:
Thanks for the quick merge

PR #894, opened 2015-08-01
Time stamps (WIP)

mfalda:
I have added a module for tracking files. Files can be tracked by checking their last write time or by computing their MD5 signature (only if their size is not greater than ~10MB). I have tried to document my module by using "///" comments and I also added a test script (described in a separate help page) to explain how I would use the module; I am not able to verify if the HTML document is generated correctly. Besides the doubts about the documentation I added a "WIP" because it is my first pull request and so I hope that it is all right.

P.S.: I had to make some slight changes to the files "AzureCloudServices.fs", "Globbing.fs" and "XUnit2.fs" in order to be able to compile FAKE; please check if them too.

### DIFF
--- a/FAKE.sln
+++ b/FAKE.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+VisualStudioVersion = 12.0.40629.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{0F0488CC-2580-4C07-9E16-3997480F0221}"
 	ProjectSection(SolutionItems) = preProject
@@ -41,7 +41,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "help", "help", "{564BD198-0
 		help\commandline.md = help\commandline.md
 		help\contributing.md = help\contributing.md
 		help\create-nuget-package.md = help\create-nuget-package.md
-		help\customtasks.md = help\customtasks.md
 		help\deploy.md = help\deploy.md
 		help\fsc.md = help\fsc.md
 		help\fxcop.md = help\fxcop.md
@@ -52,6 +51,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "help", "help", "{564BD198-0
 		help\specifictargets.md = help\specifictargets.md
 		help\teamcity.md = help\teamcity.md
 		help\templates\template-project.html = help\templates\template-project.html
+		help\timesignatures.md = help\timesignatures.md
 		help\typescript.md = help\typescript.md
 		help\users.md = help\users.md
 	EndProjectSection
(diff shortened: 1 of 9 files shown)


xavierzwirtz:
Sure, iirc the crc32 calculation I'm using in the cache will end up being faster than md5 though.

forki:
ok let's rephrase that: can we create 1 common module for such things? and
it would be cool if we dogfood that module with the caching stuff. Maybe
you two guys have ideas for that.

2015-08-10 18:06 GMT+02:00 Xavier Zwirtz <notifications@github.com>:

xavierzwirtz:
I take it back, according to [this](http://stackoverflow.com/questions/996843/when-is-crc-more-appropriate-to-use-than-md5-sha1/10172245#10172245) the difference between crc32 and md5 is negligible. I have no problem switching to using it for the cache.

However, you said you are using last write time as well. I have heard, and experienced, windows being very inconsistent about writing that attribute correctly. I'm afraid if we switch to that the cache will intermittently fail to detect changes.

mfalda:
Good idea to move it to a lower level in the FAKE architecture! I also considered to configure the module  as a computational expression, however by integrating it in the FAKE cache it will be even more transparent to the user.
I started to develop this module for myself because it was not clear to me whether FAKE keeps track of the file it builds; in the examples I see in the site I see that it can invoke msbuild, however a build system should be independent.
I used the write time as an easy starting point, but soon I realized that a checksum would be more reliable, although slower, and so I implemented MD5 too: If you say me that crc32 is ok and fast then timestamps can be omitted.

xavierzwirtz:
These were the start times of the app after adding caching:

Takes 3312 milliseconds to interpret script without saving cache.
Takes 3326 milliseconds to interpret script and save cache.
Takes 260 milliseconds to run cached dll.

Its doing a crc32 validation of all loaded scripts on every run, and seems plenty fast.

mfalda:
I am still studying FAKE architecture, however I have in mind a possible implementation of the new policy (based on crc32), and I will try it in the next weeks.

If I am not wrong, in the file "TargetHelper.fs" there is the function "runSingleTarget" that appears to me to be the core of the exec mechanism. In practice, the strategy could be:

1. Identify a point in which "initTS" could be invoked.
2. Understand whether are there other mailboxes around (I use a message passing strategy to serialize writes in the config file). In any case I think that they could co-exist.
3. Modify the (possibly previously identified) function(s) responsible for the execution of the targets and enrich it (or them) with the crc32 checks.
4. Decide how to deal with the final results (in my example I simply checked for the existence of the unique final result, but it could also have been altered, therefore a crc32 check is still needed).
5. Find where the final results are currently checked: The rebuilding of the last step(s) has to be forced in the presence of a changed result. I have to understand what happens if, for example, I delete a file belonging to the last target.

```F#
/// Runs a single target without its dependencies
let runSingleTarget (target : TargetTemplate<unit>) =
    try
        if errors = [] then
            traceStartTarget target.Name target.Description (dependencyString target)
            let watch = new System.Diagnostics.Stopwatch()
            watch.Start()
            target.Function() // <= run this function only if it is not up-to-date
            addExecutedTarget target.Name watch.Elapsed
            traceEndTarget target.Name     
    with exn ->
        targetError target.Name exn
```

mfalda:
I think that what I have in mind would be a too heavy structural change with respect to current FAKE architecture: Now targets can only depend on other targets, while my idea is more similar to a file-based build system (for example [Python doit](http://pydoit.org/) has both); when FAKE will add this kind of logic, cached signatures could also become more integrated. Meanwhile I think that the simple functions I have proposed can be sufficient ("!!" operator, for example, is always used inside targets, not outside them, as a file-based build system would allow), at least for my aims. I have implemented also a few functions for CRC32 signatures.
If you think that my module could be useful, feel free to improve it and to add it to the official FAKE libraries.

matthid:
@mfalda Not sure if I followed everything but it sounded like some deep changes to the FAKE architectur. With FAKE 5 we are decoupling a lot of stuff and we might be able to support your scenario now. Feel free to check it out on my fork (at the moment): https://github.com/matthid/FAKE/tree/coreclr and feel free to send a PR there.

matthid:
Closing this due to inactivity. Please consider contributing this again as FAKE 5 module! Thanks for your work on FAKE and sorry for the HUGE delay :(

Please let me know if you need help porting this to netcore.

PR #895, opened 2015-08-02
Skip retry logic on Mono where it causes a crash

rneatherway:
Reference: https://github.com/fsharp/fsharp/issues/440


forki:
we already have a isMono function. And is there a way to work around this?

rneatherway:
This script file doesn't reference any of the FAKE assemblies, so I don't think I can access `isMono`. I have tried to find other simple ways to express this, but they cause the same NullReferenceException.

PR #906, opened 2015-08-12
Pass OutputDataReceived to logfn instead of trace

koenmetsu:


forki:
thx

inosik:
Seems like this broke the build.

koenmetsu:
Sorry about that, should be logfn with a format or just log, not sure which one is better. I'm not in a position to test this until tomorrow morning CEST, so feel free to revert or fix if urgent.

inosik:
Actually I was about doing something else, but this change didn't let me compile. I reverted it for now in my local repo, but it isn't urgent, so I can wait.

forki:
Will fix tomorrow

koenmetsu:
Fixed in https://github.com/fsharp/FAKE/pull/907

PR #908, opened 2015-08-13
Fixed WatchChanges not properly removing subdirectories from watch list

daniel-chambers:
In the scenario where `WatchChanges` is used with multiple includes, if one include is a subdirectory of another, WatchChanges will not watch the correct directory.

For example:
```fsharp
use watcher = 
    !! "parent/*.*" ++ "parent/child/*.*"
    |> WatchChanges (fun changes ->
        tracefn "%A" changes)
```
`/parent/child` will be watched (incorrectly), instead of `/parent`.

This PR fixes this issue and refactors the code a little in order to implement a couple of tests to ensure the correct behaviour when eliminating subdirectories from the watch list.

### DIFF
--- a/src/app/FakeLib/ChangeWatcher.fs
+++ b/src/app/FakeLib/ChangeWatcher.fs
@@ -45,7 +45,7 @@ let WatchChanges (onChange : FileChange seq -> unit) (fileIncludes : FileInclude
     let dirsToWatch = 
         dirsToWatch |> Seq.filter (fun d -> 
                            dirsToWatch
-                           |> Seq.exists (fun p -> p.StartsWith d && p <> d)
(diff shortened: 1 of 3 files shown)


forki:
could you please look at the build error?

daniel-chambers:
Damn, it seems like a failure when building on Mono because of paths in the tests. Do you guys have a way of automatically handling the difference between \ on Windows and / on *nix?

forki:
there should be a normalizepath function somewhere in the repo. just call that

daniel-chambers:
Fixed :)

I used an internal function from Globbing, which makes me a little uncomfortable, but it was the only one I found.

PR #914, opened 2015-08-19
Logging improvement in Fake.Deploy

MorganPersson:
Log to file when running as a service + errors to eventlog.
Log to console when running as console app


forki:
cool thx.

PR #917, opened 2015-08-21
Add date to ReleaseNotes type definition

amazingant:
Adds a new "Date" field as a `DateTime option` to the `ReleaseNotes` record type, and performs simple parsing of dates from the complex release notes format. The regex I used finds dates in the format of `YYYY-MM-DD`, with the option to use any of `- /.` as the separators, but I pass the matched result to `DateTime.TryParse`. Should be able to change the regex (or add additional regexes) to match other date formats, but really, who doesn't use ISO-8601? :wink:

### DIFF
--- a/src/app/FakeLib/ReleaseNotesHelper.fs
+++ b/src/app/FakeLib/ReleaseNotesHelper.fs
@@ -75,7 +75,7 @@ let parseVersions =
         assemblyVersion, nugetVersion
 
 let parseDate =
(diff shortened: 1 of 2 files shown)


forki:
very nice.

amazingant:
Sorry, should have thought to include a backwards-compatible constructor. Thanks for cleaning up after me :grimacing:

forki:
No worries. I'm breaking things in FAKE for over 5 years now ;-)

PR #918, opened 2015-08-22
Add "or a package name" to NuGet install doc

Kazark:
I tested this: while `NugetInstall` does work with a `packages.config`,
it also just works with a package name, e.g. `NUnit.Runners`; in fact,
despite the fact that the documentation calls the parameter `packageFiles`,
it is called `packageName` in the code.

However, I haven't been able to regenerate the documentation yet to
confirm that I understand how the documentation works, because I'm on
Linux. I'll have to try on Windows, unless someone more knowledgeable
can just confirm that this will update the API docs.


forki:
thx a lot

Kazark:
@forki Thank you! This is a great tool.

PR #920, opened 2015-08-23
Soft dependencies

jlewicki:
This PR adds the soft dependency operators ?=> and <=?. 

If B has a soft dependency on A, that means that if A and B are part of a dependency graph, A must be run before B.  But B does not require that A be run.

Soft dependencies can help address this issue raised here:
http://stackoverflow.com/questions/21334283/fake-patterns-to-separate-target-build-order-from-dependencies


forki:
live

very very cool

PR #921, opened 2015-08-23
RunTargetOrListTargets()

ctaggart:


ctaggart:
Only today did I discover the `--list-targets` option.
http://fsharp.github.io/FAKE/specifictargets.html

image

It would be even better if `listTargets()` included the documentation for the target.
image

forki:
cool. thx

PR #927, opened 2015-08-24
Deployment configurable timeout

MorganPersson:
There is a hardcoded timeout of 20 minutes, unfortunately I need more than that right now.


forki:
could you please check the red build?

MorganPersson:
I will fix it.

forki:
thx

PR #928, opened 2015-08-27
Add support for new Xunit2 runner -noappdomain flag

jhamm:
This flag exists in versions starting in 2.1beta4. Before executing the test process with the parameter NoAppDomain = true, this target will detect if the xunit console runner supports this flag. If it does not, the
target will set the parameter to false.
This should allow this target to work on any version of xunit2 console runner.


forki:
very cool. thx

After a while we should remove the discover step and just assume it's there.

PR #930, opened 2015-08-27
Better user input helper

vbfox:
I was using the `UserInputHelper` and noticed a few problems and I also couldn't resist adding a few features.

Bugs fixed :
* In the previous code backspace moved back in the input but didn't remove
  the previous character from screen.
* Backspace could move before the starting point of the text, visually
  erasing the prompt. It is now impossible.
* Previously getUserInput wasn't generating a new line after the prompt but
  getUserPassword was doing it. The behavior is now unified (They both generate one)

Features added
* The prompt now display in white while the value entered by the user is
  gray (as before).
* Password prompts now give a visible feedback in the
  form of '*' characters instead of nothing.


forki:
thx

PR #931, opened 2015-08-28
Issue 929: 

jlewicki:
My previous commit attempted to use the same ordering algorithm (determineBuildOrder) for both serial and parallel builds. The ordering was self-consistent, but was not the same as the ordering reported by PrintDependencyGraph.

I restored the earlier behavior of using determineBuildOrder function only for parallel builds, and the order targets are executed in serial builds now matches that reported by PrintDependencyGraph


forki:
thx for the fast fix!

utopius:
Awesome, thx for fixing.

PR #932, opened 2015-08-28
Fixed bug where first 1024 bytes were uploaded using FTP

gareth-evans:
Only the first 1024 were being uploaded because the binary reader wasn't being passed to the recursive function.


forki:
ouch.

PR #939, opened 2015-09-04
HockeyAppHelper download team restriction

thomasvdb:
Added setting to HockeyAppHelper to restrict app downloads to specific teams according to the API documentation of HockeyApp (http://support.hockeyapp.net/kb/api/api-versions)


forki:
thx a lot

PR #942, opened 2015-09-08
Don't retry too much in Octokit.fsx

vbfox:
This PR changes 2 things in `OctoKit.fsx`:

* It avoid retrying hundred of times when chaining operations (4 successive operations like in project scaffold was retrying more than 200 times)
* It doesn't even retry if the error is an authorization error (Invalid password, login require 2-factor auth,  token can't do that, ...) as there is no chance that retrying changes the ouput.

### DIFF
--- a/modules/Octokit/Octokit.fsx
+++ b/modules/Octokit/Octokit.fsx
@@ -15,6 +15,12 @@ type Draft =
 
 let private isRunningOnMono = System.Type.GetType ("Mono.Runtime") <> null
 
+/// A version of 'reraise' that can work inside computation expressions
+let private captureAndReraise ex =
+    System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(ex).Throw()
+    Unchecked.defaultof<_>
+
+/// Retry the Octokit action count times
 let rec private retry count asyncF =
     // This retry logic causes an exception on Mono:
     // https://github.com/fsharp/fsharp/issues/440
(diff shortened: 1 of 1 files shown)


forki:
thx a lot

forki:
I had to revert this for now, since it breaks my release process somehow:

image

anay ideas?

vbfox:
First impression is that it's a normal Http error and the code failed after 4 retries but I'll test it a little more to see if I can reproduce.

I'll send a PR before that adding logging to retries without any of the things I changed there so we can see how many times a normal request is retried (like NuGet push is doing). Might help with a diagnostic.

forki:
yes seems even if I revert I get issues. At least your version catches the exception and fails properly.
So I reapply your chanegs and need to investigate

forki:
/cc @shiftkey @haacked did you change something in Octokit? Always getting 

     System.Net.Http.HttpRequestException: Error while copying content to a stream.

Haacked:
@forki Nothing i can think of that would affect this. Have you tried to nail it down? What are you doing?

forki:
It seems it only happens on https://github.com/fsprojects/Paket/releases

my other release builds still seem to work. (it's always using the same approach). Maybe I hit some upload limit in Paket (we already have 833 releases)?

forki:
@Haacked this is now happening on more and more projects even when I don't update any Octokit related code. Seems to be an issue with the response from github API. Did it change?

Haacked:
@forki hmm, might want to contact Support: https://github.com/contact
I personally don't know of any change that would affect this.

forki:
ok thx. will try to create a plain C# + octokit repro tommorow.

shiftkey:
Did this end up being related to the fix made in https://github.com/octokit/octokit.net/pull/895?

Issue #941, opened 2015-09-08
FakeLib.ReportGeneratorHelper doesn't support 'Badges' report type


RMCKirby:
I've recently been playing around with integrating `OpenCover` and `ReportGenerator` into my hobby projects and they work great.

I also use FAKE to build my solutions and after searching the source discovered that `FakeLib` includes helpers for both these tools - awesome!

However, the `ReportGeneratorHelper` does not support a report type of 'Badges' - this was recently added as seen at https://github.com/danielpalme/ReportGenerator.

It would be nice to have the helper updated to support this feature.

PR #943, opened 2015-09-08
Extend ReportGeneratorHelper to add Badges report type

RMCKirby:
Implements the following feature: https://github.com/fsharp/FAKE/issues/941

I've no familiarity with `F#` so there's probably some things that don't adhere to best practice/could be cleaner...
I've also added a covering spec test as well as this was previously non-existent.


forki:
very cool.

PR #944, opened 2015-09-09
Added method to return whether or not a value exists for a registry key

dmorgan3405:
Found a need for this a few times, and thought it would be nice to have it in the core.


forki:
cool stuff. thx

Issue #940, opened 2015-09-07
Using xbuild for iOSBuild instead of mdtool


thomasvdb:
Is there a possibility to migrate to xbuild for building iOS projects instead of using mdtool?
Would there be any disavantages or reasons not to do it?

PR #945, opened 2015-09-10
iOSBuild relies on xbuild instead of mdtool

molinch:
iOSBuild relies on xbuild instead of mdtool (Xamarin pushes to use xbuild instead of mdtool).

This implies some changes for iOSBuild parameters:
- Configuration only contains the configuration now (Debug, Release, Adhoc, AppStore) and no longer the platform. This is an additional property called Platform (iPhone vs iPhoneSimulator)
- BuidlIpa true/false to indicate wether you want an IPA file to be produced

### DIFF
--- a/src/app/FakeLib/XamarinHelper.fs
+++ b/src/app/FakeLib/XamarinHelper.fs
@@ -37,18 +37,27 @@ type iOSBuildParams = {
     ProjectPath: string
     /// Build target, defaults to Build
     Target: string
-    /// Build configuration, defaults to 'Debug|iPhoneSimulator'
+    /// Build configuration, defaults to 'Debug'
     Configuration: string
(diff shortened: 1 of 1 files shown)


molinch:
@gshackles As said on Twitter this is the PR.

thomasvdb:
Great!
If the pull request is merged, the issue #940 can be closed.

forki:
thx a lot

thomasvdb:
Having a strange issue with the location of the IPA file.
If I specify the OutputPath to, for example, C:\Projects\iOS\ then all the sources are placed inside that folder during the build but the IPA file is placed one directory higher (C:\Projects in the example).

Any idea what the issue could be?

molinch:
Absolutely no idea, we don't use this parameter on our side. Does it behave correctly with android builds?

PR #949, opened 2015-09-11
Added instructions for installing FAKE on deployment server

darrelmiller:


forki:
Cool. Now I just have to add similar paragraph about paket, which makes things even better

forki:
Seems broken build is unrelated

PR #952, opened 2015-09-16
Specify type of 'UploadTestResultsXml' first argument

vbfox:
The argument was infered to be `'a` but the type should be `TestResultsType`.

It made both completion and the documentation at https://fsharp.github.io/FAKE/apidocs/fake-appveyor.html hard to read and understand.


forki:
thx

PR #954, opened 2015-09-17
Better unknow target handling & build.cmd

vbfox:
## Changes

TLDR:

* Used `SETLOCAL` in `build.cmd` to ensure that it doesn't touch the global environment
* Also made `build.cmd` handle build target names quoted or not as first arg
* Replaced all direct access to the target dictionary with the `getTarget` function that provide an nice error display.

## Why

I had a problem today where FAKE wasn't building some of our projects anymore, after searching the real reason was that I built fake in the same terminal a little bit before and that `build.cmd` set the environment variable `TARGET` to `"Default"` (With the quotes).

The problem with that is that while FAKE `build.cmd` pass it on the command line, and the extra quotes are removed from there our own build script doesn't and just call :

    RunTargetOrDefault "Default"

*(The fact that we also use a task named `Default` made that a lot harder to debug, took me time to understand who set this evironment variable)*

But `RunTargetOrDefault` will also use an environment variable  `TARGET` if it exists, except that it doesn't trim quotes :D

And then fake was failing with the not really helpful :

    System.Reflection.TargetInvocationException: Exception has been thrown by the target of an
    invocation. ---> System.Collections.Generic.KeyNotFoundException: The given key was not
    present in the dictionary.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,4 +1,5 @@
 @echo off
+SETLOCAL
 
 cls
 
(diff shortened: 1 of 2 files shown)


forki:
thx

PR #956, opened 2015-09-21
Script path cache bust

### LINKED ISSUE #951: FAKE caching and copying files
I just found a bit odd behavior which (I suspect) might be due to the new caching that FAKE does.

I had some slides, copied them to a new place and run the build script. I did not change my `build.fsx` (so I suspect the file had the same creation time and hash). When running FAKE, it would behave as if running the script from the old location (which has different `__SOURCE_DIRECTORY__` and so it was showing wrong slides).

As soon as I made a change in `build.fsx`, everything started working fine. Could this be a FAKE caching issue?

xavierzwirtz:
Invalidates the script cache if any loaded scripts path changes. Fixes #951


forki:
very cool

PR #962, opened 2015-10-01
Added TestFile (analogous to TestDir) to FileHelpers

MiloszKrajewski:
Added TestFile (analogous to TestDir) to FileHelpers

### DIFF
--- a/src/app/FakeLib/FileHelper.fs
+++ b/src/app/FakeLib/FileHelper.fs
@@ -343,6 +343,14 @@ let TestDir path =
     else 
         logfn "%s not found" di.FullName
         false
+        
+/// Checks if the file exists
+let TestFile path = 
+    let fi = fileInfo path
+    if fi.Exists then true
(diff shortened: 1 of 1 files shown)


src/app/FakeLib/FileHelper.fs
```diff
@@ -343,6 +343,14 @@ let TestDir path =
     else 
         logfn "%s not found" di.FullName
         false
+        
+/// Checks if the file exists
+let TestFile path =
```
forki: don't we have somethin like FIleExists`?

MiloszKrajewski:
We do not have FileExists (in FileHelper). We do have "File.Exists(...)" (in System.IO) though. On the other hand "TestDir" has been implemented even if we also have "Directory.Exists(...)". TestFiles, in my opinion, fits better FAKE DSL. They are all wrappers around something after all.

forki:
thanks

PR #963, opened 2015-10-02
Don't assume that mono is on the path for El Capitan

nosami:
It is not guaranteed to be in $PATH if System Integrity Protection is on (default)


forki:
thx

PR #965, opened 2015-10-06
dotCover runner for Xunit2

bjartwolf:
(Added support for running dotCover using the Xunit2 test runner. Couldn't find tests for similar libraries, so I haven't added any tests. Tried a little bit locally and it seemed to work for simple cases. I will use this at work, so I'm happy to test it out on more complex real-life scenarios.

Also, it seems there are some differences in how the different test runner helpers are structured, I couldn't really figure it out...


forki:
thx a lot

PR #968, opened 2015-10-07
type correction

yever:


forki:
this branch is usally updateed automatically. you don't need to send pull  requests. thx anyways

yever:
Noted.

Issue #946, opened 2015-09-10
Relative path issues with fake script compilation to .fake folder.


v-zubritsky:
Not sure, If it's a problem of fake or mine but nevertheless.
Suppose we've got next directories structure
$/
&nbsp;&nbsp;&nbsp;&nbsp;.nuget/
&nbsp;&nbsp;&nbsp;&nbsp;packages/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Fake/
&nbsp;&nbsp;&nbsp;&nbsp;cmd/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;run.cmd
&nbsp;&nbsp;&nbsp;&nbsp;fsx/
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;run.fsx

Where `run.cmd` is a fake script bootstrapper, which downloads fake to the `packages` folder with use of `.nuget/nuget.exe` and starts `../fsx/run.fsx` then.
And `run.fsx` has next content

    #I @"../packages/" 
    #r @"FAKE/tools/FakeLib.dll"

    open Fake
    open System.IO
    Directory.SetCurrentDirectory (__SOURCE_DIRECTORY__)

    Target "Default" (fun _ -> 
         trace "Running default target"
    )

    RunTargetOrDefault "Default"

After executing the script, empty folder `.fake` will be created inside `cmd` folder, while compilated `FSI-ASSEMBLY.dll` will be saved inside `fsx` one. Also, after all the targets will be completed successfully, we will get`System.IO.DirectoryNotFoundException` inside `Fake.FSIHelper.runFAKEScriptWithFsiArgsAndREdirectMessages` at `FSIHelper.fs#340` 

Everything will be ok without `Directory.SetCurrentDirectrory(__SOURCE_DIRECTORY__)`, but I want all the pathes used in script to be relative to the `fsx` files itself and not the `cmd` bootstrapper.

PR #969, opened 2015-10-07
Ensure FSI-ASSEMBLY.dll path exists

khanage:
fixes fsharp/FAKE#946 - relative paths seem to be confused between the path of the fake.exe and the path of the project.

I'm happy for someone who knows more than I do to do a deeper dive, but this seems adequate for what I assume is a temporary location.


forki:
cool. thx a lot

PR #972, opened 2015-10-09
DotCoverMSTest support

bjartwolf:
Added support for using DotCover with MSTest.

The original MSTest runner runs all test assemblies in a separate run. That might have some benefits, and I don't want to change this for the DotCover support without a proper discussion as it might break backwards compatibilty for users using MSTest with multiple assemblies. For DotCover it is quite nescessary, otherwise we create too many DotCover reports.

Tried it on five test assemblies with 1100 tests and it seems to work fine.

Will use both this and the XUnit2DotCover quite a lot going forward.


forki:
thanks!

Issue #966, opened 2015-10-06
Feature request: more RegistryHelper functions


yever:
I've been using the RegistryHelper to get some paths and I think that it would be helpful to have functions to enumerate subkeys and to access them. Those are provided by ``Microsoft.Win32``, but it would be nice to have FAKE wrappers for them.

I would suggest to wrap the following methods from ``Microsoft.Win32.RegistryKey``:
- ``GetSubKeyNames()``
- ``DeleteSubKey()``
- ``GetValueNames()``

Would you accept a pull request for this?

forki:
Sure.

yever:
Wow, that was quick :)

PR #976, opened 2015-10-13
RegistryHelper

yever:
Here are some additional registry function, as per #966.

I think the function ``createRegistrySubKey`` should be changed to follow the pattern of the other ones, but I didn't want to change the API without consulting.

It should:
1. ``use`` the key, since it's an IDisposable
2. have a signature with ``baseKey``, ``subKey`` and ``keyName``.

Additional cleanup should include renaming ``keyName`` to ``valueName`` or ``value`` in ``setRegistryValue`` and ``deleteRegistryValue``.

### DIFF
--- a/src/app/FakeLib/RegistryHelper.fs
+++ b/src/app/FakeLib/RegistryHelper.fs
@@ -5,7 +5,7 @@ module Fake.RegistryHelper
 open Microsoft.Win32
 
 /// Registry base keys.
-type RegistryBaseKey = 
+type RegistryBaseKey =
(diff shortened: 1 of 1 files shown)


forki:
thx. can you send another PR for the additional fixes? But please keep API backwards compatible.

yever:
You're welcome, @forki.
I'll send another PR with some cleanup. 

As for the API-changing fix, I still think it would make sense to change the signature of ``createRegistrySubKey``. Should I open an issue? Could this be put on hold until the next MAJOR version change?

PR #978, opened 2015-10-13
Add maintainers note to README

tpetricek:
We clarified the status of the repositories under "github.com/fsharp" at the last F# Foundation meeting.

The summary is that FAKE and F# Data are welcome to stay under the "fsharp" organization. We'd just like to add an explicit note to the README saying that those are independent projects.


forki:
thx.

PR #980, opened 2015-10-14
Cleanup RegistryHelper.fs

yever:
- Group functions meaningfully: (1) key, (2) value, (3) subkey.
- The paramater indicating a registry value name is called "name" rather than "value", following the convention of Microsoft.
- deleteRegistrySubKey follows the signature of createRegistrySubKey (that was there before): baseKey (HKEYLocalMachine, etc.) + subKey. This means that the subKey string must indicate the full path to the subkey (e.g. "SOFTWARE\\Microsoft\\VisualStudio").


yever:
@forki, here are the additional fixes we discussed in #976. This would close #966.

forki:
thx

Issue #935, opened 2015-09-01
Have default fsx file name for Fake.Deploy


koenmetsu:
Try calling a pre-defined .fsx file (eg: deploy.fsx) before falling back on the first matching file.

`let scriptFile = FindFirstMatchingFile "*.fsx" activeDir` (from [here] (https://github.com/fsharp/FAKE/blob/master/src/app/Fake.Deploy/DeploymentHelper.fs#L69))

This would allow us to have multiple fsx files under the root.

PR #981, opened 2015-10-15
Implemented Issue 935

MorganPersson:
This will prefer the following files
deploy.fsx
install.fsx
setup.fsx
over any other fsx file found. If all 3 files listed above exists it will be picked in the order listed above


forki:
thx #935 closed

PR #983, opened 2015-10-16
Added curl parameters ConnectTimeout and MaxTime and changed upload host.

molinch:
On our side we have a very slow network and all HockeyApp upload were failing in time-out.

So I added curl parameters ConnectTimeout and MaxTime. I also changed upload host to be upload.hockeyapp.net since it leads to faster uploads. You can see: http://support.hockeyapp.net/discussions/problems/33355-is-uploadhockeyappnet-available-for-general-use
Since we POST data we don't care about anything that could be cached by CloudFare.

### DIFF
--- a/src/app/FakeLib/HockeyAppHelper.fs
+++ b/src/app/FakeLib/HockeyAppHelper.fs
@@ -113,6 +113,12 @@ type HockeyAppUploadParams = {
 
     /// Release download status (can only be set with full-access tokens)
     DownloadStatus: DownloadStatusOption
+
+    /// Time in seconds to connect to HockeyApp backend
+    ConnectTimeout: int
+
+    /// Max time in seconds for the upload
+    MaxTime: int
 }
 
 /// The default HockeyApp parameters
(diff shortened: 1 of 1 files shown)


molinch:
Shell Exec now also takes into consideration the MaxTime parameter.

molinch:
Is any extra info needed? Could we have some news?

forki:
Did you look into the build errors?

molinch:
@forki  It complains about `AppVeyor was unable to build non-mergeable pull request` therefore for me it's not very clear what should be done

forki:
Also why don't you make the two parameters a timespan? This makes it easier for users to reference seconds vs minutes

forki:
Ok ignore the build errors

molinch:
I will use a TimeSpan then and add this commit

forki:
cool. just let me know when it's ready

forki:
ping

matthid:
Closing this due to inactivity. Please consider contributing this again!

PR #988, opened 2015-10-27
Added helper to support plain robocopy and mirroring.

tonikaihola:


src/app/FakeLib/RoboCopyHelper.fs
```diff
@@ -0,0 +1,38 @@
+﻿[<AutoOpen>]
```
forki: please remove autoopen. we try to get rid of these

tonikaihola:
Thank you for your comment forki.

Some changes:
* Removed AutoOpenAttribute
* Hid implementation details so that end users won't depend on RoboCopyOptions (until stable)
* Added exit code handling

forki:
cool thx

PR #990, opened 2015-10-29
Option ignore failing tests dot cover

bjartwolf:
When DotCover discovers failing tests in fails with errorcode -3. I added an option to ignore failing tests so that I can run it to completion and let for example TeamCity fail the build based on the test reports.


bjartwolf:
Forgot to mention that the defaults are set so that existing code should work as before (break on test errors)

forki:
thx

PR #993, opened 2015-11-01
add basic npm support

mastoj:
This small feature makes it really easy to run `Npm` task for build frontend projects. This also opens up for running `gulp`, `grunt` or some other frontend build tool by calling them from `Npm`. 

To get started on Windows:

Add the following to `build.cmd`:
```
NuGet.exe "Install" "Node.js" "-OutputDirectory" "packages" "-ExcludeVersion"
NuGet.exe "Install" "Npm.js" "-OutputDirectory" "packages" "-ExcludeVersion"
```

Then you can run task like:
```
Target "Web" (fun _ ->
    Npm (fun p ->
      { p with
          Command = Install Standard
          WorkingDirectory = "./src/FAKESimple.Web/"
      })

    Npm (fun p ->
      { p with
          Command = (Run "build")
          WorkingDirectory = "./src/FAKESimple.Web/"
      })
)
```

Haven't tried it on *nix yet, but it should work there as well since you can override the path to npm. I have also provided defaults for *nix and Windows.

### DIFF
--- a/src/app/FakeLib/FakeLib.fsproj
+++ b/src/app/FakeLib/FakeLib.fsproj
@@ -55,6 +55,7 @@
     <Compile Include="ServiceControllerHelper.fs" />
     <Compile Include="GuardedAwaitObservable.fs" />
     <Compile Include="ProcessHelper.fs" />
+    <Compile Include="NpmHelper.fs" />
     <Compile Include="AppVeyor.fs" />
     <Compile Include="TaskRunnerHelper.fs" />
     <Compile Include="Globbing\Globbing.fs" />
--- a/src/app/FakeLib/NpmHelper.fs
+++ b/src/app/FakeLib/NpmHelper.fs
@@ -1,5 +1,5 @@
 ﻿/// Contains function to run npm tasks
-module NpmHelper
+module Fake.NpmHelper
 open Fake
 open System
 open System.IO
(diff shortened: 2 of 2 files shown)


bjartwolf:
I was able to make this work on Windows like this, running tasks in Grunt etc
```fsharp
Target "RestoreLocalPackages" (fun _ -> 
     RestorePackageId (fun p -> { p with ExcludeVersion = true 
                                         ToolPath = "../.nuget/NuGet.exe"
                                         OutputPath = "packages" 
                                         Version = Some (new System.Version(4,0,0)) }) "Node.js" 
     RestorePackageId (fun p -> { p with ExcludeVersion = true 
                                         ToolPath = "../.nuget/NuGet.exe"
                                         OutputPath = "packages" 
                                         Version = Some (new System.Version(2,13,1)) }) "Npm.js" 
                                         
)


Target "PreBuild" (fun _ ->
    Npm (fun p ->
            { p with
                Command = Custom ("install grunt") })
    Npm (fun p ->
            { p with
                Command = Run ("fakeprebuild") })
    Npm (fun p ->
            { p with
                Command = Run ("faketest") })
)

Target "NpmInstall" (fun _ ->
    Npm (fun p ->
            { p with
                Command = Install Standard
                WorkingDirectory = "Nrk.Programspiller.Frontend" }))```

The trick which could probably be documented in a separate page about using Npm with Fake is that to run grunt tasks without installing node.js and grunt etc globally I had to do this:
```javascript
  "scripts": {
	"faketest": "Frontend\\packages\\Node.js\\node.exe -e \"require('grunt').tasks(['karma:ci', 'zip'], {color: false});\"",
	"fakeprebuild": "Frontend\\packages\\Node.js\\node.exe -e \"require('grunt').tasks(['publish'], {color: false});\""
  },```

forki:
please let me know when if this is ready to merge.

mastoj:
@bjartwolf do you think it is good enough to work? I think so, since it is really hard to know exactly how the users want to use npm. The helper tries to be helpful in that you can override where npm is. 

I can try to run it using a global installed version of node and npm since that is another use case as well as try it on osx. I'll try to do that later today.

mastoj:
I've now tried it with global `npm` in Windows and also in OSX and get it to run without problem. So if you think it looks good I think it is ready be merged.

forki:
very cool!

bjartwolf:
I see it is already closed but I agree and thanks, very useful for us!

PR #1000, opened 2015-11-09
Rename confusing parameter in FSI helper

iainnicol:
The parameter is a list of environmental variables, not command line
arguments.


forki:
LGTM thanks

PR #1004, opened 2015-11-15
added upload with longer timeout

detroitpro:
I am not sure if there is a better way of doing this, or anything in F#.

The upload of a 50MB webjob keep failing due to the default timeout.


forki:
thx

PR #1007, opened 2015-11-17
add tests and fix regex

albertjan:
Also adds some tests. Just to be sure. 

Followup to #1006 as that would not parse versions containing more then 5 characters.


albertjan:
@forki sorry about the first round, this should do the trick.

forki:
thx

PR #1008, opened 2015-11-17
Fix case on MsBuild LogFile option

mavnn:
This works all lower case with MsBuild, but silently fails to create log files with xbuild.


forki:
thx

PR #1009, opened 2015-11-17
Added initial implementation of DynamicsCRMHelper

DigitalFlow:
Hey all,

I'm currently implementing a new helper for Dynamics CRM.
The use of it is to be able to export solutions (update component containers) from Dynamics CRM and being able to store those solutions in a VCS or even import them before doing unit tests for ensuring that the system is in the correct state for doing the unit tests.
Tests as well as a possibility for downlevelling the solution archive into xml representation will follow.

Kind Regards,
DigitalFlow

### DIFF
--- a/src/app/FakeLib/DynamicsCRMHelper.fs
+++ b/src/app/FakeLib/DynamicsCRMHelper.fs
@@ -1,0 +1,198 @@
+﻿module Fake.DynamicsCRMHelper
+
+open System
+open System.Configuration
+open System.IO
+
+/// Specify which action Solution Packager should be invoked with
+type SolutionPackagerAction = 
+    | Extract
+    | Pack 
+     override s.ToString() =
+        match s with
+        | Extract -> "/a:Extract"
+        | Pack -> "/a:Pack"
+
+/// Specify Package Type for usage with Solution Packager
+type PackageType = 
+    | Unmanaged
+    | Managed
+    | Both
+    override p.ToString() =
+        match p with
+        | Unmanaged -> "/p:Unmanaged"
+        | Managed -> "/p:Managed"
+        | Both -> "/p:Both"
+
+/// Parameters for executing Dynamics CRM Helper functions
+type DynamicsCrmHelperParams =
+    {
+        /// Url of CRM Organization / Discovery Service URL if using AllOrganizations
+        Url : string
+        /// Username of user that should be used to connect to the CRM Organization. Leave blank to use default credentials
+        User : string
+        /// Password of user that should be used to connect to the CRM Organization. Leave blank to use default credentials
+        Password : string
+        /// TimeOut for each function. Set to a higher value, i.e. 60 minutes if using AllOrganizations
+        TimeOut : TimeSpan
+        /// Tool Directory where Solution Exchanger is stored
+        ToolDirectory : string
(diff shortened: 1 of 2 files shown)


DigitalFlow:
What I was wondering about:
Is it ok that I added the NuGet Packages for CRM SDK? I use them for connecting to the CRM services.

forki:
Aehm that's not a good idea

DigitalFlow:
Ok, so I'll remove those packages and connect differently. I wonder why the mono build on Travis is failing, while the AppVeyor build with MSBuild works...

DigitalFlow:
@forki: How sounds using Fsharp Data TypeProviders for connecting to the CRM to you? Seems if I'd need them as NuGet packages too however.
https://www.nuget.org/packages/FSharp.Data/2.2.5/
https://www.nuget.org/packages/FSharp.Data.TypeProviders/

forki:
we do we need to add packages at all?

DigitalFlow:
Ok so I checked my options and I don't see an alternative to the CRM packages that does not produce a lot of overhead for me. 
Shall we close the Pull Request in this case?

forki:
mhm. I think we should create an addition project then. We already have way to much stuff in this repo.

DigitalFlow:
All right then

forki:
sorry. Is there a command line version we can call from here?

DigitalFlow:
I just began developing a command line version, I'll check with my boss whether I am allowed to open source it. Are you thinking of making it a complete standalone tool, or a subproject of FAKE?

forki:
What we usually do is to create only a wrapper around the command line in fake (That stuff that calls the command line tool),

DigitalFlow:
Yes thats what I did for the SolutionPackager part, but I guess you mean for this whole module then.
I'll have to develop a command line version anyhow, I would just have to check with my boss for open sourcing it and writing a FAKE wrapper.

DigitalFlow:
Ok so it goes on. I am allowed to open source the tool.
The plan is to leave solution packager part of this pull request as it is, because it does not need any packages. For the importing and exporting I will write a small command line tool and a wrapper for FAKE so that we can use the wrapper and don't need any nuget packages in here.

DigitalFlow:
Ok, my command line tool is done for now: https://github.com/DigitalFlow/Dynamics-CRM-Solution-Exchanger
The wrapper is already working in this PR, I already set up a build in our CI which exports a solution from CRM every night and pushes it into our VCS.

DigitalFlow:
@forki: Can you imagine why the build is failing? Local builds run just fine, I didn't add any assemblies that would need to be referenced and the previous build passed at least for the previous commit, while I only changed an XML tag in the current commit...

forki:
seems unrelated. mono fails with mono 4.2 issue and appveyor with nuget.

src/app/FakeLib/DynamicsCRMHelper.fs
```diff
@@ -0,0 +1,143 @@
+﻿[<AutoOpen>]
```
forki: please remove the autoopen

DigitalFlow:
Ok, I removed the auto open tag.

DigitalFlow:
Ok, so I just added some documentation and some comments.
Do you need anything else for it being merged?

forki:
cool. thx

PR #1018, opened 2015-11-27
add functionality for deleting files and folders to FTP Helper

bogdangaliceanu:
While working on an automated deployment scenario, we noticed that the ability to delete stuff from an FTP server would come in handy


forki:
cool thx

PR #1020, opened 2015-11-29
Add BASH completion for FAKE tasks

nosami:
Adds case insensitive tab completion to BASH for FAKE tasks. 

I wasn't sure where to put the files for this. I tested this on both OSX (El Capitan) and Ubuntu 15.10. 

In Ubuntu, just running the install.sh should set things up. On OSX, the user will need to also insert bash-completion if they haven't already done so.

Maybe this could be made easier.

### DIFF
--- a/bash-completion/README.md
+++ b/bash-completion/README.md
@@ -1,0 +1,22 @@
+# BASH tab completion for possible FAKE tasks
+
+## Prerequisites
+
+If you have tab completion for other commands (such as git parameters) then you can skip to [Install](#install-completion-for-fake)
+
+##### OSX
+```bash
+brew install bash-completion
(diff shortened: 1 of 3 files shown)


cdrnet:
Excellent! Does the installer also work on Windows (assuming bash-completion itself is already set up e.g. by git-bash)?

vilinski:
cool stuff! Is it also usable with zsh?

nosami:
@cdrnet I haven't tested it with Windows. It can probably be made to work.

@vilinski No. It won't work with zsh as zsh uses a completely different system. I can work on it though.

Neftedollar:
This is great! Thank you!

Issue #1024, opened 2015-12-01
Jarsigner does not support override of the default SignatureAlgorithm and MessageDigestAlgorithm


Dexyon:
Currently the properties of the AndroidSignAndAlign do not allow an override of the 'Signature Algorithm' or 'MessageDigestAlgorithm'.

Overview of the current 'AndroidSignAndAlignDefaults'
```
let AndroidSignAndAlignDefaults = {
    KeystorePath = ""
    KeystorePassword = ""
    KeystoreAlias = ""
    JarsignerPath = "jarsigner"
    ZipalignPath = "zipalign"
```

I have added the follow two parameters via the defaults
```
    SignatureAlgorithm = "SHA1withRSA"
    MessageDigestAlgorithm = "SHA1"
```

Dexyon:
Already merged in the Master

PR #1025, opened 2015-12-01
Allow override of the signature algorithm and message digest algorithm

Dexyon:
Fix for issue discussed in 'Jarsigner does not support override of the default SignatureAlgorithm and MessageDigestAlgorithm'

Fix for: #1024


Dexyon:
Mmm looks like the Travis CI is failing more builds than mine..

PR #1033, opened 2015-12-07
only add relative path prefix if not rooted path

reidev275:
pull request for #1032 MSBuildHelper BuildWebsite fails with invalid OutDir path.


forki:
thx

PR #1036, opened 2015-12-09
[WP] Added more comfortable types to WiXHelper, flagged old ones obsolete

DigitalFlow:
Hey all, 

I did some slight changes to the WiXHelper, so that WiX Features, Custom Actions and Custom Action Executions don't get passed around as concatenated strings, but as sequences of their types.
The old types and functions using them were flagged as obsolete.

@forki: Two questions remaining:
1: I have a module called "Test.Fake.Deploy.Web.WixHelper" which resides in \src\test\Test.Fake.Deploy.Web. It does not seem to be included in the solution, where would this go?
2: I'd like to contribute the minimal working example of a WiX Target as section to the github.io page (https://fsharp.github.io/FAKE/). Where can I add this section? 
Edit: Ah found it, goes to /help.

Kind Regards

### DIFF
--- a/src/app/FakeLib/WiXHelper.fs
+++ b/src/app/FakeLib/WiXHelper.fs
@@ -52,6 +52,196 @@ let getFilesAsWiXString files =
     |> Seq.map (fileInfo >> wixFile)
     |> toLines
 
+/// WiX File Element
+type WiXFile =
+    {
+        Id : string
+        Name : string
+        Source : string
+    }
+    override w.ToString() = sprintf "<File Id=\"%s\" Name=\"%s\" Source=\"%s\" />"
+                                w.Id w.Name w.Source
+
+/// Defaults for WiX file
+let WiXFileDefaults = 
+    {
+        Id = "fi"
+        Name = ""
+        Source = ""
+    }
+
+/// Specifies whether an action occur on install, uninstall or both.
+type InstallUninstall = 
+    | Install
+    | Uninstall
+    | Both
+    override w.ToString() = 
+        match w with
+        | Install -> "install"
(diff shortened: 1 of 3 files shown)


DigitalFlow:
Don't really get, why "C:\projects\fake\src\app\FakeLib\FSharpFormattingHelper.fs(43,23): error FS0971: Undefined value 'command : string'" is failing. Didn't do any changes to this, will have to look at this...

DigitalFlow:
Ok so the new types work pretty well, I'm already using them in some projects.
The minimal working example markdown is done too, is it automatically added to the page index?
I still don't know however, whether the tests are fine in Test.Fake.Deploy.Web (I guess not)?
Rest should be fine.

src/app/FakeLib/WiXHelper.fs
```diff
@@ -588,10 +923,11 @@ let generateWiXScript fileName =
             <Media Id=\"1\" Cabinet=\"media1.cab\" EmbedCab=\"yes\" />
 
             <Directory Id=\"TARGETDIR\" Name=\"SourceDir\">
-              <Directory Id=\"ProgramFilesFolder\" Name=\"ProgramFiles\">
+              <Directory Id=\"@Product.ProgramFilesFolder@\" Name=\"ProgramFiles\">
```
mexx: I'm not experienced with WiX, but from other `Directory` entries, shouldn't the placeholder be put into `Name`?

mexx: Ok, got it, it's only the placeholder for `ProgramFilesFolder` or `ProgramFiles64Folder`, never mind.

PR #1040, opened 2015-12-11
Replace hard reference on gacutil path with automatic search for late…

shahaya:
…st gacutil.exe

The existing path "Microsoft SDKs/Windows/v8.0A/bin/NETFX 4.0 Tools/gacutil.exe" doesn't exist on a Win 10 + Visual Studio 2015 dev environment.
RegAsmHelper solves a similar problem by searching for the latest regasm tools path (using EnvironmentHelper).
This patch solves this analogously.


forki:
thx

PR #1048, opened 2015-12-21
Changes in Octokit.fsx

eiriktsarpalis:
This PR attempts to work around [a known issue](https://github.com/octokit/octokit.net/issues/963) in the Octokit.Net library, in which user-supplied timeouts are not passed to the underlying `HttpClient` object. This results in timeout errors when uploading github release assets that are sufficiently large.

I have also included a new `uploadFiles` method that can upload multiple assets in parallel.


forki:
thx

PR #1049, opened 2015-12-21
TC: Add open-close block helpers

tolu:
Added helper methods for TeamCity blocks

**Usage**
```fs
Target "MyTarget" (fun _ ->
  sendOpenBlock "MyTarget"
  // target code here
  sendCloseBlock "MyTarget"
)
```

This nifty little thing gets you from this:

Before

to this (targets in an expandable tree view):

After

**Futures**
As a next step it would be awesome to use this feature by default in the target output (when running in TeamCity) instead of going the long way round and polluting every target with `sendOpenBlock` / `sendCloseBlock`.

### DIFF
--- a/src/app/FakeLib/TeamCityHelper.fs
+++ b/src/app/FakeLib/TeamCityHelper.fs
@@ -25,6 +25,12 @@ let sendToTeamCity format message =
 let sendStrToTeamCity s = 
     if buildServer = TeamCity then postMessage (LogMessage(RemoveLineBreaks s, true))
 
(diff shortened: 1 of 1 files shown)


forki:
thx

dahlbyk:
Would it make sense to model this as `IDisposable`?

```fsharp
Target "MyTarget" (fun _ ->
  use sendBlock "MyTarget"
  // target code here
)
```

forki:
I think you should not use this in user code. It should be used in target and task helpers.

PR #1050, opened 2015-12-21
Replace SignHelper with SignToolHelper in FakeLib.fsproj

pdfforge:
Both files were included in the source code, but only the SignHelper was
included in the project file. As SignToolHelper provides support for the
same call syntax as the SignHelper, the SignHelper was removed in this
commit.


forki:
thx

PR #1051, opened 2015-12-21
Change SignToolHelper syntax to reflect common call syntax

pdfforge:
Until now, the call syntax of the SignToolHelper would have been
something like this:

     let signParams = {
         DevCertificate = cert
         Certificate = None
         FilesToSign = !! (buildDir @@ "**/*.exe")
         TimeStampUrl = None
     }

     Sign "packages/signtool" signParams

This does not go well with piping file sequences as it is used all
over FAKE. After this change, SignTool can be called like this:

     let signParams = {
         DevCertificate = cert
         Certificate = None
         TimeStampUrl = None
     }

     !! (buildDir @@ "**/*.exe")
       |> Sign "packages/signtool" signParams

From my point of view, this is more the "FAKE" way of doing things.

This is a breaking change though, as it changes the sytax of this helper and changes the SignParams type.

### DIFF
--- a/src/app/FakeLib/SignToolHelper.fs
+++ b/src/app/FakeLib/SignToolHelper.fs
@@ -21,8 +21,6 @@ type SignCert = {
 
 /// Parameters used for signing.
 type SignParams = {
-    /// The assemblies to sign
-    FilesToSign : seq<string>
     /// The dev certificate that will be used when the real certificate can not be found
     DevCertificate : SignCert
     /// The optional real certificate that will be used when it is found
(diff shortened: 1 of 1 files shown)


forki:
yes this would have been better. But since it will break A LOT of builds I think it's not a good idea to change it.

pdfforge:
Sorry for bothering again with this issue.

After some investigation, it happens that the file has never been in the build before (not even in the initial commit b21b39e). So in my opinion, this can safely be merged, as it had not been included.

If you don't agree, that's ok of course ;-)

PR #1053, opened 2015-12-21
added Alias and HiddenTarget.

pms1969:
- Alias allows dependencies trees to be constructed and run as if they were defined that way.
- HiddenTarget just leaves the Target out of the listTargets and PrintTargets, allowing for a cleaner target list that doesn't include things you might not want to run directly.

### DIFF
--- a/src/app/FakeLib/TargetHelper.fs
+++ b/src/app/FakeLib/TargetHelper.fs
@@ -45,7 +45,7 @@ let TargetDict = new Dictionary<_,_>()
 /// [omit]
 let Aliases = new List<_>()
 
-/// Alias List
+/// Hidden Targets List
 /// [omit]
 let Hidden = new List<_>()
 
@@ -278,15 +278,11 @@ let Alias name body =
     Target name body
     Aliases.Add name
 
-/// Creates an alias
+/// Creates a hidden target
 ///
 /// Example:
 ///
-/// Target "T1" DoNothing
-/// Target "T2" DoNothing
-/// Alias "Alias1" (fun _ ->
-///     "T1" ==> "T2" |> Run
-/// )
+/// HiddenTarget "T1" DoNothing
 let HiddenTarget name body =
     Target name body
     Hidden.Add name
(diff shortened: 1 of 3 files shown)


pms1969:
No idea how to write unit tests for these changes.  looked all thru the test project, but couldn't find any targethelper tests to add too.

pms1969:
I don't understand the AppVeyor error. :-s  I've obviously done something a bit squiffy...  Any ideas?

forki:
No appveyor was broken before. Will look into this tomorrow

pms1969:
Sorry, I know you're busy.  Any progress?

forki:
I'm not yet convinced that this is a good idea. Still thinking about it. Especially the ==> inside of targets is scary.

pms1969:
That's fair enough.  i know it works very well, as that's how I use it, but maybe some other mechanism to do the same thing?  Say Alias equates to *=> which then trails on the rest of the chain, and this isn't evaluated until it tries to run the alias?

pms1969:
I've found a better way of expressing what objective I'm trying to accomplish with this.

In Rake, you have tasks, and they can have dependencies.  like this 
`
task 'buildthing' => "prebuildthing" do
  puts "building"
end
`
so every time that "buildthing" gets called, it makes sure that prebuildthing is run before it.  Fake has the equivalent using the `==>` and `=?>` syntaxes.  I prefer the Fake implementation, because it allows for those dependency chains to be gathered in one place, and it makes tracking them down where a dependency chain is breaking down much easier

Rake also has dependency lists.  Like so:`task :go => [:first, :second, :buildthing, :postbuildthing]`

This creates a target that describes a chain of other targets that should be called when it is called, and will also inject any of their dependencies.  so in our example here, that means that when `:go` is invoked, it will call `:first => :second => :prebuildthing => :buildthing => :postbuildthing`.  As it stands this is what I'm trying to achieve by describing my task dependencies within another task.

matthid:
I guess the same can be achieved via https://github.com/fsharp/FAKE/pull/920 (http://fsharp.github.io/FAKE/soft-dependencies.html). Let me know if I miss something and we can re-open this.

PR #1055, opened 2015-12-22
Fix encoding of AssemblyInfo as written by ReplaceAssemblyInfoVersions

pdfforge:
The function ReplaceAssemblyInfoVersions was reading files with the
encoding set by build param or defaulting to the system default, as
ReadFile and WriteFile do under the hood.

The AssemblyInfo files are always encoded in UTF-8. Therefore, the file
encoding that is actually used should not depend on a build param.


forki:
thx

PR #1057, opened 2015-12-26
Add support for appcast generation

coolya:
Appcasts are a RSS extension[1] that can be used to publish updates to
client. It is used by the popular sparkle[2] framework to publish updates
to OS X clients.

Currently not all features of the appcast format are supported and I have not added tests. I would like to get feedback if such a feature would get into fake.

[1] http://microformats.org/wiki/appcast-formats
[2] http://sparkle-project.org

### DIFF
--- a/src/app/FakeLib/Appcast.fs
+++ b/src/app/FakeLib/Appcast.fs
@@ -37,6 +37,10 @@ type AppcastItem = {
     shortVersion : string option;
     /// Mime type of the update file, usualy octetstream
     mimetype : MimeType
+    /// Optional DSA signature for the archive. It is recommended to use this if the app itself is not signed
+    dsaSignature : string option;
+    /// Optional miminal system version for the update
+    minimumSystemVersion : string option;
 }
 
 /// Configuration data for the appcast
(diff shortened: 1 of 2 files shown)


forki:
yes this sounds useful

forki:
ready to merge?

coolya:
Will push my latest changes this evening. Then it’s ready.

coolya:
I'm done.

src/app/FakeLib/Appcast.fs
```diff
@@ -0,0 +1,110 @@
...
+    title : string;
+    /// Short description
+    description : string;
+    /// Language of your app
+    language : string;
+    /// Updates published to client, can habe multiple updates e.g. for different OS versions
+    items : AppcastItem list;
+}
+
+/// writes an appcast to a file
```
forki: can we have a little more docs. like a sample or something.

coolya: Like this or is it to complex?

forki:
thx

PR #1061, opened 2016-01-01
Updated as per latest git version support.

kunjee17:
As per latest git 2.6.x there is no COMMITMESSAGE file but it is replaced by COMMIT_EDITMSG file. I guess that is same file you are looking for. It is having last commit message. I have keep the original code commented (only file name changed.). If it is required.

### DIFF
--- a/src/app/FakeLib/Git/CommitMessage.fs
+++ b/src/app/FakeLib/Git/CommitMessage.fs
@@ -9,7 +9,8 @@ open System.IO
 
 /// Returns the commit message file.
 let getCommitMessageFileInfo repositoryDir = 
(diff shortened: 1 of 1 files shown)


forki:
I think we need to look for both (as a fallback)

kunjee17:
@forki Oh yes. I should have thought that. But I don't know how can I check git version and put condition there. :(

forki:
just test which file exists

kunjee17:
@forki yup. I ll do it in weekend. :)

forki:
ping

kunjee17:
@forki few hours... Sorry for missing this weekend. I m doing it.

kunjee17:
You can discard this merge request. I have given new one. I don't know how to merge one. Sorry. :(

PR #1064, opened 2016-01-05
NUnit 3 support

fstaffa:
* support for NUnit3, based on NUnit 2 support
* no compatibility layer for NUnit2, users will have to create the arguments from scratch
* tested all supported parameters and they work fine
* missing - automated tests, is there some way how such extension would be tested in Fake ?
* missing - do I need to test somehow that support for TeamCity [working directory](https://github.com/fsharp/FAKE/pull/1064/files#diff-c6cffee77df9960127aff31f4784d000R165) works as expected? I took the logic from NUnit client and I am not sure what is the expected behavior

### DIFF
--- a/src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
+++ b/src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
@@ -1,0 +1,280 @@
+﻿[<AutoOpen>]
+module Fake.Testing.NUnit3
+
+open System
+open System.Text
+open System.IO
+open Fake
+
+/// Process model for NUnit 3 to use.
+type NUnit3ProcessModel = 
+    | DefaultProcessModel
+    | SingleProcessModel
+    | SeparateProcessModel
+    | MultipleProcessModel with 
+    member x.ParamString =
+        match x with
+        | DefaultProcessModel -> ""
+        | SingleProcessModel -> "Single"
+        | SeparateProcessModel -> "Separate" 
+        | MultipleProcessModel -> "Multiple"
+/// The --domain option controls of the creation of AppDomains for running tests. See [NUnit-Console Command Line Options](http://www.nunit.org/index.php?p=consoleCommandLine&r=2.6.4)
+type NUnit3DomainModel = 
+    /// The default is to use multiple domains if multiple assemblies are listed on the command line. Otherwise a single domain is used.
+    | DefaultDomainModel
+    /// No domain is created - the tests are run in the primary domain. This normally requires copying the NUnit assemblies into the same directory as your tests.
+    | NoDomainModel
+    /// A test domain is created - this is how NUnit worked prior to version 2.4
+    | SingleDomainModel
+    /// A separate test domain is created for each assembly
+    | MultipleDomainModel with
+    member x.ParamString =
+        match x with
+        | DefaultDomainModel -> ""
+        | NoDomainModel -> "None"
+        | SingleDomainModel -> "Single"
+        | MultipleDomainModel -> "Multiple"
+
+/// The --framework option in running NUnit 3. There are three kinds - VXY, which means either .NET framework or Mono, NetXY (use .NET framework with given version)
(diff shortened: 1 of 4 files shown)


src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
```diff
@@ -0,0 +1,234 @@
...
+        | MultipleDomainModel -> "Multiple"
+
+type NUnit3ErrorLevel = TestRunnerErrorLevel
+
+type NUnit3Params =
+    { /// The path to the NUnit3 console runner: `nunit3-console.exe`
+      ToolPath : string
+
+      /// NUnit console runner name. ( `nunit3-console.exe`)
+      ToolName : string
```
inosik: Do we really need `ToolPath` & `ToolName`? Most (or all?) other helpers just have a `ToolPath` parameter.

src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
```diff
@@ -0,0 +1,234 @@
...
+      /// If not specified, all agent processes run tests at the same time, whatever the number of assemblies.
+      /// This setting is used to control running your assemblies in parallel.
+      Agents : int option
+
+
+      /// Controls how NUnit loads tests in processes. See: [NUnit3ProcessModel](fake-testing-nunit3-nunit3domainmodel.html).
+      Domain : NUnit3DomainModel
+      
+      /// Allows you to specify the version of the runtime to be used in executing tests.
+      Framework : string
```
inosik: Would be nice if this would be a union type rather than `string`.

src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
fstaffa: Makes sense, will merge these two.

src/app/FakeLib/UnitTest/NUnit/NUnit3.fs
fstaffa: I thought about it, but I am not sure about which values should be provided. NUnit documentation does not list these values as nicely as for NUnit3DomainModel or NUnit3ProcessModel. Instead, it accepts any value and checks at runtime that it is [available](https://github.com/nunit/nunit/blob/master/src/NUnitEngine/nunit.engine/RuntimeFramework.cs#L252). The best that can be done is listing some of them (e.g. mono, net-4.5, v4.0, 2.0, mono-4.0) and adding option for Other(string). Do you think this would be better?

forki:
Is this ready to merge?

fstaffa:
It is. I have tested it with all in this wrapper and it works fine.
@inosik proposed changing the argument of Framework from string to discriminated union but it probably won't bring much benefit.

forki:
ok then please change it now, so that we don't have breaking change later

fstaffa:
Ok, will add there common runtimes which are supported by NUnit 3 and we can expand them later as needed

fstaffa:
I have changed runtime to discriminated union. I have predefined .net frameworks (2.0, 3.5, 4.0, 4.5) and added Other for any other value.
I was not able to get it running (or get NUnit3 running at all) with mono runtime so I didn't predefine any mono runtime and it will have to be defined manually using Other.
Otherwise this is ready for merge.

inosik:
Maybe this might help you: https://github.com/nunit/nunit/blob/master/src/NUnitConsole/nunit3-console/ConsoleOptions.cs#L106
And this as well: https://github.com/nunit/nunit/blob/master/src/NUnitEngine/nunit.engine.tests/RuntimeFrameworkTests.cs#L219-L244

fstaffa:
Thank you, this looks good. I added them (except frameworks 1.0 and 1.1 which should not be supported by NUnit 3)

johnkors:
+1 on this feature. :)

forki:
thanks a lot

PR #1066, opened 2016-01-11
Add support for the Paket pack parameter buildPlatform

trondd:
I notice that the atom plugin removed the training whitespace. If this is a problem I'll submit a new patch without the whitespace changes.


forki:
thx

PR #1067, opened 2016-01-12
Added possibility to bulk update assembly infos with file includes

DigitalFlow:
Hey all,

I added a version of the function for bulk assembly information replacement, which features FileIncludes.
This way we can remove single assembly infos from replacement, while still being able to process them in bulk.

Kind Regards


forki:
thx

PR #1068, opened 2016-01-12
Added wix page to navigation

DigitalFlow:
Hey all,

added the minimal working example for WiX setup generation target to the navigation bar.
Hope this is enough for making it work, the GenerateDoc target does not work for me, so I can't verify right now.

Kind Regards


forki:
there are two template files - please check that both contain the link

DigitalFlow:
Found it.

PR #1073, opened 2016-01-13
AssemblyInfoFile: add functions to read and update attributes

zzdavid:
Here is a try at adding functions to read and update attributes in assembly info files, which was something I missed when starting to use FAKE
- UpdateAttributes
- GetAttributeValue
- GetAttribute
- GetAttributes

Regexes are used to match attributes. It's not perfect, but probably enough for most cases. The functions should work for both C#, F#, VB, and C++, but I have actually only tested them with C#.


forki:
thanks

Issue #992, opened 2015-11-01
Build error locally


mastoj:
Just cloned the master repo on Windows and ran `build.cmd`. The build fails with the following exception:

```
Failures:                                                                     
                                                                              
compile csfiles to dll                                                        
» should compile to dll (FAIL)                                                
Fake.MSBuildHelper+BuildException: Exception of type 'Fake.MSBuildHelper+Build
Exception' was thrown.                                                        
   at Fake.CscHelper.Csc(FSharpFunc`2 setParams, FSharpList`1 inputFiles) in C
:\Dev\git\FAKE\src\app\FakeLib\CscHelper.fs:line 125                          
   at Test.FAKECore.compile_csfiles_to_dll.<>c.<.ctor>b__8_2() in C:\Dev\git\F
AKE\src\test\Test.FAKECore\CscSpecs.cs:line 60                                
                                                                              
Contexts: 250, Specifications: 670, Time: 23.77 seconds                       
  666 passed, 1 failed, 3 ignored                                             
Running build failed.                                                         
Error:                                                                        
System.Exception: MSpec test failed on C:\Dev\git\FAKE\test\Test.Fake.Deploy.dll, C:\Dev\git\FAKE\test\Test.Fake.Deploy.Web.dll, C:\Dev\git\FAKE\test\Test.Fake.Deploy.Web.File.dll, C:\Dev\git\FAKE\test\Test.FAKECore.dll, C:\Dev\git\FAKE\test\Test.Git.dll.                                                         
   at Microsoft.FSharp.Core.Operators.FailWith[T](String message)             
   at Fake.MSpecHelper.MSpec(FSharpFunc`2 setParams, IEnumerable`1 assemblies)
                                                                              
   at FSI_0001.Build.clo@138-8.Invoke(Unit _arg6) in C:\Dev\git\FAKE\build.fsx:line 139                                                                     
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target)              
```

It says on github that it passes, so it is a little bit weird that it doesn't when I run it locally.

bentayloruk:
I've just had this fail for me too.  In my case, I believe it is caused by there being a space in my username.  The test in question uses `Path.GetTempPath()` and then combines this path with file names and passes to `csc.exe`.  The temp paths are not quoted in the generated command, so `csc` is not invoked correctly.  I will dig into it and PR if I have the time, but thought I'd mention it here too.

mastoj:
[...]

PR #1081, opened 2016-01-17
Add shared sln R# settings with ShadowCopy off.

bentayloruk:
FAKE tests that depend on files can't be run when R# is set to shadow copy assemblies being tested.  This file switches this off for all R# users that have the shared sln settings enabled.

Requested by @forki in #992.

NOTE:  This means you can't build the sln, while your tests are running.


forki:
Thx

PR #1086, opened 2016-01-19
Support target and args in build.cmd on Windows.

bentayloruk:
For example, enable `build.cmd CreateNuget -st` by passing all args from build.cmd to
FAKE, if build.cmd has at least one arg.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -13,8 +13,10 @@ if errorlevel 1 (
   exit /b %errorlevel%
 )
 
-SET TARGET=Default
+SET FAKE_PATH=packages\build\FAKE\tools\Fake.exe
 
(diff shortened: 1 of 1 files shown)


forki:
can you do the same change for the build.sh?

bentayloruk:
I was afraid you were going to say that! :grimacing:

bentayloruk:
It turns out the `build.sh` already supports it.  The existing script is as follows:

`packages/build/FAKE/tools/FAKE.exe $@ --fsiargs -d:MONO build.fsx`

In this case `$@` works with something like `sh build.sh Test -st`.

forki:
Awesome

PR #1087, opened 2016-01-19
Hockey app

rflechner:
Hi,

HockeyApp module is useful for uploading an apk from the build script, but there is a timeout restriction as 2 minutes on upload time.
This PR introduce a parameter making it configurable, because it can be longer than 2 minutes.

Regards


forki:
thx

PR #1089, opened 2016-01-19
Fix case sensitivity on package name when search references in Paket.lock

klettier:
I get a "KeyNotFoundException" when I try to match my references from "paket.references" with those in "paket.lock" because sometimes cases are not the same on package name.


klettier:
You are right ! 
I removed the Trim and updated the Regex to success only on package line (starting with four spaces)

forki:
thx

PR #1103, opened 2016-01-22
Remove Paket managed FCS reference

tpetricek:
The NuGet package still came with `FSharp.Compiler.Service.dll` (as well as the renamed one) and `FakeLib.dll` had a reference to it. This should remove that.

After `build CreateNuGet` and using the binaries from the package, I was able to run FsLab journal with an incompatible version of FCS :)

### DIFF
--- a/src/app/FakeLib/FakeLib.fsproj
+++ b/src/app/FakeLib/FakeLib.fsproj
@@ -221,26 +221,6 @@
   </Choose>
   <!-- Ordinary dependencies managed by Paket as usual -->
   <Choose>
-    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.0')">
-      <ItemGroup>
-        <Reference Include="FSharp.Compiler.Service">
-          <HintPath>..\..\..\packages\FSharp.Compiler.Service\lib\net40\FSharp.Compiler.Service.dll</HintPath>
-          <Private>True</Private>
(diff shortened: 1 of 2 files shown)


forki:
thx

forki:
not sure what the reason is, but now I get:

image

/cc @matthid

forki:
Looks like explicit #I of the sourcelink path made it work

forki:
alpha is released, Dogfooding works, our company internal build works.

if everyone agrees then I release that stuff.

/cc @bjartwolf

forki:
Release notes: http://fsharp.github.io/FAKE/RELEASE_NOTES.html

matthid:
So the issue is unrelated and fixed with https://github.com/fsharp/FAKE/pull/1104 (see https://github.com/fsharp/FAKE/commit/0d830d65c7d5ab999811e3b78af0242757325f58)

tpetricek:
FsLab journal runner works! And adding:

    printfn "%A" (typeof<Microsoft.FSharp.Compiler.SourceCodeServices.FSharpChecker>.Assembly.FullName)

prints the correct version of FCS :)

matthid:
Lets hope we never need the same for `FSharp.Formatting`, for example when snippets try to use `FSharp.Compiler.Service`.... O wait I'm already doing it: https://matthid.github.io/Yaaf.FSharp.Scripting/IntroExamples.html ;)

forki:
release is in progress! thank you guys!

PR #1108, opened 2016-01-23
resolves #1095 - adds paket restore to the PaketHelper

### LINKED ISSUE #1095: Restoring paket.dependencies using FAKE
Might be just me who can't find how to do it but I'm after the ability to do a "paket restore" from FAKE. There's lots of helpers on how to do so in Nuget but in the Paket module there only seems to be the ability to create and push packages.

There's an assumption that I'm getting FAKE via Paket and that paket.dependencies file is the same as required for the solution I'm building. In my case FAKE is already retrieved from elsewhere and I want to restore the paket.dependencies for a solution before building it.

Any help would be appreciated - thanks.

mvkra:
This adds the Restore functionality to the PaketHelper. Let me know if there is any issues.

### DIFF
--- a/src/app/FakeLib/PaketHelper.fs
+++ b/src/app/FakeLib/PaketHelper.fs
@@ -58,6 +58,25 @@ let PaketPushDefaults() : PaketPushParams =
       DegreeOfParallelism = 8
       ApiKey = null }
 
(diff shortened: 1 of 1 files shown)


forki:
Thx

Issue #1109, opened 2016-01-23
TeamCity should build with nodereuse-false as default?


bjartwolf:
As commented on http://theimowski.com/blog/2016/01-18-continuous-delivery-with-fake-paket-and-teamcity/index.html I think running with 
```
MSBuildDefaults <- { MSBuildDefaults with NodeReuse = false}
```
is the way to go on TeamCity, as for example the stylecop.dll's will be locked for, I think, 60 seconds if nodereuse is on. This causing clean to fail if the next build comes within 60 seconds of the previous completing. This caused issues for us and we have been running with node-reuse off in all our build scripts for months now.

theimowski:
maybe another option could be if you checked the "clean checkout directory" before building on TeamCity?

bjartwolf:
Yes. But I'm not sure how to detect that. Have to look into it and check if this is an environment variable we can read in Fake.

forki:
MSBuildDefaults <- { MSBuildDefaults with NodeReuse = false}

we could set this automatically when we detect TC as host

bjartwolf:
That is what I was thinking, I'll send a PR in the not to distant future.

PR #1110, opened 2016-01-25
Disable NodeReuse on TeamCity, it can lead to consecutive builds failing

bjartwolf:
If nodereuse is on, TC can not clean files before after the nodes have been shut down (60 seconds
http://blogs.msdn.com/b/msbuild/archive/2007/04/16/node-reuse-in-multiproc-msbuild.aspx)
This means consecutive builds will fail. This might also be true for other build servers.

This was discussed in https://github.com/fsharp/FAKE/issues/1109
I don't think figuring out the clean-strategy from TC is straight forward, so I assume it's easier to just disable nodereuse on TC.

### DIFF
--- a/src/app/FakeLib/MSBuildHelper.fs
+++ b/src/app/FakeLib/MSBuildHelper.fs
@@ -6,6 +6,7 @@ open System
 open System.IO
 open System.Configuration
 open System.Xml.Linq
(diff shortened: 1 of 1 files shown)


forki:
LGTM 

/cc @partTimeDev might solve our issue as well, right?

PR #1116, opened 2016-01-27
Wix directory hierarchy

stefanlux:
Added the ability to create a full directory component hierarchy, where directories can contain components and directories.
Now you are able to use attachServiceControlToComponents and attachServiceInstallToComponents to on a directory component hierarchy.

There is on minor breaking change:
The property Start in WiXServiceControl is now an option to provide the ability not to start a service on install. If someone has set the property manually, the property has to be changed from e.g. Both to Some(Both)

### DIFF
--- a/src/app/FakeLib/WiXHelper.fs
+++ b/src/app/FakeLib/WiXHelper.fs
@@ -309,8 +309,11 @@ let generateComponentRef (setParams : WiXComponentRef -> WiXComponentRef) =
         failwith "No parameter passed for component ref Id!"
     parameters
 
+type WiXDirectoryComponent = 
+    | C of WiXComponent
+    | D of WiXDir      
 /// Component which wraps files into logical components and which allows to 
-type WiXComponent = 
+and WiXComponent = 
     {
         Id : string
         Guid : string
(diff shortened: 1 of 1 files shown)


src/app/FakeLib/WiXHelper.fs
```diff
@@ -96,13 +100,23 @@ type WiXServiceControl =
         Id : string
         Name: string
         Remove : InstallUninstall
-        Start : InstallUninstall
+        Start : InstallUninstall option
```
forki: is this breaking?

stefanlux:
Yes. But it came just to my mind to extend the InstallUninstall type, like "Never" or "None", instead of option, to accomplish the same task.

src/app/FakeLib/WiXHelper.fs
stefanlux: Yes. But it came just to my mind to extend the InstallUninstall type, like "Never" or "None", instead of option, to accomplish the same task.

forki: yes plese try to keep breaking changes to a very low

stefanlux:
Removed the breaking change from the last commit. The InstallUninstall type now has the new option "Never"

forki:
thx!

PR #1117, opened 2016-01-27
Add helpers to send .Net coverage settings to TeamCity

vbfox:
We use dotCover from nuget ([JetBrains.dotCover.CommandLineTools](https://www.nuget.org/packages/JetBrains.dotCover.CommandLineTools/)) so needed `sendTeamCityDotCoverHome`, I implemented the rest of the settings just for completeness.


forki:
thx

PR #1119, opened 2016-01-29
FakeLib: Added reference to FSharp.Compiler.Service

danielmue:
- Added FSharp.Compiler.Service to paket.references. New entries for .net 4.0.x, 4.5.x and 4.6.x

### DIFF
--- a/src/app/FakeLib/FakeLib.fsproj
+++ b/src/app/FakeLib/FakeLib.fsproj
@@ -200,15 +200,6 @@
   <Import Project="$(FSharpTargetsPath)" />
   <!-- Hand-crafted reference to the renamed version of FCS. You must run build.cmd/build.sh first! -->
   <Choose>
-    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.0')">
-      <ItemGroup>
-        <Reference Include="FAKE.FSharp.Compiler.Service">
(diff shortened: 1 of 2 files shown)


forki:
In don't think this is good. We reference our own FAKE.FSharp.Compiler.Service.dll which is a custom FCS

danielmue:
I was not able to build the solution (reference to FSharp.Compiler.Service was not found).

forki:
ok but all references should go to FAKE.xxxx

forki:
I assume devs no have to run the build before they can open VS. /cc @tpetricek

danielmue:
Ah, okay. I did not expect that (I did run ```paket.exe restore``` only). Overlooked the FAKE.*prefix.

build.sh does not work in my environment, I keep getting a 403 from nuget:

```
Something went wrong while downloading NuGet.CommandLine 3.3.0
Message: Could not download NuGet.CommandLine 3.3.0 from http://www.nuget.org/api/v2/package/NuGet.CommandLine/3.3.0.
    Der Remoteserver hat einen Fehler zurückgegeben: (403) Unzulässig.
  ==> Last trial
```
Probably our company proxy. :-/

tpetricek:
Yeah, this is annoying. 

It could be fixed by invoking the assembly copy & rename from the MSBuild script. I tried to do this at first - you can use some `Copy` task to copy the dependency into the right place - but I couldn't figure out how to do more inside the MSBuild script. 

Perhaps the MSBuild could invoke the FAKE build with the one specific target to copy the DLL in a pre-build step, if there is a way to do this?

danielmue:
Or maybe create a special FAKE.FSharp.Compiler.Service nuget package?

PR #1122, opened 2016-01-30
added all compiler args

cloudRoutine:
params are passed as a list of type safe args instead of a record with a list of magic strings

### DIFF
--- a/src/app/FakeLib/FscHelper.fs
+++ b/src/app/FakeLib/FscHelper.fs
@@ -2,294 +2,59 @@
 /// There is also a tutorial about the [F# compiler tasks](../fsc.html) available.
 module Fake.FscHelper
 
+open System
 open Microsoft.FSharp.Compiler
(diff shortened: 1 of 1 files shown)


src/app/FakeLib/FscHelper.fs
```diff
@@ -141,6 +151,372 @@ let fsc (setParams : FscParams -> FscParams) (inputFiles : string list) : int =
...
+/// Compiles one or more F# source files with the specified parameters.
+/// ## Parameters
+///
+///  - `setParams` - Function used to overwrite the default Fsc parameters.
+///  - `inputFiles` - The F# input files.
+///
+/// ## Sample
+///
+///     ["file1.fs"; "file2.fs"]
+///     |> Fsc [Out ""
```
forki: should be Compile

forki:
Ok let's try that ;-)

Thx for making it much less breaking. We will see if we still break someone.

forki:
catch

PR #1126, opened 2016-02-04
Add PicklesHelper for generating living documentation with Pickles

magicmonty:
see http://www.picklesdoc.com


src/app/FakeLib/PicklesHelper.fs
```diff
@@ -0,0 +1,188 @@
+[<AutoOpen>]
```
forki: please remove autoopen. we try to get rid of these

forki:
awesome. thx

PR #1127, opened 2016-02-04
Replace system directory separator with "/" in ArchiveHelper

stefanlux:
To extract a Tar Archive in Linux created with the ArchiveHelper in Windows, it is required to replace the System directory separator with "/".
This change does not affect the archive extraction in Windows.


forki:
thx

PR #1133, opened 2016-02-06
Create separate environ variable helpers

mglodack:
Currently, `setProcessEnvironVar` is exactly the same as `setEnvironVar` as noted [here](https://msdn.microsoft.com/en-us/library/z46c489x(v=vs.110).aspx).

##### What has changed?
- Mark `setEnvironVar` as obsolete and note to use `setProcessEnvironVar`.

##### What has been added?
- Create individual environment variable setters that wrap the `Environment.SetEnvironmentVariable` method.

### DIFF
--- a/src/app/FakeLib/EnvironmentHelper.fs
+++ b/src/app/FakeLib/EnvironmentHelper.fs
@@ -34,6 +34,12 @@ let environVars target =
 /// Sets the environment variable with the given name
 let setEnvironVar name value = Environment.SetEnvironmentVariable(name, value)
 
+/// Sets the environment variable with the given name for the current user.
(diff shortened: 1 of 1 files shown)


forki:
I don't hink we should make that obsolete. It's not harming that much, right?

mglodack:
@forki It's not harming. But for consistency, I thought that `setEnvironVar` would eventually be removed in a later version of `FAKE`. 

Now all of the new exposed methods would include `User`, `Machine`, or `Process`.

mglodack:
What do you think @forki ?

forki:
I like the additions, but I don't really think we should make that other one obsolete

mglodack:
@forki I've removed the `Obsolete`

forki:
thx

PR #1137, opened 2016-02-12
Added SonarQube support

Slesa:
and added Centigrade Gmbh as user.


Slesa:
New pull request created due to wrong history

PR #1140, opened 2016-02-16
Added SonarQube support

Slesa:
and Centigrade as user

### DIFF
--- a/help/sonarcube.md
+++ b/help/sonarcube.md
@@ -1,0 +1,47 @@
+# Analyze your code with SonarQube
+
+From the [web page](http://sonarqube.org):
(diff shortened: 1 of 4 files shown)


forki:
awesome thx

philipooo:
Maybe you could also add the optional /d:sonar.branch parameter.
In my case thats useful because analysis of feature branches will be seperated from the masters analysis if I populate it with e.g. git`s branch name.

Slesa:
Should this possibly be a list of parameters to add with the /d switch or a simple string?

philipooo:
A list of parameters to add with the /d switch would be also very nice and would bring lots of flexibility.

Slesa:
Maybe you could test https://github.com/fsharp/FAKE/pull/1172, please?

Issue #1143, opened 2016-02-17
WatchChanges defaults to include sub directories


mastoj:
When watching files FAKE will look up the folder of each file and watch that folder including the subdirectories. This is a bad default since it can potentially lead to watching to many files and thus failing the build. Happened me in a quite small repo, but for some reason I reach the limit since also the .git folders packages and others were watched.

Two alternative solutions. 

1. Not include subdirectories
2. New overload where one can specify the behavior.

I think number 2 is best since other might be dependent on the feature as is.

dsyme:
fixed already

PR #1144, opened 2016-02-17
New version of WatchChanges that support options

mastoj:
The first option added is the to ignore subdirectories.

This is a first fix for: https://github.com/fsharp/FAKE/issues/1143


forki:
thanks

Issue #1142, opened 2016-02-17
WixHelper FillInWixScript with none empty path not working 


BoeseB:
When you call [FillWixScript](https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/WiXHelper.fs#L1293-1293) of the WixHelper module with a path where it should search for Scripts that is non empty it doesnt work.
In the sources you can see the parameters for combining the the provided path with the file search pattern is swapped. Shouldn't it be `(wiXPath @@ "*.wxs")` instead of  `("*.wxs" @@ wiXPath )` ?

Same for [FillInWiXTemplate ](https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/WiXHelper.fs#L1340-1340)

PR #1145, opened 2016-02-18
Fixed Issue #1142: Arguments of CombinePaths are switched in WixHelper 

BoeseB:
Fix for #1142


forki:
thanks for taking care of this

PR #1146, opened 2016-02-19
NuGet auto version bug fix

rflechner:
Hi,

Something changed on nuget.org servers.
When I proposed to auto increment NuGets versions, nuget.org was handling the "Accept" header of my requests.

Actually, when I want a JSON, the servers of nuget.org reply a XML Content-type.

So there is a fix checking response content-type and parsing XML.

I have published https://www.nuget.org/packages/ScrapySharp/2.4.0 this morning to test my workaround.

Regards,

rflechner
nuget_bug
nuget_bug2


forki:
thx

PR #1149, opened 2016-02-23
Added support for Paket's Pack `minimum-from-lock-file`

romansp:
Added support for Paket's Pack `minimum-from-lock-file` switch.


forki:
thanks!

PR #1159, opened 2016-03-02
Fix IgnoreTestCase

laedit:
use the Ignored param as stated in AppVeyor docs: http://www.appveyor.com/docs/build-worker-api#update-tests

### DIFF
--- a/src/app/FakeLib/AppVeyor.fs
+++ b/src/app/FakeLib/AppVeyor.fs
@@ -135,7 +135,7 @@ let TestFailed testSuiteName testCaseName message details =
         (EncapsulateSpecialChars message) (EncapsulateSpecialChars details)
 
 /// Ignores the test case.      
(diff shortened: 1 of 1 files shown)


forki:
what about the message param?

laedit:
It wasn't used and have no proper usage in the AppVeyor docs so I assumed that it wasn't necessary.

forki:
yes but this now broke the signature of the function. Which is probably OK in this case, but it also broke the build.

laedit:
Sorry, I was too quick.

If you are ok with the suppression of the `message` param I will fix the build, otherwise I will restore it.

forki:
I suggest you add it again. even if it doesn't do anything useful.

laedit:
Done, sorry for the disturbance.

forki:
thx

PR #1165, opened 2016-03-03
use compileFiles in compile as instructed in [<Obsolete "Use FscHelpe…

davidpodhola:
…r.compileFiles instead">]

Nothing should really change for now. Base for a workaround "fix" to https://github.com/fsharp/FAKE/issues/689 that [makes it impossible to build Fable plugins on build machine without Visual Studio 2015](https://github.com/fsprojects/Fable/blob/master/build.fsx#L67-L79) see https://github.com/fsprojects/Fable/issues/35#issuecomment-190324604


forki:
thx

PR #1166, opened 2016-03-03
Workaround for #689 - run fsc.exe

davidpodhola:
#689 issue reported and also spotted on other places like https://github.com/fsprojects/Fable/issues/35#issuecomment-190324604

I took very simplistic approach and simply try to call `fsc.exe` if `UseFsxExe` parameter is supplied like

```
FscHelper.compile [
            FscHelper.UseFscExe
            FscHelper.Lib [ "bin/Debug" ]
        ]
```
This is probably oversimplistic and does not solve the originally reported issue. On the other hand it works. Should be removed in the future when the root cause of #689 is fixed.

No tests supplied, sorry, but [tested agains the repository from the #689](https://github.com/davidpodhola/CompileFsx/commit/a98cadf24e86c37bab92886aa05cb83f2ea478e9).

### DIFF
--- a/src/app/FakeLib/FscHelper.fs
+++ b/src/app/FakeLib/FscHelper.fs
@@ -321,7 +321,7 @@ type FscParam =
     ///  Display the commandline flags and their usage
     | Help
     ///  Compile using fsx.exe
-    | UseFsxExe
+    | useFscExe
 
 (* - ADVANCED - *)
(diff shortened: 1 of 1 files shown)


forki:
How about FscHelper.compileWithFSC?

davidpodhola:
OK, this is much better. I will close this and create a new one. Thanks!

PR #1167, opened 2016-03-04
Fixed issue with IDs that did not start with a letter

DigitalFlow:
This makes sure, that components and directories created by wiXDir function always start with a letter.


forki:
thanks

PR #1172, opened 2016-03-09
New sonar options

Slesa:
Added missing parameters /d and /s to the SonarQube parameters.

### DIFF
--- a/help/sonarcube.md
+++ b/help/sonarcube.md
@@ -20,7 +20,7 @@ compilation has finished. Then the result is collected and send to the SonarQube
         {p with
          Key = "MyProject"
          Name = "Main solution"
-         Version = "1.0.0" })
+         Version = "1.0.0" }
         )
 
     Target "EndSonarQube" (fun _ ->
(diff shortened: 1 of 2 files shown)


philipooo:
This already looks very good!

But am I right saying it's currently not possible to pass the corresponding values for the parameters?
e.g. `Settings = [("sonar.debug",True);("sonar.login","qwe34234iqeri")]`

There is one use-case that is not yet covered by this solution: We (company) are not storing the credentials in our SonarQube.Analytics.xml, instead we are passing them directly to the command line by our buildserver. Since the analyser does not store the credentials in the filesystem the execution of `end` fails because of the lack of the credentials. We have to pass `sonar.login` (and optionally `sonar.password`, if someone is not using tokens) again to the call of `end`.

A solution could be to add an additional value to the tuples in the settings-property like:
`Settings = [("sonar.debug",True,Begin);("sonar.login","qwe34234iqeri",BeginEnd)]`

Cheers

Slesa:
Probably the settings currently have to be `Settings = [("sonar.debug=True");("sonar.login=qwe34234iqeri")]`. Should I add a second field EndSettings? Tuples in settings is a bit uncomfortable, isn't it?

forki:
thanks

philipooo:
Ok, of course its possible to add the value directly to the list, but the sample does not show this.
https://github.com/fsharp/FAKE/pull/1172/files#diff-83519ec9f82c298670e2d0087edf4c98R59

A second field EndSettings would be totaly fine for me.

PR #1174, opened 2016-03-12
Azure Web Jobs now get created during deploy if they do not already exist

isaacabraham:
This was using a non webjob-specific API to perform deploys which meant you had to first create the webjob some other way.

Now it uses a specific webjobs API so the webjob is created on deploy if it doesn't already exist.


forki:
thx

isaacabraham:
No GIF? ;-)

forki:
tumblr_inline_o3a9e7ooze1raprkq_500

isaacabraham:
lol

PR #1175, opened 2016-03-13
Update canopy and IIS docs

jwood803:
Adds a bit more detail to let users know that FAKE.IIS is a separate package to be installed.


forki:
Why would Canopy require fake.iis?

jwood803:
In the example given in the doc on the Canopy page FAKE.IIS is needed or the `createConfigFile` and the `HostWebsite` functions.

forki:
Ah Ok. Maybe you can clarify this even a bit more then. Thanks

jwood803:
Sure, I'll update it

jwood803:
Updated some of the verbiage that should make it a bit clearer on the user of FAKE.IIS

forki:
thx

PR #1180, opened 2016-03-16
added namespace/class/method args for xunit2

Yantrio:


forki:
awesome

forki:
tumblr_inline_o1uk0ghq9w1raprkq_500

Let's see how this works in practice ;-)

PR #1181, opened 2016-03-17
Update Octokit module for GHE connections.

ianbattersby:
I can imagine there is a better or more idiomatic way of doing this, but for now this seems to work for me.

### DIFF
--- a/modules/Octokit/Octokit.fsx
+++ b/modules/Octokit/Octokit.fsx
@@ -93,6 +93,26 @@ let createClientWithToken token =
         return github
     }
 
+let createGHEClient url user password =
+    async {
+        let credentials = Credentials(user, password)
+        let httpClient = new HttpClientWithTimeout(TimeSpan.FromMinutes 20.)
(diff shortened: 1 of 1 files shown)


forki:
I wonder if this can replace the existing client functions?

ianbattersby:
True, could be an default parameter maybe (e.g. `(?url) token`)? Although I don't know (yet) if like C# all default arguments need to be on the end and if so it's a shame to lose url as the first param. I'll try and find some time to play.

forki:
just let me know

ianbattersby:
I'm learning optional arguments are only allowed on type members which seems a bit much. I was wondering if it might also be possible to have some kind of discriminated union type (?) as the last argument with you could match a `string` or `None` inside, but might small spikes haven't been too successful.

forki:
I gueds it's best to merge it like that. But maybe you could convert GHE to GithubEnterprise or something

forki:
thx

PR #1183, opened 2016-03-21
Add contents of `AssemblyInformationalVersionAttribute` to the `AssemblyVersionInformation` class

amazingant:
Since the `AssemblyInformationalVersion` attribute often contains information not available in the `AssemblyVersion` attribute, such as beta or rc information, certain features (like logging) of other software could benefit from having the more descriptive version information.

Decided to work this up because we recently moved to using `FAKE` to build some older internal projects at work, which means our `AssemblyInfo` files are now auto-generated based on the release notes file. Beta 4 for version `1.2.3` used to be manually-entered as `1.2.3.4`; now the version is entered as `1.2.3-beta4`, and the auto-generated `AssemblyInfo` file writes this as `1.2.3`. Since bugs can come and go quickly during betas, having the rest of the version text would greatly improve the usefulness of logging.


forki:
cool

Issue #1059, opened 2015-12-28
DotCover report and sendTeamCityDotNetCoverageImport


kerams:
According to the docs, this function sends dotcover, partcover, ncover and ncover3 reports to TC using `sendToTeamCity "##teamcity[importData type='dotNetCoverage' path='%s']"`. However, when I pass in my `DotCover.dcvr`, the coverage does not show up in TC. If I specify the exact tool using `sendToTeamCity "##teamcity[importData type='dotNetCoverage' tool='dotcover' path='%s']"`, it all works fine.
Do you think this is an issue with TeamCity or FAKE?

bjartwolf:
I believe the dotcover mentioned in the first command is the one you get if you run the .dcvr through DotCoverReport to create an ndepend report file 
'''
Target "CreateDotCoverXmlReport" (fun _ ->
   DotCoverReport (fun p -> { p with 
                                Source = artifactsDir @@ "MSTest.dcvr"
                                Output = artifactsDir @@ "dotCoverReport.xml"
                                ReportType = DotCoverReportType.NDependXml} ) false)
'''

I think we should clarify this in the docs and make a special .dcvr helper to do the dcvr import. I struggled with this as well.

kerams:
Thanks for clearing this up. Is it only the importing of the entire snapshot (as produced by `DotCoverNUnit`) that requires `tool='dotcover'`?

bjartwolf:
That is my understanding, at least. And I think that is easier to do, then the CreateDotCoverXmlReport should be unnescessary.

baronfel:
I just got bit by this today as well.  I'd be up for writing a PR that obsoletes this function, and creates a new function that provides a DU for the type of coverage tool, if people would use that.

PR #1195, opened 2016-03-31
Make a new method for sending a coverage file to TeamCity

baronfel:
Per #1059 this is a new function that adds the missing 'tool' property to the TC message for coverage reports.  Using this instead of the existing coverage I get the expected summaries in TeamCity:

image

The DU types come from the documentation at the help page for [Manual Coverage Reporting] (https://confluence.jetbrains.com/display/TCD9/Manually+Configuring+Reporting+Coverage)


forki:
thx

PR #1203, opened 2016-04-06
Changes to kudu API.

isaacabraham:
The Kudu API now copies directories recursively and allows you to filter out files within that folder structure.

Also now exposes a function to calculate the web job folder structure so that you can manually deploy to the webjob folder yourself.


forki:
thx

Issue #1202, opened 2016-04-06
Fake.MSBuild tries to use mono to run MSBuild.exe


amazingant:
### Description

See the subject; FAKE is trying to use mono to run `MSBuild.exe` on my system. Seems to only happen on my Windows 10 system?

### Repro steps

1. Open powershell, navigate to a project directory

2. Run build script, watch it return an error

### Expected behavior

The build should run

### Actual behavior

The build stops with the following error message (file paths censored):

```
Running build failed.
Error:
System.Exception: Start of process mono failed. The system cannot find the file specified
   at Fake.ProcessHelper.ExecProcessWithLambdas@76-16.Invoke(String message) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 76
   at Fake.ProcessHelper.ExecProcessWithLambdas(FSharpFunc`2 configProcessStartInfoF, TimeSpan timeOut, Boolean silent, FSharpFunc`2 errorF, FSharpFunc`2 messageF) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 76
   at Fake.MSBuildHelper.build(FSharpFunc`2 setParams, String project) in C:\code\fake\src\app\FakeLib\MSBuildHelper.fs:line 344
   at Fake.MSBuildHelper.MSBuildWithProjectProperties@389.Invoke(String project) in C:\code\fake\src\app\FakeLib\MSBuildHelper.fs:line 389
   at Microsoft.FSharp.Primitives.Basics.List.iter[T](FSharpFunc`2 f, FSharpList`1 x)
   at Fake.MSBuildHelper.MSBuildWithProjectProperties(String outputPath, String targets, FSharpFunc`2 properties, IEnumerable`1 projects) in C:\code\fake\src\app\FakeLib\MSBuildHelper.fs:line 389
   at FSI_0005.Global_build_script.clo@204-10.Invoke(String x) in <my build script here>:line 205
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source)
   at FSI_0005.Global_build_script.clo@201-8.Invoke(Unit _arg7) in <my build script here>:line 203
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 489
```

### Known workarounds

Running the same build script on a Windows 7 system appears to work fine.

### Related information 

* Windows 10
[...]

Issue #1019, opened 2015-11-27
isMacOS misbehaves


albertjan:
```
trace (sprintf "Mac: %b" isMacOS)
trace (sprintf "Plat: %O" Environment.OSVersion.Platform)
trace (sprintf "ID: %O" PlatformID.MacOSX)
```

returns:

```
Mac: false
Plat: Unix
ID: MacOSX
```

mono version:

```
Mono JIT compiler version 4.2.0 (Stable 4.2.0.179/a224653 Tue Oct  6 11:27:49 PDT 2015)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen
```

on Yosemite

tried FAKE 4.9.2 and 4.9.3

albertjan:
This is probably a mono issue, but they don't appear to want to fix it. I'm surprised no one else ran into this.

inosik:
Looking at the [source of System.Environment](https://github.com/mono/mono/blob/4504107/mcs/class/corlib/System/Environment.cs#L233) and commit history, it seems it never actually worked. Maybe we should copy the [code](https://github.com/Microsoft/vscode-mono-debug/blob/8a53868/src/common/Utilities.cs#L35) from the Mono debugger for VS Code.

forki:
PR welcome ;-)

2015-12-17 17:54 GMT+01:00 Ilja Nosik <notifications@github.com>:

PR #1044, opened 2015-12-17
Change isMacOS function - closes #1019

inosik:
As Mono always returns `Unix` for `System.Environment.OSVersion.Platform`, we need another way to check if we are on OS X.

Tested on OS X 10.11.2.

PR #1204, opened 2016-04-06
Prevent false-positive in EnvironmentHelper.isMacOS on Windows

amazingant:
If the current-directory is on a mapped drive, e.g. the M:\ drive, calling File.Exists with "/usr/bin/osascript" translates to "M:\usr\bin\osascript" -- this exists in some cases where Windows is running as a virtual machine on an OSX host.

Lots of code-comment to make this clear. Should still cover #1019 while also resolving #1202.

### DIFF
--- a/src/app/FakeLib/EnvironmentHelper.fs
+++ b/src/app/FakeLib/EnvironmentHelper.fs
@@ -120,8 +120,13 @@ let isUnix = Environment.OSVersion.Platform = PlatformID.Unix
 /// Determines if the current system is a MacOs system
 let isMacOS =
     (Environment.OSVersion.Platform = PlatformID.MacOSX) ||
-      // osascript is the AppleScript interpreter on OS X
-      File.Exists "/usr/bin/osascript"
+      // Running on OSX with mono, Environment.OSVersion.Platform returns Unix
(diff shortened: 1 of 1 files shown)


inosik:
I think changing the check to
```fsharp
let isMacOS =
    (Environment.OSVersion.Platform = PlatformID.Unix) && // <-- Change OR to AND and OSX to Unix
      // osascript is the AppleScript interpreter on OS X
      File.Exists "/usr/bin/osascript"
```
should be enough.

The problem is that on Mono, `Environment.OSVersion.Platform` always returns `Unix`. This property never returned `MacOSX`, actually. See #1019 and #1044.

inosik:
```fsharp
isMacOS = isUnix && File.Exists "/usr/bin/osascript"
```
would be even simpler.

amazingant:
@inosik It would be simpler, but it would also behave incorrectly later if anyone ever felt like fixing mono (or .NET Core?) to correctly return `MacOSX`.

inosik:
@amazingant you're right, didn't think about this.

Speaking of .NET Core, this API is (probably?) going to change anyways. I remember an issue about it at dotnet/corefx.

Issue #1187, opened 2016-03-25
When calling FAKE from powershell, it runs successfully then exits with an error


0x53A:
FAKE Version:

```
Installing 'FAKE 4.22.8'.
Successfully installed 'FAKE 4.22.8'.
```

This powershell script 

```
    $ErrorActionPreference = "Stop"
	# restore fake from nuget
	&".\.nuget\nuget.exe" "Install" "FAKE" "-OutputDirectory" "packages" "-ExcludeVersion"
	# and build using fake
	&".\packages\FAKE\tools\Fake.exe" "build.fsx"
```

Runs successfully and then exits with an error:

```
...
Hello World from FAKE
Finished Target: Default

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target                       Duration
------                       --------
First                        00:00:00.0002082
RestoreNugget                00:00:03.1739934
DownloadBinaryDependencies   00:00:00.1209889
Clean                        00:00:00.0040734
BuildApp                     00:00:15.0406521
BuildUnitTests               00:00:04.5540232
ExtractTools                 00:00:13.9669876
RunUnitTests                 00:01:12.0171539
CoverageReport               00:00:01.5417982
BuildDocumentation           00:00:17.6886626
HarvestHeat                  00:00:00.2742397
ZipUpOutput                  00:00:02.3542330
BuildSetup                   00:00:29.4801323
Default                      00:00:00.0004666
Total:                       00:02:40.2605144
Status:                      Ok
---------------------------------------------------------------------
FAKE.exe : 
In Zeile:5 Zeichen:2
+     &".\packages\FAKE\tools\Fake.exe" "build.fsx"
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

![image](https://cloud.githubusercontent.com/assets/4236651/14045490/049c991e-f29b-11e5-9f53-9ba379c52b17.png)

  
This behaviour does not occur with 

'FAKE 4.7.2'.

![image](https://cloud.githubusercontent.com/assets/4236651/14045682/792a2db8-f29c-11e5-898f-eccc60cd85b9.png)

forki:
It doesn't show an error. What's fake's exit code?

0x53A:
[...]

PR #1207, opened 2016-04-07
[Don't merge] demonstrate #1187

0x53A:
This demonstrates https://github.com/fsharp/FAKE/issues/1187:

If ``$ErrorActionPreference = "Stop"`` was set, powershell converts any sub-process writing to STDERR into an exception.

When the output of FAKE is redirected (either file-redirection, or just running it in ISE), it writes an empty string to STDERR after it successfully finishes.
The exit code of FAKE is 0.

My guess is that it flushes some kind of buffer before exiting.

I did a manual binary search with the package versions, and it looks like 4.14.5 is the last GOOD version, 4.14.7 is the first BAD version.

I will try to find the exact commit next week...

### DIFF
--- a/FAKE.sln
+++ b/FAKE.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.24720.0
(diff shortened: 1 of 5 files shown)


forki:
What about this now?

0x53A:
This script runner was just quickly thrown together.
I will try to create a unit test which checks that a successful build script does not write to stderr.

PR #1211, opened 2016-04-10
Call WaitForExit twice to catch all output

kkm000:
`Process.WaitForExit()` *must* be called after `WaitForExit(Int32)` to synchronize asynchronous output events (http://stackoverflow.com/a/16095658/1149924, https://msdn.microsoft.com/en-us/library/ty0d8k56.aspx)

This is likely why our output under AppVeyor gets truncated, X-ref fsprojects/FsLexYacc#53


forki:
thanks for taking care of this

PR #1212, opened 2016-04-10
Use UTF-8 encoding for process output on Mono

juergenhoetzel:
Fixes encoding issus on systems when encoding property is not set.

### DIFF
--- a/src/app/FakeLib/ProcessHelper.fs
+++ b/src/app/FakeLib/ProcessHelper.fs
@@ -7,6 +7,7 @@ open System.ComponentModel
 open System.Diagnostics
 open System.IO
 open System.Threading
+open System.Text
 open System.Collections.Generic
 open System.ServiceProcess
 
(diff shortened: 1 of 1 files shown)


forki:
good

@nosami @Yantrio is that that famous "chinese mono bug"?

nosami:
Quite possibly :) If it is, then we have the same bug in Xamarin Studio somewhere :)

forki:
Release in progress. Would love to hear if you can still reproduce Chinese
translation.
On Apr 10, 2016 19:17, "Jason Imison" <notifications@github.com> wrote:

Quite possibly :) If it is, then we have the same bug in Xamarin Studio
somewhere :)

—
You are receiving this because you modified the open/close state.

Reply to this email directly or view it on GitHub
<https://github.com/fsharp/FAKE/pull/1212#issuecomment-208019809>

nosami:
Haven't seen it once since the chat in gitter the other day.

Yantrio:
I'll give it a spin when I start work and let you know :+1:

forki:
Looks like it broke others: https://github.com/fsharp/FAKE/issues/1213

Should I revert?

forki:
Ok I reverted.

@juergenhoetzel I think the idea is good, but it seems we need to be more careful where to enable this.

/cc @kentcb

juergenhoetzel:
> @nosami @Yantrio is that that famous "chinese mono bug"?

out

Output looks like chinese indeed.

forki:
yes that' what @Yantrio and @nosami reported

Yantrio:
Can I ask if you are getting it on osx only, or if it's on linux too @juergenhoetzel ?

juergenhoetzel:
@Yantrio I use Arch Linux only (Mono 4.4.0.40 currently)

PR #1214, opened 2016-04-11
Fixed failure to handle spaces or other special characters when running mono exes.

jtsinnott:
In EnvironmentHelper.platformInfoAction, psi.FileName needs to be surrounded in quotes so that if the filename of the exe has a space in it, the call to "mono " does not fail.

Simply surrounded psi.FileName with double quotes, which should handle all special characters and whitespace.


forki:
thx

Issue #1213, opened 2016-04-11
Start of process xbuild failed


kentcb:
### Description

Since the 4.24.1 release at the weekend, my build now fails with:

```
System.Exception: Start of process xbuild failed. StandardErrorEncoding is only supported when standard error is redirected
  at Fake.ProcessHelper+ExecProcessWithLambdas@80-16.Invoke (System.String message) <0x8883148 + 0x00027> in <filename unknown>:0 
  at Microsoft.FSharp.Core.PrintfImpl+StringPrintfEnv`1[TResult].Finalize () <0x2dd67d8 + 0x00030> in <filename unknown>:0 
  at Microsoft.FSharp.Core.PrintfImpl+Final2@235[TState,TResidue,TResult,A,B].Invoke (Microsoft.FSharp.Core.FSharpFunc`2 env, A a, B b) <0x2dd60f8 + 0x000ca> in <filename unknown>:0 
  at Microsoft.FSharp.Core.OptimizedClosures+Invoke@3266-1[T2,T3,TResult,T1].Invoke (T2 u, T3 v) <0x2dd5ef0 + 0x0002b> in <filename unknown>:0 
  at Microsoft.FSharp.Core.OptimizedClosures+Invoke@3253[T2,TResult,T1].Invoke (T2 u) <0x2dd5e90 + 0x00024> in <filename unknown>:0 
  at Fake.ProcessHelper+ExecProcessWithLambdas@80-18.Invoke (System.String arg20) <0x8883128 + 0x0001a> in <filename unknown>:0 
  at Microsoft.FSharp.Core.FSharpFunc`2[T,TResult].InvokeFast[V] (Microsoft.FSharp.Core.FSharpFunc`2 func, Microsoft.FSharp.Core.T arg1, Microsoft.FSharp.Core.TResult arg2) <0x2dd5b98 + 0x0007b> in <filename unknown>:0 
  at Fake.ProcessHelper.ExecProcessWithLambdas (Microsoft.FSharp.Core.FSharpFunc`2 configProcessStartInfoF, TimeSpan timeOut, Boolean silent, Microsoft.FSharp.Core.FSharpFunc`2 errorF, Microsoft.FSharp.Core.FSharpFunc`2 messageF) <0x8881658 + 0x008ff> in <filename unknown>:0 
[...]

PR #1215, opened 2016-04-11
Use UTF-8 encoding for process output on Mono

juergenhoetzel:
The original pull request was indeed too intrusive: I wasn't aware that this will lead to an runtime error, if stdout/stderr is a terminal. 

This is a revised version of commit 2d63a45a8ab2479234856434c6ab5549b8880017,
which broke external processes when standard error/output is not redirected: Refs #1213

This commit only sets the encoding when output is redirected.


forki:
ok. let's try that one

PR #1216, opened 2016-04-11
add TeamCity to the list of build servers that fails on error stream output

baronfel:
The default FAKE boostrap writes a few lines at Important severity to StdOut, resulting in a failed build even before we get a chance to set the default console trace listener.

This breaks my build :(
image

So this PR just adds TeamCity to the list of build servers that would prefer that you not write important message to StdErr. Error messages do still write to StdErr, though.


forki:
thx

PR #1217, opened 2016-04-11
Add detail to API doc: Rename works on directories

Kazark:
I tested the code and found that `Rename` works on directories. Also I could not find another way to rename directories so I assume this is idiomatic (particularly since the function name doesn't have the word `File` in it).


forki:
thx

PR #1219, opened 2016-04-12
Added open/close block to teamcity target tracing

Yantrio:
This will give us a collapsable tree view in the teamcity build log.


forki:
thx

PR #1220, opened 2016-04-12
Add function for dual signing

Robin1985:
The AppendSignature function can be used to append a SHA256 signature for dual signing

### DIFF
--- a/src/app/FakeLib/SignToolHelper.fs
+++ b/src/app/FakeLib/SignToolHelper.fs
@@ -65,6 +65,42 @@ let Sign (toolsPath : string) (parameters : SignParams) (filesToSign : seq<strin
 /// Signs all files in filesToSign with the certification file certFile, 
 /// protected with the password in the file passFile. 
 /// The signtool will be search in the toolPath.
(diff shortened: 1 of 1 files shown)


src/app/FakeLib/SignToolHelper.fs
```diff
@@ -65,6 +65,42 @@ let Sign (toolsPath : string) (parameters : SignParams) (filesToSign : seq<strin
 /// Signs all files in filesToSign with the certification file certFile, 
 /// protected with the password in the file passFile. 
 /// The signtool will be search in the toolPath.
```
forki: something look wrong with the xml comments!? that's from the old function, right?

Robin1985: yes, sorry they belong to the obsolete function and I forgot moving them after inserting the new code

forki: please add new XML comments and let me know. looks like a nice addition.

Robin1985: ok, done and created a new pull request so I will close this one, thanks

PR #1222, opened 2016-04-12
Reversed the order of actions in traceStartTarget

Yantrio:
Now traceStartTarget the opposite order of traceEndTarget and teamcity encapsulates the entire target and it's tracing inside the block


forki:
thanks.

PR #1223, opened 2016-04-13
Robin1985 add function for appending signature

Robin1985:
added XML comments and moved them to the proper places


forki:
thanks

PR #1228, opened 2016-04-26
fix parameter split (fixes #1226)

### LINKED ISSUE #1226: Passing arguments with '=' inside does not work as expected
### Description

I'm trying to pass in an argument from my build system to the build script. This argument is a BASE64 string. It looks something like this: c3VwZXJzZWNyZQ==. When the argument reaches my code, it's missing the trailing == (the padding).

### Repro steps

1. Save the following script:
```
#r @"packages/FAKE/tools/FakeLib.dll"
open Fake

let secret = getBuildParamOrDefault "secret" ""

Target "Test" (fun _ ->
    traceFAKE "SECRET %s" secret
)

"Test"
RunTargetOrDefault "Test"
```

2. Call it with ```FAKE.exe repro.fsx Test secret=c3VwZXJzZWNyZQ==```

### Expected behavior

Expected the line ```SECRET c3VwZXJzZWNyZQ==``` to be printed, showing that the value is correct with what has been passed in.

### Actual behavior

```SECRET c3VwZXJzZWNyZQ``` is printed (notice the missing ==).

### Known workarounds

I could manually pad the base64 string to the correct length, but that's messy. I've tried all sorts of escaping, without any luck.
I could also temp store it in an env variable I assume, but since the value is a secret, I'd rather not.

### Related information 

* Operating system - Windows 10 Pro Insider Preview 14316
* Branch - FAKE - F# Make "4.25.4"

Stift:
As in #1226 correctly said FAKE was splitting the build parameter incorrectly when in the value part an '=' existed. I just fixed the split of the param to prevent the vanishing of anything after the '=' in build parameters.
before: 

    param=value=any -> param value
    param=value=== -> param value

now: 

    param=value=any -> param value=any
    param=value=== -> param value===

### DIFF
--- a/src/app/FAKE/CommandlineParams.fs
+++ b/src/app/FAKE/CommandlineParams.fs
@@ -6,11 +6,14 @@ let printAllParams() = printfn "FAKE.exe [buildScript] [Target] Variable1=Value1
 
 let parseArgs cmdArgs = 
     let splitter = [| '=' |]
+    let split (arg:string) = 
+      let pos = arg.IndexOfAny splitter 
+      [| arg.Substring(0, pos); arg.Substring(pos + 1, arg.Length - pos - 1) |]
     cmdArgs
     |> Seq.skip 1
     |> Seq.mapi (fun (i : int) (arg : string) -> 
            if arg.Contains "=" then 
-               let s = arg.Split splitter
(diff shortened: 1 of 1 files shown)


forki:
thx

PR #1230, opened 2016-04-29
MSBuildHelper: Add BuildWebsite(s)Config.

ianbattersby:
Adds the ability to build websites with a configuration other than `Debug`.

Sorry it took so long to pull-request this @forki after our chat many months ago, completely slipped my mind!

### DIFF
--- a/src/app/FakeLib/MSBuildHelper.fs
+++ b/src/app/FakeLib/MSBuildHelper.fs
@@ -429,37 +429,52 @@ let MSBuildReleaseExt outputPath properties targets projects =
     let properties = ("Configuration", "Release") :: properties
     MSBuild outputPath targets properties projects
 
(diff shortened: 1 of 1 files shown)


forki:
Is this breaking?

ianbattersby:
Nope, should be backwards compatible, on purpose 👍

ianbattersby:
PS. I did change a parameter name on an existing method to be consistent, but given F# as I know it (and the signature otherwise being the same) I thought this was safe.

forki:
Good. Thanks. Will merge / release in Monday. If I get my Internet
connection back.

forki:
Yes changing name is fine in fake. Changing types not so much...

forki:
cool. thanks!

PR #1234, opened 2016-05-05
HockeyAppHelper.fs - added parameter 'AppId'

DavidHamburg:
This parameter is required for UWP apps that targeting windows mobile. HockeyApp is not able to differentiate between windows and windows mobile by uploading a package. If you try to upload without the AppId, HockeyApp will create a new app targeting windows.


forki:
cool thanks!

PR #1235, opened 2016-05-10
Node reuse on build servers causes trouble for subsequent builds

trondd:


forki:
So this would enable it on TC right?

trondd:
Bah, of cause the last minute change would fuck it all up. New patch in the air

trondd:
If you want to be conservative I just change it to `(buildServer = TeamCity || buildServer = TeamFoundation)`

forki:
Yes please

forki:
ok NodeReuse = isLocalBuild would exclude TC. That would be ok

Is that what you meant?

trondd:
Yes, but as I said, if you want to be on the safe side I can just limit it to TFS and TC. We're currently only using those two build server, but I would Guess others have the same trouble.

forki:
probably better yes.

trondd:
I created a new pull request with just the single correct commit to avoid cluttering the history, so just drop this one.

PR #1236, opened 2016-05-10
Update Pickles helper to reflect the latest changes to pickles

magicmonty:


forki:
very cool. Thanks a lot

PR #1239, opened 2016-05-13
correct NuGet install verbosity parameter

eltjo-k:
When the verbosity option to the NuGet Install command is set to `NugetInstallVerbosity.Quiet` the following message is shown:

```
C:\...\tools\NuGet\nuget.exe install C:\...\packages.config -outputdirectory "./packages/" quiet -prerelease -nocache -nonInteractive -configFile "./tools/nuget/nuget.config"
install: invalid arguments.
usage: NuGet install packageId|pathToPackagesConfig [options]
```

The problem is that instead of the word `quiet` there should be written `-verbosity quiet`. This is corrected in this pull request. One of the following three options is now added to the NuGet Install command:
```
-verbosity detailed
-verbosity normal
-verbosity quiet
```


forki:
thanks for taking care of this!

PR #1244, opened 2016-05-18
Extended WiXHelper types for supporting creation of 64bit setups

DigitalFlow:
From #1243:

> As I found out, for WiX being able to create a setup that installs to ProgramFiles64, there need to be > certain values set in the package node and in all component and file nodes.
>
> These values are:
>
> Package: Platform="x64"
> Component: Win64="yes"
> File: ProcessorArchitecture="x64"

I added an "architecture" param to the component creation functions and introduced a "Platform" parameter for `FillInWixScript` function.
Default values for all types is "64bit".
The change in the component creation functions however can be regarded as breaking change, since the additional architecture param would have to be added for existing scripts.
Should I change something about that?

Kind Regards,
Florian

### DIFF
--- a/src/app/FakeLib/WiXHelper.fs
+++ b/src/app/FakeLib/WiXHelper.fs
@@ -56,15 +56,28 @@ let getFilesAsWiXString files =
     |> Seq.map (fileInfo >> wixFile)
     |> toLines
 
+type Architecture = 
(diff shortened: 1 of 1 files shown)


forki:
cool stuff. thanks

PR #1247, opened 2016-05-23
NUnit3: don't set a timeout on the nunit3-console process

iainnicol:
A process level timeout is unhelpful, because the timeout is meant to be per test.


forki:
thanks. maybe we need another parameter then.

PR #1248, opened 2016-05-24
Added basic support for bitbuckets piplines CI

TWith2Sugars:
Announcement here: https://blog.bitbucket.org/2016/05/24/introducing-bitbucket-pipelines-beta-continuous-delivery-built-within-bitbucket/

So far it's just the environment variables from https://confluence.atlassian.com/bitbucket/environment-variables-in-bitbucket-pipelines-794502608.html and a new DU case.


forki:
cool. keep it coming

PR #1249, opened 2016-05-24
XamarinHelper: surround jarsigner input file path with quotes

MichalDepta:
Added surrounding the input file path with quotes just like it's being done for the jarsigner tool path and the output path. This allows for whitespaces in the input path.


forki:
thanks!

PR #1250, opened 2016-05-24
Changed the cache path to be relative to script location

### LINKED ISSUE #1114: Feature request: Please create the script cache in the .fake folder below the script
If you call your script from various different places, you end up with .fake folders all over your hard drive.  So I'd like to propose that the .fake cache folder is created under the directory where the script resides, not the directory from where it is called.

ChrSteinert:
Fixes #1114 
Simply changes to cach directory by inserting the directory of the script in front of the cach path.


forki:
I guess that makes sense. thanks

forki:
thanks!

Issue #1177, opened 2016-03-14
Consider adding support for DocFx


Martin-Bohring:
[DocFx](https://github.com/dotnet/docfx) is the new documentation builder from the .Net Foundation.

It follows a similar approach as [F# Formatting](https://github.com/tpetricek/FSharp.Formatting) (e.g. mixing reference documentation with conceptional pages and the output is markdown as well), but with support for c# (unfortunately there is no F# support, yet).

Of course it is possible to just shell out a process to use it during build time, but an it just works experience is still the better option.

forki:
Yes I would probably accept a PR for such a helper.

forki:
implemented in #1251

PR #1251, opened 2016-05-27
Feature/#1177 docfx helper

shortsn:
Hi, this adds a Helper for DocFx (requested in #1177).
It works like a charm in combination with the [DocFx Commandline Package](https://www.nuget.org/packages/docfx.msbuild/) as build dependency.

### DIFF
--- a/src/app/FakeLib/DocFxHelper.fs
+++ b/src/app/FakeLib/DocFxHelper.fs
@@ -1,0 +1,53 @@
+﻿[<AutoOpen>]
+/// Contains helper functions to run the documentation tool "docfx".
+module Fake.DocFxHelper
+
+open System
+
+/// The parameter type for DocFx.
+type DocFxParams = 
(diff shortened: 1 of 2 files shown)


forki:
can you please push a zero diff commit to trigger the build again? thx.

shortsn:
@forki -> done

forki:
awesome. thanks

forki:
I removed the autoopen, because we don't want to create new modules with that

shortsn:
Ah ok, good to know.

bentayloruk:
@shortsn just adding a note for others that end up here.  The DocFx package you linked to has been unlisted.  I *think* the replacement is [docfx.console](https://www.nuget.org/packages/docfx.console/).  Works for me anyway, thanks!

shortsn:
@bentayloruk, you're right. Thanks for this addition.

PR #1252, opened 2016-05-31
Bamboo buildNumber environment variable in case sensitive behavior

philipooo:
When using the bamboo environment variable in capital letters it won't work on our OSX (mono) buildserver. So this helper need to use the bamboo environment variable in its original form to make it work on all platforms.


forki:
thanks for taking care of this one.

PR #1253, opened 2016-05-31
Include docs for all versions of combinePaths

theprash:
A colleague of mine looked at the hover tip for `</>` but found no information. So I've copied the doc string from `@@`.


forki:
:heart:

Issue #1254, opened 2016-06-01
Fake.AssemblyInfoFile Regex for attributes in VB.NET is case-sensitive


forgetaboutit:
### Description

Functions in the Fake.AssemblyInfo module don't work as expected because they don't find all attributes in an AssemblyInfo.vb file with uppercased `<Assembly: ...>` attributes. This is unhelpful since Visual Studio uppercases the `assembly:` part by default. The problem seems to be that the regex (`regexAttrNameValueVb`) for matching the assembly attributes is used in a case-sensitive way.

### Repro steps

0. In Visual Studio, enable Extras -> Options -> Text Editor -> Basic -> Advanced -> "Pretty listing (reformatting) of code"
1. Create an `AssemblyInfo.vb` file in a VB.NET project and edit the versions in the file for example
2. Use the `GetAttributes` function on the file to extract all attributes

### Expected behavior

`GetAttributes` should return a sequence containing the attributes in the `AssemblyInfo.vb` file

### Actual behavior

`GetAttributes` returns an empty sequence

### Known workarounds

Disable the aforementioned option in Visual Studio and manually lowercase the `<Assembly: ...>` to `<assembly: ...>`.

### Related information 

* Observed on FAKE version 4.26.0.
* Problem still exists in the source for version 4.28.0

forgetaboutit:
If someone told me how to fix it, I'd happily provide a pull request. I'm not sure, however, what would be the best solution to fix the issue. Do you think that matching the regexes case-insensitively in general would be acceptable?

PR #1255, opened 2016-06-02
#1254 Support VB.NET's case-insensitive assembly attributes

forgetaboutit:
Use case insensitive regex matching when looking for assembly attributes in `.vb` assembly info files.

Fixes issue #1254.


src/app/FakeLib/AssemblyInfoFile.fs
```diff
@@ -14,6 +14,13 @@ let private regexAttrNameValueFs = @"^\s*\[\<\s*assembly:\s*(?<name>\w+?)\s*\((?
 let private regexAttrNameValueCpp = @"^\s*\[\s*assembly:\s*(?<name>\w+?)\s*\((?<value>.*)\)\s*\]\s*;\s*$"
 let private regexAttrNameValueVb = @"^\s*\<\s*assembly:\s*(?<name>\w+?)\s*\((?<value>.*)\)\s*\>\s*$"
 
+// additional regex options for matching assembly attributes
+let private regexAttrOptionsCs = Option.None
+let private regexAttrOptionsFs = Option.None
+let private regexAttrOptionsCpp = Option.None
+// VB.NET is case-insensitive
+let private regexAttrOptionsVb = Option.Some (RegexOptions.IgnoreCase)
+
```
inosik: I don't think we need these.

src/app/FakeLib/AssemblyInfoFile.fs
```diff
@@ -284,15 +291,22 @@ let private removeAtEnd (textToRemove:string) (text:string) =
...
-    let regex = 
-        if assemblyInfoFile.ToLower().EndsWith(".cs") then regexAttrNameValueCs
-        elif assemblyInfoFile.ToLower().EndsWith(".fs") then regexAttrNameValueFs
-        elif assemblyInfoFile.ToLower().EndsWith(".vb") then regexAttrNameValueVb
-        elif assemblyInfoFile.ToLower().EndsWith(".cpp") then regexAttrNameValueCpp
+    let (regex, additionalOptions) =
+        if assemblyInfoFile.ToLower().EndsWith(".cs") then (regexAttrNameValueCs, regexAttrOptionsCs)
+        elif assemblyInfoFile.ToLower().EndsWith(".fs") then (regexAttrNameValueFs, regexAttrOptionsFs)
+        elif assemblyInfoFile.ToLower().EndsWith(".vb") then (regexAttrNameValueVb, regexAttrOptionsVb)
+        elif assemblyInfoFile.ToLower().EndsWith(".cpp") then (regexAttrNameValueCpp, regexAttrOptionsCpp)
```
inosik: Simply return `RegexOptions.IgnoreCase` for VB and `RegexOptions.None` for everything else.

forgetaboutit: You're right. I will fix this.

forgetaboutit:
I fixed the issues you pointed out and updated the commit accordingly.

inosik:
:+1:

Maybe you should also extend the specs [here](https://github.com/fsharp/FAKE/blob/60acb2f035cb39f2bd40a6c58a4de10d915da248/src/test/Test.FAKECore/AssemblyInfoSpecs.cs#L265)?

forgetaboutit:
Digging into the code a bit more, I discovered the `UpdateAttributes` function, which is also exhibits the same problem. I will fix this as well and extend the specs accordingly. Please wait with the merge until I'm done.

forgetaboutit:
The new commit addresses both issues in `GetAttributes` and `UpdateAttributes` and extends the specs accordingly. I can think of nothing else that would prevent merging the fixes, but I'd be happy to fix any issues.

forki:
Awesome. thanks for taking care of this

PR #1258, opened 2016-06-02
Add BowerHelper module

wooderz:
Have a BowerHelper much like the NpmHelper


ianbattersby:
@wooderz Looks good to me! I suspect however @forki is on holiday sailing so not sure how quickly you'll get it in :(

forki:
awesome. thanks!

PR #1264, opened 2016-06-08
Fixed issue in bulk component creation functions where IDs where invalid

DigitalFlow:
Hey all,

generating directory and file IDs by using the hash code in the first functions for bulk component creations seemed to not be a that good idea, since it can produce invalid IDs.
Streamlined creation of IDs to utilize the "new" approach for generating IDs.

Kind Regards,
Florian


forki:
:heart:

PR #1266, opened 2016-06-13
Add helper function to execute Sysinternals PsExec

Kazark:
This is a small thing, so you may or may not want it, but I've found use for this helper function wrapping up a call to [Sysinternals PsExec](https://technet.microsoft.com/en-us/sysinternals/psexec.aspx).

I throw an error if it fails; maybe that is not consistent with the rest of your APIs? That's what I wanted it to do for my stuff, but feel free to suggest a change on that.

I have not added any automated tests for it, but I have used it myself in my own scripts and it works (w00t w00t manual testing).


forki:
cool thanks!

Issue #742, opened 2015-04-05
Question: Fuchu unit tests


haf:
Hi,

How would I configure FAKE to run Fuchu? It's just an executable; or FAKE could itself load the DLL and run all tests in it.

I suppose I could use ProcessHelper; does it automatically take care to prefix mono, like albacore does, or would I have to add that manually?

forki:
Yes IIRC we detect mono. /cc @mausch

mausch:
See also https://github.com/fsharp/FAKE/issues/234

haf:
Ok, laundry list:

 - Same sort of failed/succeeded outputs as other unit tests
 - Ability to specify a computer-readable output for saving into the CI environment
 - A task that invokes the executable, since that's the 'form' that FAKE has for its tasks.

mausch:
About "same output as other tests / computer-readable output" I was considering ( https://github.com/mausch/Fuchu/issues/1 ) generating NUnit XML output, which is probably understood by every tool out there.

About the FAKE task, do you have any pseudo-code in mind of how it would look like when calling it?

pjean:
Hi,

Can we have an update about Fuchu integration ?
I'm newbie to F# and Fake. If there is a workaround to run Fuchu with Fake, can you provide an example, please ?

Thanks

mausch:
@pjean  Here's an example: https://github.com/mausch/Fuchu/blob/5ef9ec303abf56f8ca5511cda3fb53bfa705007c/build.fsx#L112 i.e. you can just run the tests executable like any other executable.

Kazark:
Is this closed by #1268 ?

mausch:
Cool, had not seen #1268 . Yeah I think this can be closed now...

PR #1268, opened 2016-06-15
Add initial support for Fuchu tests

Kazark:
Something like this has been discussed at #234, and there is an open issue for this more specifically at #742. And while there seem to be some questions of how to do this in the best way, or to polish Fuchu itself, I am of the opinion that it is better to have something than nothing. It is rather reassuring to n00bs particularly when they see that tools are at least aware of each other, and this is one of the big draws of FAKE.

The code is derived from the example from [Fuchu itself](https://github.com/mausch/Fuchu/blob/5ef9ec303abf56f8ca5511cda3fb53bfa705007c/build.fsx#L119). Note that I have not tested this on Mono myself; I was trusting it to work correctly since it came from Fuchu. I have however tested it on Windows. If you want me to test it on Linux, I'm happy to do so.

### DIFF
--- a/src/app/FakeLib/FuchuHelper.fs
+++ b/src/app/FakeLib/FuchuHelper.fs
@@ -1,0 +1,19 @@
+﻿module Fake.FuchuHelper
+
+/// Execute Fuchu tests from one or more assemblies.
+/// Multiple assemblies are run concurrently.
+/// ## Parameters
+/// 
+///  - `testExes` - The paths of the executables containing Fuchu tests to run.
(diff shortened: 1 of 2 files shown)


src/app/FakeLib/FuchuHelper.fs
```diff
@@ -0,0 +1,19 @@
+﻿module Fake.FuchuHelper
```
Kazark: Should I have marked this `[<AutoOpen>]`?

forki:
cool thanks

Issue #1194, opened 2016-03-31
Object Disposed Exception


Nickolas-:
After some time my CI start to raise unhandled Object Disposed exception. IOS works fine, problem in android.

```
Target "android-package" (fun () ->
    AndroidPackage (fun defaults ->
        {defaults with
            ProjectPath = "project path"
            Configuration = "Release"
            OutputPath = "out path"
        })
    |> AndroidSignAndAlign (fun defaults ->
        {defaults with
            KeystorePath = "Tollminder.keystore"
            KeystorePassword = getBuildParamOrDefault "pass" ""
            KeystoreAlias = "tollminder"
            ZipalignPath = "path/build-tools/23.0.2/zipalign"
        })
    |> ignore
)

```

After in build it try to create "Signed" apk.

```
Verifying alignment of ...*Aligned.apk (4)...Finished Target: android-package

Starting Target: android-hockey (==> android-package)
      50 META-INF/MANIFEST.MF (OK - compressed)
   17780 META-INF/TOLLMIND.SF (OK - compressed)
   35700 META-INF/TOLLMIND.RSA (OK - compressed)
   36850 AndroidManifest.xml (OK - compressed)
   38699 res/anim/abc_fade_in.xml (OK - compressed)
   38989 res/anim/abc_fade_out.xml (OK - compressed)
   39296 res/anim/abc_grow_fade_in_from_bottom.xml (OK - compressed)
   39755 res/anim/abc_popup_enter.xml (OK - compressed)
   40091 res/anim/abc_popup_exit.xml (OK - compressed)
   40446 res/anim/abc_shrink_fade_out_from_bottom.xml (OK - compressed)
   40908 res/anim/abc_slide_in_bottom.xml (OK - compressed)
   41209 res/anim/abc_slide_in_top.xml (OK - compressed)
   41514 res/anim/abc_slide_out_bottom.xml (OK - compressed)
   41815 res/anim/abc_slide_out_top.xml (OK - compressed)

Unhandled Exception:
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Stream has been closed'.
  at System.IO.FileStream.BeginRead (System.Byte[] array, Int32 offset, Int32 numBytes, System.AsyncCallback userCallback, System.Object stateObject) <0x1a2aca0 + 0x00256> in <filename unknown>:0
[...]

PR #1275, opened 2016-06-19
Correct waiting for android packaging and fix for mono processes encoding

alexsorokoletov:
#1213 processes started on mono now always use utf-8 for input/output

Regarding #1194 
I've updated process start function in XamarinHelper to correctly wait for the process to complete and then trace all the output. Unfortunately, I was not able to find out why asyncShellExec is not working as expected. My gut feeling is that the problem is somehow related with Mono 4.4.0 release (before I had no problem with build processes.

Regarding #1213 
I had *sometimes* problem with the encoding of some of the tools output (curl, jarsigner, zipalign). 
Started to look into this because of the #1194 issue I had and found out that fix for #1213 didn't cover all the process start functions and `XamarinHelper.fs` was not using the correct one.

Appreciate any feedback and suggestions how to make the PR better.


forki:
thanks!

PR #1276, opened 2016-06-20
#1194 update fix for mono encoding in ProcessHelper.fs

alexsorokoletov:
to work in all scenarios.

This PR comes as fix for the incomplete pull request #1275 made by me previously.
Problem was that I made incorrect if statement inside another if and thus the encoding enforcement did not work correctly in all cases such as when called from GitHelper or HockeyAppHelper modules.

Apologies for that.


forki:
thanks so much for taking care of this!

alexsorokoletov:
@forki it's not yet working. Can you help me understand how do I test the locally built F# Fake without getting it into the nuget?

alexsorokoletov:
I'm actually repeating steps of https://github.com/fsharp/FAKE/commit/3f33e8c47da28aef8855513359f723f55d002f33#diff-f3936af8db25ef04f3cd0ed22cbc36e5
(commit 2d63a45 regarding #1212 and #1213).

Probably the right way to do is to undo this stuff that I broke in and update specifically HockeyAppHelper and XamarinHelper and GitHelper if needed.

forki:
you can run the build?!

alexsorokoletov:
@forki I do.

PR #1277, opened 2016-06-20
Undo all changes in ProcessHelper made for #1194

alexsorokoletov:


forki:
I wonder if this is needed for all calls on mono!?

alexsorokoletov:
@forki I'm undoing all changes I made in the process helper.
Fixes in XamarinHelper are not related to these changes and I only wanted to solve mono encoding problem for everyone. Obviously, problem is much more complicated that I (silly) thought.

I'll take time and see where we are still using incorrect process start procedures (I have one place in mind called HockeyAppHelper) and make sure we fix the problem there instead of making silver bullet.

forki:
ok. let me know when you think I should merge this

alexsorokoletov:
@forki this is a good question you ask. I first thought it is needed but some scenarios do not need this. So instead of going from "everyone needs it" it is better to go from "add to those who needs it"

alexsorokoletov:
@forki Please merge this as soon as practically possible. I broke hundreds builds by this change, let's bring them back online. This request brings everything in the ProcessHelper to version 4.28 which was stable in these terms.

forki:
ok. release in progress

PR #1278, opened 2016-06-20
improve default npm path handling

krgn:
I added a test to check whether `npm` exists in `/usr/bin` as a standard location in addition to the existing default `/usr/local/bin`. Also, I changed the shebang of `build.sh` such that it uses `/usr/bin/env bash` instead of hard-coding the path to bash (which does not work on systems like NixOS).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
 if test "$OS" = "Windows_NT"
 then
   # use .Net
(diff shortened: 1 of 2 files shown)


TheAngryByrd:
Would it make more sense to try to use what's on the environments path first?  Then maybe default to these?

krgn:
Yes, that is really the better way to do it. Gimme a sec

krgn:
It now attempts to extract the path via `which` on Unix systems, and sets the default to `/usr/bin/npm` if that is not possible for some reason.

forki:
cool thanks.

PR #1288, opened 2016-06-29
Fix Fake.Deploy downloadString

sillyotter:
Switch from incorrect `GetRequestStream()` to
`GetResponse().GetResponseStream()` to read results of `GET` requests.  Also
adjusted docs to make it clear that the server URL you should provide to
the command line tools is `http://localhost:8080/fake`, not
`http://localhost:8080/`


forki:
thanks

PR #1287, opened 2016-06-29
Load solution in VS2015.

mkoerner:
Removed duplicated entry in paket.dependencies and changed order of entries in .fsproj (kept the project from loading in VS2015).

PR #1292, opened 2016-07-05
Fixed broken project introduced in #1268

derwasp:
`FakeLib.fsproj` was failing to open in Visual Studio 2015 with this comment:

> The project 'FakeLib.fsproj could not be opened because opening it would cause a folder to be rendered multiple times in the solution explorer.

More details on the issue [here](http://marcinjuraszek.com/2014/03/folders-in-f-projects-how-to-do-it-what-to-avoid.html)


derwasp:
Oops, duplicate of #1287

PR #1297, opened 2016-07-12
changed typo in continuous web job path

lukebuehler:
it used to be "continous" but it needs to be "continuous"

I pulled my hair out why kudu in azure would not run my web job, until I noticed this typo!


forki:
thx

PR #1302, opened 2016-07-13
Update XUnit2 module to fail gracefully

ianbattersby:
As discussed on twitter the current `XUnit2` library module does not fail gracefully as show in the comparison below;

xunitteamcity

This change mirrors the approach taken by the `NUnit3` module (I believe).

**Edit:** It's worth noting this in fact changes the current (and perhaps 'expected') behaviour 🎱


forki:
Cool.

ianbattersby:
I'm now happy this is fine and the other problems we are seeing related to [TeamCity being stuck in the 90's](https://youtrack.jetbrains.com/issue/TW-39760). Good to merge?

Issue #1161, opened 2016-03-02
FscParam.References works incorrectly


fukuang:
I have a target in `build.fsx` looks like:

    Target "BuildTest" (fun _ ->
        ["./test/Test.fs"]
        |> FscHelper.Compile [Out "./build/Test.dll"
                              FscHelper.Target FscHelper.TargetType.Library
                              References [
                                  "./packages/NUnit/lib/net45/nunit.framework.dll"
                                  "./packages/FsUnit/lib/net45/FsUnit.NUnit.dll"
                                  "./build/Core.dll"]
                             ]
    )
    
I executed it with PowerShell in Cmder, and I got 

    FSC with args:[|"--out:./build/Test.dll"; "--target:library";
    "-r:./packages/NUnit/lib/net45/nunit.framework.dll;-r:./packages/FsUnit/lib/net45/FsUnit.NUnit.dll;-r:./build/Core.dll";
    "./test/Test.fs"|]
    startup (1,1)-(1,81) parse error Assembly reference './packages/NUnit/lib/net45/nunit.framework.dll;-r:./packages/FsUnit/lib/net45/FsUnit.NUnit.dll;-r:./build/Core.dll' was not found or is invalid    

The multiple references are treated as one single string.

amgaera:
A quick workaround for this issue is to use multiple `Reference` params:

```fsharp
Target "BuildTest" (fun _ ->
    ["./test/Test.fs"]
    |> FscHelper.Compile [Out "./build/Test.dll"
                          FscHelper.Target FscHelper.TargetType.Library
                          FscHelper.Reference "./packages/NUnit/lib/net45/nunit.framework.dll"
                          FscHelper.Reference "./packages/FsUnit/lib/net45/FsUnit.NUnit.dll"
                          FscHelper.Reference "./build/Core.dll"
                         ]
)
```

baronfel:
@forki can this be closed since the PR is merged?

PR #1304, opened 2016-07-16
Fixed FscParam.References issue #1161

edgarsanchez:
#1161 It used to generate "-r:path1;-r:path2;-rpath3" which is wrong
Now it generates "-r:path1 -r:path2 -r:path3"


forki:
thanks

PR #1305, opened 2016-07-17
Convert relative path to absolute path when creating NAntXmlTraceListener

### LINKED ISSUE #1295: Cannot find Results.xml output file if current directory is changed.
### Description

Changing the current working directory breaks XML logging.

### Repro steps

* Start a build script via CruiseControl or specifying an output file.
* In the script, change the current directory with `System.IO.Directory.SetCurrentDirectory`.
* Perform some action that would log to the output file.

### Expected behavior

The log message is added to the file in the original location, regardless of the current directory.

### Actual behavior

Throws:

    System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\xxx\xxx\output\Results.xml'

### Notes

The stack trace points [here](https://github.com/fsharp/FAKE/blob/7fc5a19cdb541c9e55d31907da61fe24d43ae630/src/app/FakeLib/TraceListener.fs#L100). `xmlOutputFile` comes from a parameter that defaults to a relative path. Turning this into an absolute path early on should fix the problem.

theprash:
This fixes #1295, to allow build scripts to temporarily change the current directory.

### DIFF
--- a/src/app/FakeLib/TraceListener.fs
+++ b/src/app/FakeLib/TraceListener.fs
@@ -97,11 +97,12 @@ let mutable AutoCloseXmlWriter = false
 /// ## Parameters
 ///  - `xmlOutputFile` - Defines the xml output file.
 type NAntXmlTraceListener(xmlOutputFile) = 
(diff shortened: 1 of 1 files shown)


forki:
thanks

PR #1306, opened 2016-07-18
Update DACPAC module.

isaacabraham:
* Actual error handling.
* Move from MSDeploy.exe to SqlPackage.exe.
* Fix Script and Report Generation modes.


forki:
thanks

PR #1308, opened 2016-07-21
expand typescript search paths on windows to cover all new typescript…

### LINKED ISSUE #1284: TypeScriptCompiler doesn't work with TypeScript 1.5 compiler.
### Description

I'm trying to use `TypeScriptCompiler` but it's failing, because it's looking for TypeScript 1.0, which is an old version. I have TypeScript 1.5 installed on my machine, and I'd expect FAKE to be smart enough to pick it up and use to compile my .ts files.

I can see 

```
let private TypeScriptCompilerPath = 
    @"[ProgramFilesX86]\Microsoft SDKs\TypeScript\1.0\;[ProgramFiles]\Microsoft SDKs\TypeScript\1.0\;[ProgramFilesX86]\Microsoft SDKs\TypeScript\0.9\;[ProgramFiles]\Microsoft SDKs\TypeScript\0.9\"
```

in https://github.com/fsharp/FAKE/blob/b7630d6e9bdb7caf7640ac38ca4166d9d7717d59/src/app/FakeLib/TypeScript.fs#L43. I think it should be able to use my TypeScript binaries from *C:\Program Files (x86)\Microsoft SDKs\TypeScript\1.5* (and probably in a way that will make it work with any new version of TypeScript, and not just 1.5).

baronfel:
… compilers, prefering most recent.

On my machine this returns the following list:

val it : string list =
  ["C:\Program Files (x86)\Microsoft SDKs\TypeScript\1.8";
   "C:\Program Files (x86)\Microsoft SDKs\TypeScript\1.5";
   "C:\Program Files (x86)\Microsoft SDKs\TypeScript\1.0"]

choosing tsc.exe from 1.8 at the end.  there may be underlying issues with ordering, but this is more future-proof than the current implementation. 

This fixes #1284

### DIFF
--- a/src/app/FakeLib/TypeScript.fs
+++ b/src/app/FakeLib/TypeScript.fs
@@ -40,8 +40,7 @@ type TypeScriptParams =
       /// Specifies the timeout for the TypeScript compiler.
       TimeOut : TimeSpan }
 
-let private TypeScriptCompilerPath = 
-    @"[ProgramFilesX86]\Microsoft SDKs\TypeScript\1.0\;[ProgramFiles]\Microsoft SDKs\TypeScript\1.0\;[ProgramFilesX86]\Microsoft SDKs\TypeScript\0.9\;[ProgramFiles]\Microsoft SDKs\TypeScript\0.9\"
+let private TypeScriptCompilerPrefix = "Microsoft SDKs" </> "TypeScript"
 
 /// Default parameters for the TypeScript task
 let TypeScriptDefaultParams = 
@@ -56,7 +55,14 @@ let TypeScriptDefaultParams =
       OutputPath = null
       ToolPath = 
             if isUnix then "tsc"
-            else findPath "TypeScriptPath" TypeScriptCompilerPath "tsc.exe"
+            else 
+                let paths = 
+                    [ System.Environment.GetFolderPath System.Environment.SpecialFolder.ProgramFiles; System.Environment.GetFolderPath System.Environment.SpecialFolder.ProgramFilesX86]
+                    |> List.map (fun p -> p </> TypeScriptCompilerPrefix)
+                    |> List.collect (fun p -> try System.IO.DirectoryInfo(p).GetDirectories() |> List.ofArray with | _ -> [])
+                    |> List.map (fun di -> di.FullName)
+                    |> List.sortDescending
+                findPath "TypeScriptPath" (String.Join(";", paths)) "tsc.exe"
       TimeOut = TimeSpan.FromMinutes 5. }
 
 let private buildArguments parameters file = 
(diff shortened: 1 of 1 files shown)


src/app/FakeLib/TypeScript.fs
```diff
@@ -40,8 +40,7 @@ type TypeScriptParams =
       /// Specifies the timeout for the TypeScript compiler.
       TimeOut : TimeSpan }
 
-let private TypeScriptCompilerPath = 
-    @"[ProgramFilesX86]\Microsoft SDKs\TypeScript\1.0\;[ProgramFiles]\Microsoft SDKs\TypeScript\1.0\;[ProgramFilesX86]\Microsoft SDKs\TypeScript\0.9\;[ProgramFiles]\Microsoft SDKs\TypeScript\0.9\"
```
forki: this is gone completely? is it backwards compatible?

src/app/FakeLib/TypeScript.fs
```diff
@@ -56,7 +55,14 @@ let TypeScriptDefaultParams =
...
       ToolPath = 
             if isUnix then "tsc"
-            else findPath "TypeScriptPath" TypeScriptCompilerPath "tsc.exe"
+            else 
+                let paths = 
+                    [ System.Environment.GetFolderPath System.Environment.SpecialFolder.ProgramFiles; System.Environment.GetFolderPath System.Environment.SpecialFolder.ProgramFilesX86]
+                    |> List.map (fun p -> p </> TypeScriptCompilerPrefix)
+                    |> List.collect (fun p -> try System.IO.DirectoryInfo(p).GetDirectories() |> List.ofArray with | _ -> [])
+                    |> List.map (fun di -> di.FullName)
+                    |> List.sortDescending
```
alexfoxgill: Sorting by name will only work for single-digit versions - this will pick 9.0 over 10.0 for example

baronfel: Yeah, I was going to address that this morning, commits incoming :)

forki:
let me know when things are ready. thx

src/app/FakeLib/TypeScript.fs
baronfel: Instead of explicitly hard-coding the versions we support by default (1.0, 0.9 currently), the new code discovers what versions the user has installed and uses the latest they have.  This _could_ invisibly upgrade some users to a more recent version of tsc, yes, but if they are dependent on a particular version of tsc we should encourage them to set their tools path.

baronfel:
Alright, version checks are in and this should be good to go.

baronfel:
@forki this is ready to go.

forki:
thx

PR #1317, opened 2016-07-21
Fixes broken "Minimal Working Example" in WiX docs

lparkermg:
The call to bulkComponentCreation had the architecture argument missing


forki:
thanks.

PR #1319, opened 2016-07-21
Add GitVersionHelper (take 2)

trondd:
Re-implemented with Newtonsoft.Json. Hopefully no trouble this time.

### DIFF
--- a/src/app/FakeLib/GitVersionHelper.fs
+++ b/src/app/FakeLib/GitVersionHelper.fs
@@ -1,8 +1,21 @@
+/// Containts helper function for GitVersion - a tool to help you achieve Semantic Versioning on your project.
+///
+/// To install GitVersion.exe on Windows, start PowerShell as Administrator and run choco install gitversion.portable -s https://chocolatey.org/api/v2"
+/// For Mac and Unix, install the NuGet version.
+
 module Fake.GitVersionHelper
 
 open FSharp.Data
 open Newtonsoft.Json
 open System
+
+type GitversionParams = {
+    ToolPath : string
+}
+
+let GitversionDefaults = {
+    ToolPath = findToolInSubPath "GitVersion.exe" (environVarOrDefault "ChocolateyInstall" currentDirectory)
+}
 
 type GitVersionProperties = {
                                 Major : int;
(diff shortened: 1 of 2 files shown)


src/app/FakeLib/GitVersionHelper.fs
```diff
@@ -0,0 +1,50 @@
...
+                                NuGetVersionV2 : string;
+                                NuGetVersion : string;
+                                CommitsSinceVersionSource : int;
+                                CommitsSinceVersionSourcePadded : string;
+                                CommitDate : string;
+                            }
+
+
+let GitVersion () =
+    let dirs = [ (environVar "ProgramData")  @@ "chocolatey" @@ "bin" ]
```
forki: is that the most common place?

baronfel: Gitversion does suggest using chocolatey to install the exe, but I think a more resilient way to find the path to the chocolatey install directory is using the environment variable `ChocolateyInstall`. Then we can add `bin` to that and get the install dir.

src/app/FakeLib/GitVersionHelper.fs
```diff
@@ -0,0 +1,50 @@
...
+    let file = "GitVersion.exe"
+    let timespan =  TimeSpan.FromMinutes 1.
+    let usage = "To install GitVersion.exe, start PowerShell as Administrator and run choco install gitversion.portable -s https://chocolatey.org/api/v2"
```
baronfel: [This](http://gitversion.readthedocs.io/en/latest/usage/usage/) describes the common usages.  I'd expect that most people would get the exe from there, especially as the exe is the primary way to get support for build/CI servers as described [here](http://gitversion.readthedocs.io/en/latest/usage/usage/)

forki: I think we need to make that work on mono and windows.

baronfel: There's a [gitversion.commandline](https://www.nuget.org/packages/GitVersion.CommandLine) nuget package, so perhaps we should use the existing tools-path-finding that we do for other nuget-delivered exes? and then we've got conventions for executing those on mono already in FAKE.

trondd:
Btw, this now works on Mac as well with the NuGet version of GitVersion

forki:
great!

PR #1320, opened 2016-07-21
Use latest TypeScript compiler version automatically

alexfoxgill:
This small feature addresses #1284 by allowing a version wildcard in `tryFindFile`. When given a path containing the wildcard `[v]`, the algorithm will search the the parent directory for subdirectories matching a regular expression which identifies digits separated by dots (e.g. `1.8`)

The `TypeScriptCompilerPath` has been updated to use the feature: 

```FSharp
let private TypeScriptCompilerPath = 
    @"[ProgramFilesX86]\Microsoft SDKs\TypeScript\[v]\;[ProgramFiles]\Microsoft SDKs\TypeScript\[v]\"
```

I have not written any tests for this method, as it is difficult to mock the filesystem, which the `tryFindFile` method relies on. Perhaps a configurable proxy module could be created to provide `Directory.GetDirectories`, `Directory.Exists` and `File.Exists` and provide override methods in test code... But I think this is overkill.

### DIFF
--- a/src/app/FakeLib/ProcessHelper.fs
+++ b/src/app/FakeLib/ProcessHelper.fs
@@ -8,6 +8,7 @@ open System.Diagnostics
 open System.IO
 open System.Threading
 open System.Text
+open System.Text.RegularExpressions
 open System.Collections.Generic
 open System.ServiceProcess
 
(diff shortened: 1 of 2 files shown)


alexfoxgill:
I've just seen another PR #1308 which seems to do this too...!

Issue #522, opened 2014-08-19
SemVerHelper.PreRelease.TryParse cannot parse valid nuget pre-release identifiers


HCanber:
The current SemVerHelper.PreRelese parser isn't following the current [SemVer spec](http://semver.org/).
For example this is a valid prerelease string `Beta.7.z.92` that the PreRelease has problems with.

This is fine because nuget does not support those either. :)

The [Nuget prerelease spec]() says
>If there's a string there, it's a prerelease version.
>The prerelease versions are given precedence in alphabetical order (well technically lexicographic ASCII sort order).

Since SemVerHelper:PreRelease.TryParse expects a string followed by an int the following strings cannot be parsed, but they should as nuget will accept them:

- `dev20140819130924` (the digits are the date + time in [ISO8601](http://en.wikipedia.org/wiki/ISO_8601)) order as the number is to large for an int. This is something I really need!
- `dev-2014-08-19T130924`

I suggest you remove the int-parsing and just use the regex  [0-9A-Za-z-]+ as [SemVer spec](http://semver.org/) defines. The spec says that prerelease with only digits should be compared numerically, and identifiers with letters or hyphens should be compared in ascii order. However Nuget specifies ascii sort order so just treating it as a string will be fine.

dsyme:
Fixed in #1325

Issue #1154, opened 2016-02-29
SemVerInfo.ToString doesn't match input


amazingant:
Given a release notes with version `2.9.75-beta2`, calling `notes.SemVer.ToString()` returns `2.9.75-beta` instead of `2.9.75-beta2`. Appears to be from [here](https://github.com/fsharp/FAKE/blob/43c4800c5f132d2c2d6832a3a59232a64d83d4c0/src/app/FakeLib/SemVerHelper.fs#L47). Not sure if it makes a difference, but Paket's `SemVerInfo` typedef contains an `Original` field that contains the original text that was parsed; FAKE's `SemVerInfo` could do the same maybe?

dsyme:
fixed

PR #1325, opened 2016-07-21
Semver parse fix to handle prereleases and build parts

### LINKED ISSUE #1324: SemVerHelper.parse doesn't handle prerelease segments that begin with a digit.
### Description

The SemVerHelper.parse method returns a None prerelease when given a SemVer string like "0.0.12-012345678".  This fails according to the spec, quoted here:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

### Repro steps

1. `Fake.SemVerHelper.parse "0.0.12-01234578" |> fun i -> i.Prerelease.IsNone <> true`

### Expected behavior

The prerelease segment is parsed as Some { Origin = "012345678"; Name = ""; Number = Some "012345678" }.

### Actual behavior

The prerelease segment is parsed as None

### Known workarounds

None

baronfel:
Fixes #1324 by rewriting the prerelease and build parsing sections to adhere to spec, while maintaining the Origin, Name, and Number portions of the current logic.

This has implications for existing uses of this code, because the prior code wasn't up to spec.  This code implements proper comparison for prerelease identifiers, and now discards the build portion in order to compare.

Thoughts?  We hit this when a git sha that we used for stamping prerelease information on a nupkg started with a number and the semver info helper broke :(

Also touches #522 around build information.

### DIFF
--- a/src/test/Test.FAKECore/ReleaseNotesSpecs.cs
+++ b/src/test/Test.FAKECore/ReleaseNotesSpecs.cs
@@ -132,7 +132,7 @@ namespace Test.FAKECore
 #### 2.0.0-alpha001 - December 15 2013
 * A";
         static readonly ReleaseNotesHelper.ReleaseNotes expected =
-            ReleaseNotesHelper.ReleaseNotes.New("2.0.0", "2.0.0-rc007", null, new[] { "A" }.ToFSharpList());
+            ReleaseNotesHelper.ReleaseNotes.New("2.0.0", "2.0.0-rc2", null, new[] { "A" }.ToFSharpList());
 
         It should_parse =
            () => ReleaseNotesHelper.parseReleaseNotes(Notes.FromString(input)).ShouldEqual(expected);
--- a/src/test/Test.FAKECore/SemVerHelperSpecs.cs
+++ b/src/test/Test.FAKECore/SemVerHelperSpecs.cs
@@ -66,7 +66,7 @@ namespace Test.FAKECore
         It should_parse_minor = () => semVer.Minor.ShouldEqual(2);
         It should_parse_patch = () => semVer.Patch.ShouldEqual(3);
         It should_parse_prerelease = () => semVer.PreRelease.ShouldEqual(
-            FSharpOption<SemVerHelper.PreRelease>.Some(new SemVerHelper.PreRelease("alpha", "alpha", FSharpOption<int>.None, new[] { SemVerHelper.PrereleaseIdent.NewAlphaNumeric("alpha") }.ToFSharpList())));
+            FSharpOption<SemVerHelper.PreRelease>.Some(new SemVerHelper.PreRelease("alpha", "alpha", FSharpOption<int>.None, new[] { SemVerHelper.Ident.NewAlphaNumeric("alpha") }.ToFSharpList())));
         It should_parse_build = () => semVer.Build.ShouldEqual("beta");
     }
 
--- a/src/test/Test.FAKECore/SemVerHelperSpecs.cs
+++ b/src/test/Test.FAKECore/SemVerHelperSpecs.cs
@@ -60,13 +60,13 @@ namespace Test.FAKECore
     public class when_parsing_semver_strings
     {
         static SemVerHelper.SemVerInfo semVer;
-        Because of = () => semVer = SemVerHelper.parse("1.2.3-alpha.beta");
+        Because of = () => semVer = SemVerHelper.parse("1.2.3-alpha+beta");
 
         It should_parse_major = () => semVer.Major.ShouldEqual(1);
         It should_parse_minor = () => semVer.Minor.ShouldEqual(2);
         It should_parse_patch = () => semVer.Patch.ShouldEqual(3);
         It should_parse_prerelease = () => semVer.PreRelease.ShouldEqual(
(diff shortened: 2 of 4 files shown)


baronfel:
Also addresses #1154

forki:
do we need to do the same in paket?

baronfel:
Yeah, but the problem there is compounded by the fact that there are two, slightly different implementations of SemVer there already split between the bootstrapper and Paket.Core.

Is it worth splitting this out somehow and/or paket file including the source file over there?

src/test/Test.FAKECore/ReleaseNotesSpecs.cs
```diff
@@ -132,7 +132,7 @@ public class when_parsing_many_pre_release_versions
 #### 2.0.0-alpha001 - December 15 2013
 * A";
         static readonly ReleaseNotesHelper.ReleaseNotes expected =
-            ReleaseNotesHelper.ReleaseNotes.New("2.0.0", "2.0.0-rc007", null, new[] { "A" }.ToFSharpList());
+            ReleaseNotesHelper.ReleaseNotes.New("2.0.0", "2.0.0-rc2", null, new[] { "A" }.ToFSharpList());
```
forki: why this change?

src/test/Test.FAKECore/SemVerHelperSpecs.cs
```diff
@@ -96,9 +96,9 @@ public class when_comparing_semvers
             () => SemVerHelper.parse("1.0.0-alpha.1")
                 .ShouldBeLessThan(SemVerHelper.parse("1.0.0-alpha.beta"));
 
-        It should_assume_alpha_is_smaller_tha_beta =
+        It should_assume_that_longer_prereleases_are_greater =
```
forki: I think it's better to keep the old tests as well

forki:
yes I think I tried to split it out into a separate file but didn't suceed

src/test/Test.FAKECore/SemVerHelperSpecs.cs
baronfel: I kept the test data, but per the spec for these inputs 1.0.0-beta is larger, because a shorter prerelease tag is bigger than a longer prerelease tag.

src/test/Test.FAKECore/ReleaseNotesSpecs.cs
baronfel: I made this change because the comparison logic in the spec changed. rc007 is now less than rc2, so the expected output for the given release notes data had to change to take the improved comparison rules into account.

forki: Oh that's breaking in fake. Not sure if we really want that.

baronfel: Current logic is wrong, though.  Part of the reason I chafed working with nuget was because they implemented almost-sorta-but-not-really-semver, so none of the existing libraries worked.  Do we want to do the same?  Besides, if people depend on that sorting, they can either 

1. use a prerelease of the form `rc.{increment}`, where increment is a number and thus compared numerically, or
2. use a prerelease of the form `rc{padding}{number}`, and be consistent, so `rc007` compared to `rc002`,

forki: The problem is usually the existing packages. But maybe this is less of a
problem now. It was a huge problem couple of years ago before we starter
with paket

baronfel:
re: splitting this out - it's in a separate file with no non-framework dependencies, so we could just paket include it in paket.  This could be an issue, though, as paket exposes SemVerInfo publically so I think we'd get name collisions, and Fake's SemVerInfo wouldn't be the same as Paket's.

baronfel:
I think we want to have a 'correct' implementation of SemVer in Fake and Paket for internal use for sure.  Then, especially in the nuget-related areas of Fake and Paket, if we need to have a stripped down version of Semver that's up to nuget spec but not SemVer proper, we can have mapping functions that assert the invariants.

baronfel:
Alternatively, we could just use kolectiv's SemVer.net (FParsec parser + nice tests) in both places? https://github.com/xyncro/semver.net. That would handle the issue with sharing types around. But then we have even more back-compat issues.

baronfel:
It would be very easy to leave the current implementation, mark it as obsolete, and include this new code as a 'SemVer2Helper' or something like that.  But Paket is the real sticking point.  Do you have any guidance here as to what you prefer?

forki:
good question. I still don't know which packages would be affected

baronfel:
hey, good news! Nuget client has started supporting SemVer 2.0 as of 3.5.0-rc1! https://emgarten.com/2016/08/09/semver-2-0-0-with-nuget-3-5-0-rc1/

So maybe this gives us some cover to move forward?

forki:
ok let's try it.

forki:
thanks!

baronfel:
Want me to mirror these changes to Paket?

forki:
yes. that would be awesome

PR #1326, opened 2016-07-22
fixup git reset helper to use checkout when file resets are requested

### LINKED ISSUE #1136: Git functions Reset.hard and Reset.soft don't work with path parameters
Calling `Reset.hard "." "HEAD" "src\Project\File.cs"` results in the error `fatal: Cannot do hard reset with paths.` because it passes a path parameter to the git `reset` command

According to [this StackOverflow question](http://stackoverflow.com/questions/11200839/why-git-cant-do-hard-soft-resets-by-path) a hard reset for a path should actually be done using `git checkout HEAD -- <path>`

baronfel:
This fixes #1136 by making the reset helper conditionally use checkout under the covers.  it will only do this if a file is specified, which is in line with the usages of git reset.  reset only takes a file path in the case where you are not doing a soft, hard, or mixed reset to a ref.

I also added an explicit open to the Fake.Git.CommandHelper module to make it easier to use this module by sending it to FSI.  This also has the nice side effect of more accurately tracing dependencies across Fake modules.

### DIFF
--- a/src/app/FakeLib/Git/Reset.fs
+++ b/src/app/FakeLib/Git/Reset.fs
@@ -3,12 +3,19 @@
 module Fake.Git.Reset
 
 open System
-open Fake
+open Fake.Git.CommandHelper
 
 let internal addArgs commit file =
     sprintf "%s%s"
       (if String.IsNullOrEmpty commit then "" else " \"" + commit + "\"")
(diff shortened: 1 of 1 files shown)


forki:
thanks

PR #1327, opened 2016-07-22
mask octo api key when tracing

### LINKED ISSUE #611: Octo task should obfuscate ApiKey in traces

baronfel:
Fixes #611 by adding in a dummy api key. It would be cool to have data objects implement a stringify-for-tracing static member that we could use in the tracer or something like that to do this automatically :-/


forki:
thanks

PR #1331, opened 2016-07-22
Add registry support in WiXHelper

rcknight:
With this change, `WiXHelper` is now capable of generating installers which modify the registry.


forki:
Awesome! thanks

PR #1334, opened 2016-07-25
Update search pattern

rachelreese:
I'm not sure if this is the place to change it, but the issues that I've been having with FSharp.Formatting (https://github.com/tpetricek/FSharp.Formatting/pull/416) are solved by removing the period here. Might be better for FSharp.Formatting to solve this? Thanks. :-)


forki:
thx

PR #1337, opened 2016-07-28
fixes >> [...] /paket-files/build/fsharp/FAKE/modules/Octokit/Octokit…

janni2:
….fsx(178,42): error FS0041: A unique overload for method 'GetAll' could not be determined based on type information prior to this program point. A type annotation may be needed. Candidates: IReleasesClient.GetAll(owner: string, name: string) : Tasks.Task<Collections.Generic.IReadOnlyList<Release>>, IReleasesClient.GetAll(repositoryId: int, options: ApiOptions) : Tasks.Task<Collections.Generic.IReadOnlyList<Release>> <<

this leads f.e. to a build error in https://github.com/fsprojects/ProjectScaffold

### DIFF
--- a/modules/Octokit/Octokit.fsx
+++ b/modules/Octokit/Octokit.fsx
@@ -28,7 +28,7 @@ type private HttpClientWithTimeout(timeout : TimeSpan) as this =
     let setter = lazy(
         match typeof<HttpClientAdapter>.GetField("_http", BindingFlags.NonPublic ||| BindingFlags.Instance) with
         | null -> ()
-        | f -> 
+        | f ->
             match f.GetValue(this) with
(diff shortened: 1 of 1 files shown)


janni2:
Hi, to quickly check, what the fix does: 
- clone https://github.com/fsprojects/ProjectScaffold
- execute build.sh/cmd initial
- see that it stops with exception in this file
- modify the one line to have a more specific signature (I used Atom + Ionide), then it steps further
- sorry about the 2 spaces, Im not sure how they came in, but at least you could see the idea

janni2:
Well, the appveyor build failure has no connection with this fix. At the one hand there are general warnings on the other hand it stops because of hanging

PR #1339, opened 2016-07-30
Fix Fsc short toggle argument format. Fixes #1218

### LINKED ISSUE #1218: FscHelper translates "Debug true" to "-g+", which is incorrect
### Description

When FscHelper is given a `Debug true` or `Debug false` in the list of arguments, it emits FSC args "-g+" and "-g-" respectively, which is incorrect and causes FSC to fail with error message "unrecognized option -g+".

This apparently happens because of this rather misleading line from the FSC help:

```
   --debug[+|-]                   Emit debug information (Short form: -g)
```

From which one would (and apparently did) conclude that the long form of this switch would be `--debug-` and `--debug+`, while short forms - `-g-` and `-g+` respectively. But this intuition is incorrect: short forms work differently than the long ones. Long form for enabling debug build is `--debug+`, and short form is just `-g`, whereas for disabling debug build one would specify `--debug-` for long form or omit the switch entirely for short form. In other words, the switch `-g-` does not exist.

### Repro steps

```
Target "Build" <| fun _ ->
  [ "src/src.fs" ]
  |> Compile [ Target Library; Debug false; Out "out.dll" ] 
```

### Expected behavior

Compile the source without debug info.

### Actual behavior

Error: "unrecognized option -g-"

### Known workarounds

If possible, omit the `Debug` argument entirely.

Alternatively, use `compileFiles` instead of `Compile` and specify arguments manually:

```
Target "ReleaseBuild" <| fun _ ->
  [ "src/src.fs" ]
  |> compileFiles [| "--out out.dll"; "-g"; "--target:library" ] 
```

ErikSchierboom:
This PR fixes the problem that was noted in #1218, in which short format toggle should not get a "+" or "-" suffix, but where their presence already indicates if the flag is being set.

### DIFF
--- a/src/app/FakeLib/FscHelper.fs
+++ b/src/app/FakeLib/FscHelper.fs
@@ -371,7 +371,7 @@ type FscParam =
         /// format a compiler arg that ends with "+" or "-": "--%s%s"
         let togl s b = sprintf "--%s%s" s (chk b)
         /// format a short form compiler arg that ends with "+" or "-": "-%s%s"
-        let stogl s b = sprintf "-%s%s" s (chk b) // for short forms       
+        let stogl s b = if b then sprintf "-%s" s else "" // for short forms       
         /// format a list of compiler args with string parameters "--%s:\"%s\""   
         let argls s (ls:string list) = sprintf "--%s:\"%s\"" s (String.concat ";" ls)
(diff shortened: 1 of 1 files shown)


forki:
thx

PR #1343, opened 2016-08-02
Issue 1342: NugetHelper.fs: GetPackageVersion

bobahml:
It seems that in order to solve the problem enough to add a point to the search pattern.


forki:
That would not help with castle and castle.core right?

bobahml:
In the console, I see now:
Version 3.3.3 found for package Castle.Core

forki:
And for caste?

bobahml:
//    Castle.Windsor-NLog.3.3.0
//    Castle.Windsor.3.3.0
 
Also it works ->
Version 3.3.0 found for package Castle.Windsor

forki:
Cool.

PR #1345, opened 2016-08-03
Fix detection of GitLab CI with current multi-runner

pdfforge:
As it seems, the latest multi-runner has changed the content of the "CI_SERVER_NAME" variable to "GitLab" instead of "GitLab CI". In the docs (http://docs.gitlab.com/ce/ci/variables/README.html), they state that we should use the variable "GITLAB_CI", as it is explained with "Mark that build is executed in GitLab CI environment". It is set to the string "true", when we are in GitLab.

### DIFF
--- a/src/app/FakeLib/BuildServerHelper.fs
+++ b/src/app/FakeLib/BuildServerHelper.fs
@@ -61,7 +61,7 @@ let travisBuildNumber = environVar "TRAVIS_BUILD_NUMBER"
 
 /// Checks if we are on GitLab CI
 /// [omit]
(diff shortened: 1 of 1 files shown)


forki:
thx!

forki:
I changed it to 

    let isGitlabCI = getEnvironmentVarAsBool "GITLAB_CI" || environVar "CI_SERVER_NAME" = "GitLab CI"

pdfforge:
Thank you for merging it! Of course, the change is fine, although the documentation says that the GITLAB_CI variable is available in all multi-runner versions since the first, I would not be surprised if there were exceptions and we know that CI_SERVER_NAME contained the proper value for old runner versions

PR #1349, opened 2016-08-08
add octopus command for push

baronfel:
Adds a simple command helper for the [push](http://docs.octopusdeploy.com/display/OD/Pushing+packages) command in octopus


forki:
thx

Issue #1341, opened 2016-07-30
FscHelper.compile fails with multiple references


ErikSchierboom:
### Description

The `FscHelper.compile` function requires assembly references. Explicitly specifying the arguments using `FscHelper.compileFiles` fixes the problem.

This is the code I used within my FAKE script to dynamically compile a set of F# files:

``` fsharp
FscHelper.compile
        [FscHelper.Out output;
         FscHelper.References [
            "tools/NUnit/lib/net45/nunit.framework.dll";
            "tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsecCS.dll";
            "tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsec.dll"]
         FscHelper.Target FscHelper.TargetType.Library]
        files
```

When this code is executed, the following error occurs:

```
startup (1,1)-(1,81) parse error Unable to find the file 'tools/NUnit/lib/net45/nunit.framework.dll -r:tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsecCS.dll -r:tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsec.dll' in any of
 /Library/Frameworks/Mono.framework/Versions/4.4.1/lib/mono/4.5
 /Users/erikschierboom/Programming/xfsharp
 /Users/erikschierboom/Programming/xfsharp/tools/FAKE/tools
```

At first I thought it had something to do with relative paths, but then I tried using `FscHelper.compileFiles`:

``` fsharp
FscHelper.compileFiles files  
        ["--target:library";
         sprintf "--out:%s" output;
         "-r:tools/NUnit/lib/net45/nunit.framework.dll";
         "-r:tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsec.dll";
         "-r:tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsecCS.dll" ]
```

Funnily enough, this works! I then noted that the order of the parameters was different, so I changed my original version to re-order the parameters:

``` fsharp
FscHelper.compile
    [FscHelper.Out output;
     FscHelper.Target FscHelper.TargetType.Library
     FscHelper.References [
         "tools/NUnit/lib/net45/nunit.framework.dll";
         "tools/FParsec/lib/portable-net45+netcore45+wpa81+wp8/FParsecCS.dll";
[...]

PR #1350, opened 2016-08-09
Fix multiple references for DLL

robkuz:
The `ToString` genereates invalid reference `-r:` parameters in case of using the `References` value constructor. Fixed!


robkuz:
fix for 
https://github.com/fsharp/FAKE/issues/1341

forki:
cool thanks

PR #1357, opened 2016-08-17
Perf tuning to prevent extra iteration

rexcfnghk:
Prevent extra iteration by composing the two functions when passed into `List.map`.


forki:
Did you measure this? Do you really think it has any effect on small lists? I mean it's not bad but I doubt it has effect

rexcfnghk:
@forki I just thought it might be better to iterate once instead of twice in that `List.map` call. Perhaps it should be called a refactoring instead of an actual perf tuning.

forki:
for larger lists this can actually be an advantage. here it's probably only nanoseconds difference and all the time is consumed by running the task itself

PR #1358, opened 2016-08-19
Update FscHelper doc

alfonsogarciacaro:
When I try the FscHelper it says all methods in the guide are obsolete so this updates the doc to use the new recommended methods. To avoid problems with `Target` I’m not opening the `Fake.FscHelper` module.


forki:
Thx

PR #1361, opened 2016-08-22
Trim assembly info attribute value

### LINKED ISSUE #1344: AssemblyInfo.GetAttribute and AssemblyInfo.GetAttributeValue double quote response
### Description

I'm sort of confused by the current response of `AssemblyInfoFile.GetAttribute`, but maybe I'm missing something.

I currently have to `Trim` the response because the expected behavior appears to be a doubly quoted strings (see below).

Would it be possible to remove the extra quotes or is this the expected behavior?

### Repro steps

Here is link to the spec that validates the current response.
https://github.com/fsharp/FAKE/blob/master/src/test/Test.FAKECore/AssemblyInfoSpecs.cs#L74

### Expected behavior

I would expect the response be something like this:

```
"1.0.0"
```

### Actual behavior

Currently, the response is of this form:

```
"\"1.0.0\""
```

### Known workarounds

My current workaround for this response is to just `Trim` the extra quotes.
```
let GetVersion () =
    match (AssemblyInfoFile.GetAttributeValue "AssemblyInformationalVersion" "AssemblyInfo.cs") with
    | Some versionString -> versionString.Trim[| '"' |]
    | None ->
      let errorMessage =
          sprintf "Error: missing attributes `%s` in %s" "AssemblyInformationalVersion" "AssemblyInfo.cs"
      traceError errorMessage
      exit 1
```

mglodack:
Resolves #1344 

Please feel free to use this diff to ignore white space changes.

https://github.com/fsharp/FAKE/pull/1361/files?w=1

### DIFF
--- a/src/app/FakeLib/AssemblyInfoFile.fs
+++ b/src/app/FakeLib/AssemblyInfoFile.fs
@@ -35,7 +35,7 @@ let private regexAttrValueVb name =
     + @"(.*)"                                       // value
     + @"(?=\)\s*\>\s*$)"                            // look-ahead for ")>"
 
-let private NormalizeVersion version = 
+let private NormalizeVersion version =
     let m = assemblyVersionRegex.Match(version)
     if m.Captures.Count > 0 then m.Captures.[0].Value
     else version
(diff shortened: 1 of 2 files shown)


forki:
cool. thanks

Issue #1355, opened 2016-08-12
System.ArgumentNullException thrown for colored console output on Mono


Dzoukr:
Ok, this is not actually FAKE`s fault, but right now I do not know where to put this important information, which could save few hours for other developer. Please, take this as something worth documenting (i can send PR, if you would hint me where to put this information):

When running FAKE on Mono (version 4.4.2) on Ubuntu 12.04/14.04, this error occures:

```
Unhandled Exception:
System.ArgumentNullException: Value cannot be null.
Parameter name: format
  at System.ParameterizedStrings.Evaluate (System.String format, FormatParam[] args) <0x7f0d4902ab20 + 0x001e7> in <filename unknown>:0 
  at System.TermInfoDriver.ChangeColor (System.String format, ConsoleColor color) <0x7f0d490256b0 + 0x0010e> in <filename unknown>:0 
  at System.TermInfoDriver.set_ForegroundColor (ConsoleColor value) <0x7f0d490258b0 + 0x00037> in <filename unknown>:0 
  at System.ConsoleDriver.set_ForegroundColor (ConsoleColor value) <0x7f0d490136e0 + 0x00048> in <filename unknown>:0 
  at System.Console.set_ForegroundColor (ConsoleColor value) <0x7f0d49012830 + 0x0000c> in <filename unknown>:0 
  at Fake.TraceListener+ConsoleTraceListener.writeText (Boolean toStdErr, ConsoleColor color, Boolean newLine, System.String text) <0x407f35e0 + 0x00052> in <filename unknown>:0 
  at Fake.TraceListener+ConsoleTraceListener.Fake-TraceListener-ITraceListener-Write (Fake.TraceData msg) <0x40216dc0 + 0x000bc> in <filename unknown>:0 
  at Fake.TraceListener+postMessage@148.Invoke (ITraceListener obj) <0x40216d90 + 0x00025> in <filename unknown>:0 
  at System.Collections.Generic.List`1[T].ForEach (System.Action`1 action) <0x7f0d48cdf440 + 0x0007e> in <filename unknown>:0 
  at Fake.TraceListener.postMessage (Fake.TraceData x) <0x40216c40 + 0x0010c> in <filename unknown>:0 
  at <StartupCode$FAKE>.$Program.main@ () <0x40184d90 + 0x01171> in <filename unknown>:0 
```

To fix this, you **have to set environment property** `SET TERM=xterm-color` ! After setting this, everything works great.

[...]

PR #1362, opened 2016-08-22
Fix for issue #1355 : System.ArgumentNullException thrown for colored output

Dzoukr:
This is my very first PR in FAKE so feel free to add any helpful comments and hints.


forki:
cool. let's see if it works!

PR #1366, opened 2016-08-23
Hint shown in for ArgumentException #1355 instead of trying to set it directly

Dzoukr:
As mentioned in discussion in #1355, it is not possible to properly set TERM variable for colored output error. We should at least tell user what is really wrong and how to fix it.


forki:
thanks

Dzoukr:
It was honor for me - I just love FAKE! I´ll try to help as much as possible.

PR #1367, opened 2016-08-24
Add Checksum[64][Type] in ChocoHelper

### LINKED ISSUE #1364: Add Checksum infos in Choco.Pack
### Description

Checksum will soon be mandatory in chocolatey and should be handled in ChocoHelper.

### Expected behavior

Add Checksum[64] and ChecksumType[64] parameters to ChocoPack.

### Known workarounds

For now, it is possible to use a template for chocolateyInstall.PS1 which add these parameters.

I will propose a PR soon.

laedit:
Fix #1364 
Also add a checksum helper.


laedit:
Since it is something general imo, I placed the Checksum type directly under `Fake` namespace, but it can be moved if it is necessary.

forki:
thx

PR #1369, opened 2016-08-26
Minor regex cleanup. I noticed this from Paket repository paket-files…

Thorium:
I noticed this from Paket repository paket-files link


forki:
nice

PR #1373, opened 2016-09-08
Add comment on top of auto-generated AssemblyInfo.fs files

eiriktsarpalis:
Thought that auto-generated `AssemblyInfo.fs` files should explicitly state their nature. Did some minor code cleanup in the process.


eiriktsarpalis:
@forki I did not add the same comment in the csharp one since it already contains another

forki:
thx!

PR #1377, opened 2016-09-12
Use nuget instead of referenced dlls. Fix SQL Server version issue.

freein:
We should not count on SQL Server version 10 being installed on the local machine and thereby providing Microsoft.SqlServer.SqlClrProvider of the correct version. Instead we should get all needed SQL Server packages from nuget.

1. Remove the dlls referenced by Fake.SQL.
2. Remove their references.
3. Microsoft.SQLServer.SMO


forki:
cool

forki:
Unfortunately I had to revert this since it's breaking all our builds on SQL Server 2008 R2. We need to find a way that's working for everybody.

/cc @isaacabraham

part-timeDev:
It seems like this change breaks our build env. We use SQL Server 2008 R2 on our build agents and in the deploy stage we try to make a backup of a datebase. Since this release, this step failes with 
"System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.SqlServer.BatchParser, Version=12.0.0.0, Culture=neutral, PublicKeyToken=89845dcd8080cc91' or one of its dependencies. The system cannot find the file specified."

Nothing else was changed on the systems.

freein:
The old libs were of version 10.5 which is for 2008 R2. I included the libs of version 12(2014). This worked using 2012 both before and after my update. However now it does not work using 2016 again.

Can we use multiple versions somehow? Or maybe we can use version 11 which we can get from here: https://www.nuget.org/packages/Unofficial.Microsoft.SQLServer.SMO. Can anyone with 2008 R2 test this?

Has not SQL Server 2008 R2 reached end of life btw? Should we still support it?

forki:
it has not reached end of life in the real world ;-)

Issue #1374, opened 2016-09-08
Error throw when FAKE script exiting


sillyotter:
### Description

Infrequently in a script I have, I receive the following exception after everything has run sucessfully and the fake script is just about to exit.

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an 
object.

at Fake.ProcessHelper.killAllCreatedProcesses$cont@483(Unit unitVar) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 485
at Fake.TargetHelper.run$cont@511(String targetName, Unit unitVar) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 560
at Fake.TargetHelper.run(String targetName) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 511
at Fake.AdditionalSyntax.RunTargetOrDefault(String defaultTarget) in C:\code\fake\src\app\FakeLib\AdditionalSyntax.fs:line 29
at <StartupCode$FSI_0005>.$FSI_0005_Build$fsx.main@()                                                                              
--- End of inner exception stack trace ---                                                                                                                                            
at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)                                                                     
at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)                                                                      
at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)                                           
at Fake.FSIHelper.runScriptCached@313.Invoke(Unit unitVar0) in C:\code\fake\src\app\FakeLib\FSIHelper.fs:line 321                                                                     
 ```
This does not happen every time. 

### Repro steps

I have not ever seen this exception until I added a bunch of async code to my script.  Specifically, I ExecProcess quite a few TopShelf based services in parallel with code like this below to  stop/start/install/uninstall services.

```F#
let topshelfServiceAction path args =
    try
        ExecProcess (fun info -> 
                        info.FileName <- path
                        info.WorkingDirectory <- Path.GetDirectoryName(path)
                        info.Arguments <- args
                        info.Verb <- "runas"                        
                        info.UseShellExecute <- true
                    ) 
                    (TimeSpan.FromMinutes 2.0) |> ignore
[...]

PR #1378, opened 2016-09-17
Replace process cache with safe alternative

sillyotter:
ProcessHelper contains a HashSet that is used to capture process id's of all executed processes for later termination.  Starting processes up in a parallel fashion can corrupt this HashSet.  Replacing it with a ConcurrentBag fixes the issue.

This fixes fsharp/FAKE#1374


forki:
thx!

PR #1379, opened 2016-09-18
Remmidemmi: call 'traceEndTask' in the finally block of a try-finally, so that it is always closed, even if the task throws an exception.

0x53A:
should fix all / most issues similar to this: https://github.com/fsharp/FAKE/issues/1079

Script

[FakeTestTests.zip](https://github.com/fsharp/FAKE/files/478621/FakeTestTests.zip)

```
Target "Test" (fun _ ->
  System.Diagnostics.Debugger.Launch() |> ignore
  try
    [ @"C:\Users\lr\Documents\visual studio 2015\Projects\FakeTestTests\test\FakeTestTests.exe" ]
    |> Fake.Testing.XUnit2.xUnit2 (fun p ->
        {p with
                XmlOutputPath = Some <| testDir + "TestResults.xml" })
  with
  | _ -> printfn "Got an Exception!"
)
```

Before
image

after
image

I have only tested the XUnit tasks, but since the same pattern was appliead in other places, I also wrapped these in try-finally.

### DIFF
--- a/src/app/FakeLib/FXCopHelper.fs
+++ b/src/app/FakeLib/FXCopHelper.fs
@@ -88,61 +88,63 @@ let FxCopDefaults =
 let FxCop setParams (assemblies : string seq) = 
     let param = setParams FxCopDefaults
     traceStartTask "FxCop" ""
-    let param = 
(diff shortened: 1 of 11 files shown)


forki:
💋

PR #1381, opened 2016-09-18
Allow to pass parameter to sonarqube end

laedit:
Useful when using SonarQube public server with token.
And add simple method for SonarQubeEnd without arguments.


forki:
cool thx

PR #1383, opened 2016-09-19
NuGetVersion: adds application/xml to request accept header

seburgi:
When trying to get the current version of a NuGet package from a private Artifactory NuGet source, we got the following exception:

	System.Net.WebException: The remote server returned an error: (406) Not Acceptable.
		at System.Net.WebClient.DownloadDataInternal(Uri address, WebRequest& request)
		at System.Net.WebClient.DownloadString(Uri address)
		at System.Net.WebClient.DownloadString(String address)
		at Fake.NuGetVersion.getLastNuGetVersion(String server, String packageName) in C:\code\fake\src\app\FakeLib\NuGet\NugetVersion.fs:line 55

It turned out that Artifactory (as of version 4.2.1) does not support the NuGet json format.
Adding "application/xml" to the accept header of the request, fixes this problem.


forki:
thx

PR #1386, opened 2016-09-20
Allow additional properties to be specified when running the SQL dacpac tooling

bruinbrown:
This PR allows for a list of Key Value pairs to be passed to the sqlpackage exe, allowing for extended configuration.

It also adds one of the keys for specifying the Azure SQL DB size to use

@isaacabraham


forki:
cool!

PR #1389, opened 2016-09-26
wixHelper fix: AllowDowngrades

jbezak:
The MajorUpgrade/@DowngradeErrorMessage attribute cannot be specified when AllowDowngrades is set to 'yes'


forki:
thx!

Issue #822, opened 2015-05-28
Add more data to AssemblyVersionInformation class


amazingant:
I currently populate my AssemblyInfo files with a smattering of data (most of it provided by default with the ProjectScaffold build.fsx, but with a couple extra values tossed in). The helpful `AssemblyVersionInformation` type definition provided with C# and F# classes (I may be mistaken, but I don't see anything generating this class for VB under `CreateVisualBasicAssemblyInfoWithConfig`, despite the comments for it indicating otherwise) currently provides one value.

With a few changes to `Fake.AssemblyInfoFile.Attribute`, and some matching changes in each of the `Create<Language>AssemblyInfoWithConfig` functions, would it be possible to include all of the data populated into the AssemblyInfo file? Hopefully including the metadata values?

I'm happy to prototype & PR the code changes for this, but I'm not sure what documentation would need to be updated with it.

amazingant:
My quick prototype here: https://github.com/fsharp/FAKE/compare/master...amazingant:add-data-to-AssemblyVersionInformation

The full FAKE solution refuses to build on my machine, so I'm not sure if there's any tests I've invalidated, but after swapping out `FakeLib.dll` in the right spots and building one of my projects, it works well:

```
namespace System
open System.Reflection

[<assembly: AssemblyTitleAttribute("Testing")>]
[<assembly: AssemblyProductAttribute("Testing")>]
[<assembly: AssemblyDescriptionAttribute("Testing a change to FAKE")>]
[<assembly: AssemblyMetadataAttribute("githash","7bcb77")>]
[<assembly: AssemblyVersionAttribute("0.1.0")>]
[<assembly: AssemblyFileVersionAttribute("0.1.0")>]
[<assembly: AssemblyInformationalVersionAttribute("0.1.0-beta")>]
do ()

module internal AssemblyVersionInformation =
    let [<Literal>] AssemblyTitle : System.String = "Testing"
    let [<Literal>] AssemblyProduct : System.String = "Testing"
    let [<Literal>] AssemblyDescription : System.String = "Testing a change to FAKE"
    let [<Literal>] AssemblyMetadata_githash : System.String = "7bcb77"
    let [<Literal>] AssemblyVersion : System.String = "0.1.0"
    let [<Literal>] AssemblyFileVersion : System.String = "0.1.0"
    let [<Literal>] AssemblyInformationalVersion : System.String = "0.1.0-beta"
```

[...]

PR #1392, opened 2016-10-03
Add all assembly metadata to AssemblyVersionInformation type

amazingant:
PR for #822 as requested by @dsyme

Replaces the `Version` and `InformationalVersion` values being generated with the `AssemblyVersionInformation` type with a value for every assembly attribute.

As the new values are named after their attribute type names, this is a breaking change for anyone expecting `Version` instead of `AssemblyVersion` and `InformationalVersion` instead of `AssemblyInformationalVersion`.

Likewise, the newly-named `AssemblyVersion` and `AssemblyInformationalVersion` values are no longer generated by default, and must be included in the collection of provided attributes.


forki:
thx

Ravadre:
Hi,

@amazingant - this PR seems to have broken generating AssemblyInfoVersion when Metadata attribute is used.

Before this PR generated code looked like this:
```
// <auto-generated/>
using System.Reflection;

[assembly: AssemblyMetadataAttribute("githash","cc5fd8")]
namespace System {
    internal static class AssemblyVersionInformation {
        internal const string Version = "LocalBuild";
        internal const string InformationalVersion = "LocalBuild";
    }
}
```

After the PR:
```
// <auto-generated/>
using System.Reflection;

[assembly: AssemblyMetadataAttribute("githash","cc5fd8")]
namespace System {
    internal static class AssemblyVersionInformation {
        internal const System.String AssemblyMetadata_githash = "githash","cc5fd8";
    }
}
```

Note, that string const has a value of `"key","value"` which is not valid code. Reason why this is the case seems to lie in L183:
`fun x -> sprintf "        internal const %s %s = %s;" x.StaticPropertyType x.StaticPropertyName x.StaticPropertyValue`

`x.StaticPropertyValue` seems to have value that goes into attribute data, when using most attributes where value is a single string, this is OK, because the property will have value like `"0.2"`, for Metadata though - it will contain 2 string values.

amazingant:
@Ravadre Sorry, forgot to supply a different `StaticPropertyValue` for metadata attributes. Fixed with #1399

Ravadre:
@amazingant Hi, that's great, thanks a lot!

Issue #1395, opened 2016-10-04
Parallel Target JIT?


BlythMeister:
I have the following targets as an example

```
Target "T1" (fun _ ->
    printfn "T1"
    System.Threading.Thread.Sleep(1000)
)

Target "T2.1" (fun _ ->
    printfn "T2.1"
    System.Threading.Thread.Sleep(1000)
)

Target "T2.2" (fun _ ->
    printfn "T2.2"
    System.Threading.Thread.Sleep(1000)
)

Target "T2.3" (fun _ ->
    printfn "T2.3"
    System.Threading.Thread.Sleep(1000)
)

Target "T3" (fun _ ->
    printfn "T3"
    System.Threading.Thread.Sleep(3000)
)

Target "T4" (fun _ ->
    printfn "T4"
    System.Threading.Thread.Sleep(1000)
)

"T1"
    ==> "T2.1"
    ==> "T2.2"
    ==> "T2.3"

"T1"
    ==> "T3"

"T2.3"
    ==> "T4"

"T3"
    ==> "T4"
```

T1 is my base target,
Off of that the T2 chain consists of 3 steps taking 1 second each
The T3 target is a single and takes 3 seconds
T4 will run when the T2 chain and T3 target have completed.

I would expect the total exectution to be 5 seconds:
T1 - 1 second
T3 & T2 Chain - 3 second
T4 - 1 second

However, T3 runs alongside T2.3.
As a result, the total run time is 7 seconds.

Should T3 not start as soon as it is possible to do so, rather than waiting?

The build output is as follows:

```
Building project with version: LocalBuild
Running parallel build with 20 workers
Starting Target: T1
T1
Finished Target: T1
Starting Target: T2.1 (==> T1)
T2.1
Finished Target: T2.1
Starting Target: T2.2 (==> T2.1)
T2.2
Finished Target: T2.2
Starting Target: T3 (==> T1)
Starting Target: T2.3 (==> T2.2)
T3
T2.3
Finished Target: T2.3
Finished Target: T3
Starting Target: T4 (==> T2.3, T3)
T4
Finished Target: T4

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
T1         00:00:01.0015316
T2.1       00:00:01.0009064
T2.2       00:00:01.0009421
T2.3       00:00:01.0007335
T3         00:00:03.0011838
T4         00:00:01.0007754
Total:     00:00:07.0572197
Status:    Ok
[...]

PR #1396, opened 2016-10-07
Partial fix for #1395

BlythMeister:
Targets now fire as early as possible.

However, targets will not move onto the next level until all targets on that level are completed.

Therefore, in the situation described in #1395 although T2.1 and T3 run along side each other, T2.2 waits until the longer running T3 has completed before starting even though it has no dependency on it.

### DIFF
--- a/src/app/FakeLib/TargetHelper.fs
+++ b/src/app/FakeLib/TargetHelper.fs
@@ -471,27 +471,47 @@ let determineBuildOrder (target : string) =
     let appendDepentantOption (currentList:string list) (dependantTarget:option<TargetTemplate<unit>>) = 
         match dependantTarget with
         | None -> currentList
-        | Some x -> List.append currentList [x.Name]
-                    |> List.distinct 
+        | Some x -> List.append currentList [x.Name] |> List.distinct 
 
     let rec SetTargetLevel newLevel target  = 
         match targetLevels.TryGetValue target with
-        | true, exDepenencyLevel -> if exDepenencyLevel.level < newLevel then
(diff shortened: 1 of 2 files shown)


BlythMeister:
Hey @forki (or any other maintainer), any chance of getting this reviewed/merged?

forki:
This looks like it's changing a lot of the internals. That's very risky.

@matthid  what do you think?

BlythMeister:
@forki @matthid I have added additional test coverage of scenarios and happy to add more if you can describe any :)

matthid:
Might be a candidate to rebase on top of FAKE vnext (https://github.com/fsharp/FAKE/pull/1281)?

However, this probably means that is not released anytime soon... As I'm currently quite busy.

matthid:
Otherwise it looks like a logical change to make... Do a lot of people depend on the parallel processing?

BlythMeister:
As a compromise, I could look to make a flag which is passed on command like to run targets as early as possible? 
That way it's not a change unless you opt in for it (which I would then do on my build script)

forki:
I don't think we need a flag. Just enough confidence that it's correct

Am 12.10.2016 19:45 schrieb "Chris Blyth" <notifications@github.com>:

BlythMeister:
Ok, I've tried to create test scenarios to cover basics,  the key to the change is the level adjustments. 
I'm not sure what else I can do to increase confidence,  but if you have anything I can try and implement it.

forki:
we will put that in alpha for next minor version

BlythMeister:
@forki there has been a few minor releases since october..any idea when this can be merge?

PR #1397, opened 2016-10-07
fix AssemblyInfo generation bug

eiriktsarpalis:
Fixes a bug in assembly info generation code which cause compilation to fail if (for example) somebody specifies two `InternalVisibleTo` attributes.


forki:
thx

PR #1406, opened 2016-10-24
Add option to emit SuppressMessage Attributes

magicmonty:
ReSharper outputs a lot of warnings, if a static class is generated in the AssemblyInfoFile.

This option adds a possibility to add the needed SuppressMessage attributes for ReSharper to the static class


forki:
thx

PR #1407, opened 2016-10-25
Add language in NuGetParams

fpellet:
Currently, Language is missing in NuGetParams.
I need this argument to create localized package (https://docs.nuget.org/ndocs/create-packages/creating-localized-packages)


forki:
thx

PR #1408, opened 2016-10-25
traceStartTask / traceEndTask: use 'use' block

0x53A:
So it seems in my previous PR (https://github.com/fsharp/FAKE/pull/1379) I missed a few cases ...

The whole try / finally stuff just adds noise (in my opinion), what do you think about this IDisposable approach?

If you like it, I will go through all files (a full text search for traceStartTask found 114 lines in 59 files, so totally manageable).

### DIFF
--- a/src/app/Fake.IIS/IISExpress.fs
+++ b/src/app/Fake.IIS/IISExpress.fs
@@ -65,15 +65,13 @@ let createConfigFile (name, siteId : int, templateFileName, path, hostName, port
 let HostWebsite setParams configFileName siteId = 
     let parameters = setParams IISExpressDefaults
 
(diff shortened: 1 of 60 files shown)


0x53A:
In many cases it explicitly either closes the task or throws, so maybe this was originally on purpose?

Another approach could be to walk the stack on close, and close everything until the specified tag.
(e.g. if I call open "A" -> open "B" -> close "A", then B is implicitly closed.)

Then I could revert my previous pr and it would "just work".

forki:
I like the idea of the use thingy. Can you add that everywhere?

0x53A:
Ok that should be everything this time

forki:
thanks so much

PR #1409, opened 2016-10-27
getBranchName works language independent

Stift:
this should make the strange situation of i10n-git installations more robust.

Tested with git 1.7 and git 2.10. 

Did not find any localized version of git. I assume the problematic installation is a git-sdk install.


forki:
thanks!

PR #1418, opened 2016-11-11
Disable NodeReuse for MSBuild on Jenkins too

magicmonty:
Disables NodeReuse in MSBuild by default if the build runs on Jenkins


forki:
thanks

Issue #1420, opened 2016-11-16
FAKE.Deploy (Start of process fsi.exe The handle is invalid)


ditadi:
### Description

Deployment fails with stack trace
```
Details:
Failure
  {Messages = seq [];
   Exception =
    System.Exception: Start of process C:\Program Files (x86)\Microsoft SDKs\F#\4.0\Framework\v4.0\fsi.exe failed. The handle is invalid.

   at Fake.ProcessHelper.ExecProcessWithLambdas@87-16.Invoke(String message) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 87
   at Fake.ProcessHelper.ExecProcessWithLambdas(FSharpFunc`2 configProcessStartInfoF, TimeSpan timeOut, Boolean silent, FSharpFunc`2 errorF, FSharpFunc`2 messageF) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 87
   at Fake.ProcessHelper.ExecProcessRedirected(FSharpFunc`2 configProcessStartInfoF, TimeSpan timeOut) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 131
   at Fake.FSIHelper.executeFSIWithScriptArgsAndReturnMessages(String script, String[] scriptArgs) in C:\code\fake\src\app\FakeLib\FSIHelper.fs:line 192
   at Fake.DeploymentHelper.doDeployment(String scriptFileName, String[] scriptArgs) in C:\code\fake\src\app\Fake.Deploy\DeploymentHelper.fs:line 88;
   IsError = true;}
```

### Repro steps

Please provide the steps required to reproduce the problem

1. Download FAKE version 4.44.2+ or later on remote server.

2. Deploy NuGet Package to that remote server.

### Expected behavior

Excepted behavior is successfull remote deployment.

### Actual behavior

Deployment fails and returns the stack trace given at the top of this post.  I modified exception catching code to return more details about the exception on my machine here are the results.

```
Details:
Failure
  {Messages = seq [];
   Exception =
    System.Exception: Start of process C:\Program Files (x86)\Microsoft SDKs\F#\4.0\Framework\v4.0\fsi.exe | 
Start Info:ProcessInfo:
Args:"C:\nuget\FAKE\tools\deployments\DeployPackage\Deploy.fsx " 
Working dir: "" 
UserName: ""
 isMono: false
 | failed. The handle is invalid.
 
[...]

PR #1422, opened 2016-11-17
Added try catch block to ignore error from setting console encoding t…

ditadi:
Pull request for the following issue https://github.com/fsharp/FAKE/issues/1420.


forki:
thanks!!

PR #1425, opened 2016-11-22
Added credentials parameter to ApplicationPoolConfig in IISHelper

mandrek44:
I added possibility to define username and password when `ProcessModelIdentityType` is set to `SpecificUser`


forki:
cool. thanks!

PR #1428, opened 2016-12-02
decorated all *Params helper records with [<CLIMutable>]

filipw:
This PR annotates all `*Params` records in the `FakeLib` helpers with `[<CLIMutable>]`.

This doesn't really affect anything on the F# side, but it makes C# interoperability, and the convenience/experience of consuming `FakeLib` there infinitely better.


forki:
Awesome! Thanks

forki:
c

PR #1429, opened 2016-12-07
Improve error logging in GitVersionHelper

trondd:
While searching for a unresolved bug where gitversion would fail when running on one of your build agents I discovered that the exit code is not of any use since all errors return 1.


forki:
excellent!

PR #1439, opened 2016-12-22
add expecto --summary support

vilinski:
- add "--summary" argument introduced in https://github.com/haf/expecto/pull/19
- also added a CustomArgs property to instantly add not yet supported expecto args. Though not sure whether thats a good idea to do so


forki:
I already added summary

vilinski:
sorry, here is a new PR instead, with CustomArgs only: https://github.com/fsharp/FAKE/pull/1441

PR #1444, opened 2017-01-06
Clarify different roles of different Nuget overloads

piers7:
Without browsing the code, existing documentation is not clear on the differences between Nuget / NugetPack / NugetPackDirectly (same doco comment for each). This edit to the doco comments attempts to clarify the differences.


forki:
Thanks

Issue #1446, opened 2017-01-08
Missing "NuGet.exe" in Getting started tutorial.


Rodhern:
I was following along with the [Getting started with FAKE](http://fsharp.github.io/FAKE/gettingstarted.html) tutorial. Most went fine. At some point the build script failed to find the .\tool\nuget\nuget.exe file, because that directory and file aren't there. Did I inadvertently skip a step in the tutorial?

Rodhern:
Today I went one step further. I added a "http" line in paket.dependencies to download nuget.exe (a paket-files folder is then created to hold the file). Now the targets "Clean", "BuildApp" and "BuildTest" complete successfully. In this case it is the "Test" target that fails; it cannot find the .\tools\Nunit\nunit-console.exe file.

smoothdeveloper:
@Rodhern thanks for reporting the issue, I think the manual was made with NUnit 2 in mind, now the console tool is a separate nuget package.

Maybe you can find this package and adjust the location in the script, if you come up with a working solution, maybe propose changes to the documentation in a PR: https://github.com/fsharp/FAKE/blob/master/help/gettingstarted.md ?

Rodhern:
Thank you for your hint; some playing around with the version numbers did the trick.

I will play just a bit more this week, and then I hope to do a simple PR, that someone with better understanding of the recommended Paket syntax will then have to beautify.

PR #1447, opened 2017-01-11
Cannot find 'nuget.exe' and 'nunit-console.exe' in getting started tutorial (#1446).

Rodhern:
This is a work-around for #1446.
I added two lines to the paket.dependencies file, to get 'nuget.exe' and an older NUnit runner.
My text phrase needs some beautification, but I think it should be clear enough that someone with Fake and Paket skills can use it as a draft.


forki:
Thx

PR #1453, opened 2017-01-17
Visual Studio aware msbuild selection

### LINKED ISSUE #1452: FAKE should select the correct msbuild, based on VisualStudioVersion parameter
When building Visual Studio extensions, msbuild version has an important significance. Each VS version has its own msbuild, and FAKE should invoke the correct one.

Right now, there's a very simple lookup for an installed MSBuild version. While this may work well for most C#/F# projects, it doesn't work with version-specific projects, such as Visual Studio extensions.

VS-specific FAKE scripts pass in the `VisualStudioVersion` parameter (either taken from the environment variables of the Developer Command Shell, or a default value) - see https://github.com/fsprojects/Paket.VisualStudio/blob/master/build.fsx#L105

FAKE should consider `VisualStudioVersion` parameter, if exists, when looking for the location of msbuild.exe

hmemcpy:
This adds support for the `VisualStudioVersion` environment variable (which is specified by VS/developer command prompt), so that a compatible msbuild version will be used.
This fixes #1452

### DIFF
--- a/src/app/FakeLib/MSBuildHelper.fs
+++ b/src/app/FakeLib/MSBuildHelper.fs
@@ -16,29 +16,58 @@ exception BuildException of string*list<string>
   with
     override x.ToString() = x.Data0.ToString() + "\r\n" + (separated "\r\n" x.Data1)
 
+
+type MsBuildEntry = {
+    Version: string;
+    Paths: string list;
+}
+
+let knownMsBuildEntries = 
+    [
+        { Version = "12.0"; Paths = [@"\MSBuild\12.0\Bin"; @"\MSBuild\12.0\Bin\amd64"] }
+        { Version = "14.0"; Paths = [@"\MSBuild\14.0\Bin"] }
+        { Version = "15.0"; Paths = [@"\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin"; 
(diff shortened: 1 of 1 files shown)


hmemcpy:
Code reviews welcome :D

src/app/FakeLib/MSBuildHelper.fs
```diff
@@ -16,29 +16,58 @@ exception BuildException of string*list<string>
...
+                                     @"\MSBuild\15.0\Bin"] }
+    ]
+
+let oldMsBuildLocations = 
+    [ @"c:\Windows\Microsoft.NET\Framework\v4.0.30319\"; 
+      @"c:\Windows\Microsoft.NET\Framework\v4.0.30128\";
+      @"c:\Windows\Microsoft.NET\Framework\v3.5\"
+    ]    
+
+let toDict items =
```
forki: can we call it toMap - since there is a dict function in FSharp.Core that actually put things in a dict

forki:
cool. does it fall back to the old behaviour?

hmemcpy:
Yup, if it fails to find `VisualStudioVersion` it will just list all the paths, like before.

forki:
cool!

hmemcpy:
Yay! Thank you!

forki:
should be released. Thank you!

PR #1467, opened 2017-02-07
Change log helper

magicmonty:
Added a helper for parsing Change logs in the format as described under http://keepachangelog.com/en/0.3.0/


forki:
We already have a ReleaseNotes helper. How does this relate?

magicmonty:
I know, but the ReleaseNotes helper parses a different format, which is incompatible to KeepAChangeLog (no sub headings, error on primary heading without version number etc,)

forki:
ok cool

Issue #1468, opened 2017-02-07
Output dependencies in graphviz format


otto-gebb:
### Description

Implement a command line option allowing to output the graph of target dependencies in the [DOT](http://www.graphviz.org/doc/info/lang.html) format.

### Example

Dependencies in the FAKE-script:

```
"BuildJs"
  ==> "WebPackage" <=> "AdminPackage" <=> "WebApiPackage"
  ==> "MainAppPackage"
```

Expected output:

```
digraph G {
{"WebPackage"  "AdminPackage"  "WebApiPackage"} -> "BuildJs"
"MainAppPackage" -> {"WebPackage"  "AdminPackage"  "WebApiPackage"}
}
```

... which would be rendered as this:

![image](https://cloud.githubusercontent.com/assets/5798003/22689613/0a7a5960-ed32-11e6-8e91-4f1a9a471360.png)

PR #1469, opened 2017-02-09
Implement --dotGraph command line option.

otto-gebb:
With this option specified, FAKE will output a graph of
target dependencies in the DOT format instead of building
anything, provided the build script actually tries to build
the target specified in the command line,
i.e. contains a call like `RunTargetOrDefault "Default"`.

This is how to render the graph from PowerShell:

```
FAKE.exe build.fsx -dg | & 'C:\Program Files (x86)\Graphviz2.38\bin\dot.exe' -Tpng -o ./graph.png
```

(For this command to work, Graphviz must be installed.)

See #1468.


forki:
can we please have a bit of documentation for this?

forki:
cool thanks!

PR #1470, opened 2017-02-11
FSharp.Compiler.Service 10.0.1

forki:
Trying to update FAKE to FSharp.Compiler.Service 10.0.1

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+#### 4.51.0 - 11.02.2017
+* Update FSharp.Compiler.Service 10.0.1
+
 #### 4.50.0 - 17.01.2017
 * Visual Studio aware msbuild selection - https://github.com/fsharp/FAKE/pull/1453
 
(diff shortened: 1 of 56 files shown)


forki:
@matthid  @dsyme any ideas why the resolution breaks?

matthid:
I think we make some assumptions in the caching code and FSharp.Compiler.Service broke at least some of them. For example the code seems to generate a differently named "entry" method (might be the wrong word) now:
https://travis-ci.org/fsharp/FAKE/builds/200609285#L2450 (I'm not sure why we only see this on travis...)

forki:
I see same thing locally and I'm not sure what to do

matthid:
https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/FSIHelper.fs#L288-L306 and https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/FSIHelper.fs#L317 might need to be updated. I don't know where the corresponding code in FCS sits. 

Can you look at the generated dll file with reflector? Maybe they added an API for that so we don't need to manipulate strings anymore? /cc @dsyme

matthid:
@forki  did you figure it out?

forki:
nope

2017-02-20 15:47 GMT+01:00 Matthias Dittrich <notifications@github.com>:

PR #1472, opened 2017-02-24
Git: recognize renamed (and other status) files

### LINKED ISSUE #1458: [Git.FileStatus] Failure to report renamed files
### Description

`Git.FileStatus`[`.getChangedFiles`] does not report renamed files but fails with `Unknown file status`

### Repro steps

1. Rename a file in git
2. Call `Git.FileStatus.getChangedFiles "."  <currentHash> ""`

### Expected behavior

`Git.FileStatus.getChangedFiles` reports renamed file as either:
- Renamed
- Deleted + Added

### Actual behavior

Fails with exception

### Known workarounds

None

### Related information 

My current use-case is not depending on either having Renamed or Deleted / Added but rather not failing. I favor adding a Renamed case to FileInfo.

CaringDev:
*This is a **breaking change**!*

Recognize renamed (and other less wellknown) files instead of bailing out with "Unknown file status"
fixes #1458 
According to [git docs](https://git-scm.com/docs/git-diff#git-diff---diff-filterACDMRTUXB82308203).

### DIFF
--- a/src/app/FakeLib/Git/FileStatus.fs
+++ b/src/app/FakeLib/Git/FileStatus.fs
@@ -6,29 +6,26 @@ open Fake
 open System
 open System.IO
 
-let private (|Status|_|) expected (actualWithCode:string) =
-    if actualWithCode.StartsWith expected then
-        let success, code = Int32.TryParse(actualWithCode.Substring(expected.Length))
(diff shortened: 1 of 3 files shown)


forki:
cool thanks

PR #1476, opened 2017-02-28
Improve error handling on SqlPackage.

isaacabraham:
* If custom path to SQL Package is used, show that one if it is invalid.
* If error occurs during SQL Package deployment, do not output error to stderr.


isaacabraham:
Don't know why Travis failed - looks fine in AV.

PR #1482, opened 2017-03-07
fix bug in getLastNuGetVersion when result is in JSON.

MaxDeg:
if multiple package hit the searchTerm, the correct one is not always selected.


forki:
thx

PR #1483, opened 2017-03-08
Add property for output path to place to built artifacts

abunk:
Add property for output path to place to built artifacts


forki:
thx.

 this was already added as "Output" parameter in a recent PR.

PR #1487, opened 2017-03-12
HockeyApp - create version and upload build to a specific version

MrLuje:
Add method HockeyAppVersion to create a new version of a App
Add VersionId parameter to HockeyApp method to upload the build to a given version of an App


forki:
thanks!

PR #1490, opened 2017-03-13
Improve log integration with AppVeyor

csmager:
There were a couple of fairly inconsequential and primarily cosmetic issues that  I'd noticed when AppVeyor runs a FAKE build script.

At the start, FAKE calls `PrintDependencyGraph`. This is written as multiple `LogMessages`, which the AppVeyor helper also writes to AppVeyor's message log in the Messages tab.  There are two issues that arise from this:

1. Adding a message to AppVeyor seems to be processed asynchronously. As a result, there is no ordering guaranteed.  Here's a snippet from a log for a failed build:

image

2. The call to `appveyor AddMessage` is included in a trace. This seems to be a little redundant:

image

3. The call to `appveyor AddMessage` fails when the message is empty:

image

4. The console output from FAKE is redirected by AppVeyor and logged. This loses all the colouring, so the log appears entirely in white.  However, AppVeyor supports ANSI color codes. 

The changes I've made address each of those points:

1. I've changed the dependency message to be written as a single multi-line message so this appears as a single message in the log on AppVeyor. This makes no difference when written to the console.

2. I've changed to disable tracing when sending a message to the AppVeyor message log.

3. I've changed to not send empty messages to the AppVeyor message log, avoiding the error that results.

4. I've added ANSI color code support and use this when we're running on AppVeyor. If other build servers have the same issue & support ANSI colors then perhaps this could be enabled for those too. I ran a quick test for all `ConsoleColor` values to show how these are rendered on AppVeyor:

image

### DIFF
--- a/src/app/FakeLib/AppVeyor.fs
+++ b/src/app/FakeLib/AppVeyor.fs
@@ -5,68 +5,68 @@ open System
 open System.IO
 
 /// AppVeyor environment variables as [described](http://www.appveyor.com/docs/environment-variables)
-type AppVeyorEnvironment = 
-    
+type AppVeyorEnvironment =
+
     /// AppVeyor Build Agent API URL
     static member ApiUrl = environVar "APPVEYOR_API_URL"
 
     /// AppVeyor Account Name
     static member AccountName = environVar "APPVEYOR_ACCOUNT_NAME"
-    
+
     /// AppVeyor unique project ID
     static member ProjectId = environVar "APPVEYOR_PROJECT_ID"
-    
+
     /// Project name
     static member ProjectName = environVar "APPVEYOR_PROJECT_NAME"
-    
+
     /// Project slug (as seen in project details URL)
     static member ProjectSlug = environVar "APPVEYOR_PROJECT_SLUG"
-    
+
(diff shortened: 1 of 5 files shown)


forki:
build is red. can you please check?

csmager:
@forki just looking at it now. I had tested the build on one of my projects and it works as I'd expect.

The error is in one of the tests and relates to an invalid char - the char is part of the escape sequence for ANSI color codes. As they tests are running on AppVeyor, I guess `buildServer` will be set to AppVeyor. It'll probably have something to do with that.

csmager:
@forki Issue seemed to be that because we were running in AppVeyor, the ANSI color codes were written to the trace (per my change). MSpec reports all the test results to HTML, and the escape sequence contains a character that's not valid in HTML. 

I changed to essentially toggle that option off in the class that contained the failing spec in an `Establish` function. I've not used MSpec before, so there may be a more appropriate way of doing that.

csmager:
Eurgh, my VS code settings have trimmed a load of whitespace in the files I've edited. If you want me to revert those, just shout.

forki:
thanks!

PR #1495, opened 2017-03-17
Feature/add f sharp41 to fsi

kaeedo:
On a clean install of windows, and only installing Visual Studio 2017, F# installs version 4.1. FAKE doesn't look for F# in the 4.1 directory, only 4.0. This add the 4.1 directory to the list of locations to search


forki:
thx

Issue #765, opened 2015-04-23
Azure StorageEmulator functionality does not work on latest SDK


isaacabraham:
It looks like the StartStorageEmulator path is hard coded to a specific path i.e.

```C:\Program Files\Microsoft SDKs\Windows Azure\Emulator\csrun.exe```

on the latest SDK, on 64 bit Windows, it should be 

```C:\Program Files (x86)\Microsoft SDKs\Azure\Storage Emulator\WAStorageEmulator.exe```

dsyme:
Closing old inactive suggestion. Please send a PR for this if you'd like the issue addressed.

PR #1499, opened 2017-03-22
Fix Azure Storage Emulator path and arguments

theprash:
This is a fix for #765. cc @isaacabraham

I have removed fields from the `AzureEmulatorParams` record because they did not make sense as parameters so this is technically a breaking change. However, none of the functions here actually take an instance of this record (or a function transforming it) so it was effectively useless being exposed as a public type. Given this fact, and since I was already changing the record, I made it private.


forki:
great! thx

PR #1502, opened 2017-03-26
Make ExitCode variable

b0urb4k1:
There is an exitCode variable in TargetHelper bug since it is set as a last function in the finally block you are not able to ever use the value to return anything else than the default value of 42. The newly introduced ExitCode.Value will be set as a value for in case of an error if there were errors found running a target now.


forki:
thx

PR #1503, opened 2017-03-27
TeamCityHelper report build status incorrect string

BlythMeister:
The status needs to be prefixed "status="

Workaround we are currently using is to pass this function "status=myStatus" rather than just "myStatus"


forki:
thanks!

PR #1507, opened 2017-03-28
URGENT: Bug Fix For Parallel From #1396

BlythMeister:
Upon testing #1396 with a complex target order from my company, i discovered issues with ordering of targets so a target could run at the same time or before it's dependency.

This change includes my complex target order as a test case also shown as flow below:
image


BlythMeister:
Wow that was quick ;)

PR #1508, opened 2017-03-28
URGENT: Another (sorry) Parallel Build Fix

BlythMeister:
Turns out walking trees is hard!

Thought up another scenario to test and found another issue (sorry)
Added a new test of this scenario and a fix to make it work.


BlythMeister:
@forki Sorry to keep messing you about with this :(

PR #1512, opened 2017-03-31
Add doc detail to combinePaths

theprash:
cc @rmunn

Helps with a pain point expressed in #1485


forki:
thanks for taking care of this.

PR #1513, opened 2017-04-05
Added a way to set the --trace parameter on the NUnit3 command line

julienadam:
This controls the level of internal logs NUnit3 will output.

(Sorry for the whitespace diffs, just the TrailingWhitespaceVisualizer VS addin doing its thing)


forki:
great! thanks

Issue #1515, opened 2017-04-10
Wrong dependency order after updating to the latest Fake causes tests to be run before built


ovatsus:
### Description

Before:

![image](https://cloud.githubusercontent.com/assets/738761/24885799/652c09f2-1e48-11e7-8387-74d0170694db.png)

After
![image](https://cloud.githubusercontent.com/assets/738761/24885748/215d77ec-1e48-11e7-8826-3dee43e3ab04.png)

### Repro steps

* Run `build All` on the current master of FSharp.Data - https://github.com/fsharp/FSharp.Data/commit/4012bd983932540bf08d98dd075169581f0c0053 - everything works as expected (first screenshot)

* Run `paket update` - https://github.com/ovatsus/FSharp.Data/commit/2571a154aef88202b6855f88f44f297131a9ac91 - wrong dependency order (second screenshot)

forki:
@BlythMeister can you please take a quick look at this? I assume it's related to your recent changes.

BlythMeister:
Yeah sure, can you please provide your dependency trees from the fsx file.

forki:
Should be at the bottom of https://github.com/fsharp/FSharp.Data/blob/master/build.fsx

BlythMeister:
Yeah just realised that.
From what i can see, there is no dependency between build tests, build and run tests.
Therefore that dependency structure states these things can run in any order.

If you have a dependency between targets you should state that in the dependencies for the build.

forki:
I think the problem is at https://github.com/fsharp/FSharp.Data/blob/master/build.fsx#L126

@ovatsus you define that the targets should run before the runtest target. That condition is fulfilled. I think want you want is to reverse the arrow.

forki:
@BlythMeister yes it seems to be unspecified in the script. I will send a pull request to fsharp.data. But it's still a bit unfortunate that we changed behavior here.

BlythMeister:
Agreed, it's a shame the behaviour changed.  But arguably it's now correct in that it obeys the dependencies you set in your script rather than being down to chance/ordering.

If you don't state that b is dependent on a, then that states they can run in any order.

forki:
It obeyed it before as well.

BlythMeister:
True, but a lot of it came down to the order dependency was specified rather than making targets actually dependant.

I can try and make those orders be the same without having to make the change?
Would be an easy test case to add.

forki:
I just looked at @ovatsus's project
[...]

PR #1516, opened 2017-04-11
Revert single thread build order change

BlythMeister:
Fix for #1515

It might be worth testing this PR on FSharp.Data before merging to ensure it's a successful revert?


forki:
thx

BlythMeister:
@forki have another PR which means we can have a single "determinebuildorder" which detemines differently depending on parallel or not

But means the logic for that is encapsulated.

Also, means we can add tests for single build order verification :)

PR #1517, opened 2017-04-11
Consolidate logic for build ordering into a single determineBuildOrder function

BlythMeister:
This enables tests to be written for parallel and non parallel build orders.
Also added bug scenario from https://github.com/fsharp/FSharp.Data to ensure fixes are correct


BlythMeister:
@forki this should help us be able to modify build targets with an extra degree of certainty over the effects it has.

It also means if we get more weird and wonderful examples of targets, we can create a test case for them to build a regression set :)

I've updated the tests to include my repo as a single worker, and also the fsharp.data ordering.

forki:
thanks for helping!

PR #1518, opened 2017-04-11
Fix duplicate build order output

BlythMeister:
The build order is currently output twice.
This appears to be a bug added as part of the appveyor logging PR


BlythMeister:
hold up, this is not ready :(

BlythMeister:
@forki this is now ready 👍

BlythMeister:
Example output:

image

PR #1519, opened 2017-04-11
Update the PrintRunningOrder to mirror logic in #1490

BlythMeister:
This should make logging work on appveyor

@forki sorry...i'm an idiot and didn't realise i didn't push this into last PR :(


BlythMeister:
@forki i'm going to stop changing things now :p

PR #1526, opened 2017-04-23
StyleCop support

stijnmoreels:
Added **StyleCop** to the modules.

I was wondering if we could add some *Integration Tests* for the different modules since a lot write to the file system and I don't think there is now any kind of such testing approach.


forki:
This is awesome. I remove the autoopen and added some tracing but otherwise it's very nice!

matthid:
Why do we need those binaries?

forki:
Urgs I missed these.

Am 27.04.2017 19:41 schrieb "Matthias Dittrich" <notifications@github.com>:

forki:
@stijnmoreels could you please come up with another pr that removes the binaries? Thx

stijnmoreels:
I didn't find them on NuGet, so my first thought was to include them in Source Control.

matthid:
https://www.nuget.org/packages/StyleCop/ ?

stijnmoreels:
No, it's about the other file.

matthid:
Why do we need the other one and what is it doing?

stijnmoreels:
We need it get recursive all the source files from projects or solutions. This way, we can retrieve the source files and make Code Project classes of them.

matthid:
Can we ignore this problem and let the user decide via globbing for example? Where is this dll coming from?

stijnmoreels:
I could come up with some solution that analyses the solution an project files another way. But I have to search for any other library that would allow me to do that.

Any help with that would be appreciated. Via globbing/regex... could be a possible solution but there's already many helpers in this repository that uses the solution or project files. That's way it made sense to me to use these.

The StyleCop helper I wrote is the remake of the StyleCopCLI repository that allows StyleCop analysis via a standard cmd. I saw they used this VSFile.dll for decomposing the solution and project files.

Thanks in advance!

matthid:
@stijnmoreels Yeah we have a lot of helpers which work on solutions, but on the other hand we have a lot of helpers working on globbing files as well. I'd rather change this than to have this binary in the repository.

Maybe we can ask the StyleCop guys where this `VSFile.dll` comes from and if they can nugetize it.
In worst case you can just call the StyleCop cli from FAKE and get rid of the dependency completely (we just tell the user to provide a stylecop executable in PATH)

JohanLarsson:
@stijnmoreels are you aware of [StyleCop.Analyzers](https://www.nuget.org/packages/StyleCop.Analyzers/)?

Issue #1528, opened 2017-04-25
FtpHelper.isFolderPresent returns false even tough the folder exists on the ftp server


bsteinemann:
### Description

I want to use the FtpHelper to deploy my build output to my web server. But the FtpHelper seems to be broken.  I tried to use the function `FtpHelper.uploadAFolder` which did only work when I uploaded the folder into the root directory or the one level below (`/subdirectory`).  

I received the following error message:
~~~
System.Net.WebException: Der Remoteserver hat einen Fehler zurückgegeben: (550)
Datei nicht verfügbar (z.B. nicht gefunden oder kein Zugriff).
~~~

I tracked the problem down to the method FtpHelper.isFolderPresent which is returning false even tough the folder exists on the server. 

~~~F#
// returns false even tough the folder exists
FtpHelper.isFolderPresent  server username pwd  "/upload/myfolder"
~~~

### Repro steps

~~~F#
FtpHelper.createAFolder server username pwd "/upload"
FtpHelper.createAFolder server username pwd "/upload/mysubdir"
sprintf "/upload/mysubdir exists: %b" (FtpHelper.isFolderPresent server username pwd "/upload/mysubdir")|> trace
~~~

Output:
~~~
Creating folder /upload
Getting FTP dir contents for
Create folder status: 257 "//upload" directory created.

Creating folder /upload/mysubdir
Getting FTP dir contents for
Create folder status: 257 "//upload/mysubdir" directory created.

Getting FTP dir contents for
/upload/mysubdir exists: false
~~~

### Expected behavior

~~~F#
// should return true because the folder exists
FtpHelper.isFolderPresent  server username pwd  "/upload/mysubdir"
~~~

### Actual behavior
~~~F#
// returns false but the folder exists
FtpHelper.isFolderPresent  server username pwd  "/upload/mysubdir"
~~~

### Known workarounds

I copied the following code into my F# script:

~~~F#
let inline regexCheck fname ftpContents = Regex.IsMatch(ftpContents, (sprintf @"\s+%s\s+" fname))
let inline private lastSlashPos (str : string) = str.LastIndexOf(@"\") + 1
let inline private getSubstring (fromPos : int) (str : string) (toPos : int) = str.Substring(fromPos, toPos)

let private checkInExistingDirList server user pwd destPath fname = 
    destPath
    |> lastSlashPos
    |> getSubstring 0 destPath
    |> getFtpDirContents server user pwd
    |> regexCheck fname

[...]

PR #1529, opened 2017-04-25
fix FtpHelper.isFolderPresent

bsteinemann:
- lastSlashPos returned the position of last backslash instead of the last slash

Issue: #1528


forki:
thx

Issue #1605, opened 2017-06-28
Setting System.Console.OutputEncoding causes problems with Node.js


inosik:
### Description

Setting the output encoding before running a Node.js process lets the process crash, if the JS prints Unicode characters (anything other than latin letters?) to the console.

*Related: [fable-compiler/fable-react_native-demo#20](https://github.com/fable-compiler/fable-react_native-demo/pull/20#issuecomment-311345571), #1461, #1215.*

### Repro steps

- Create `build.fsx`:

  ``` fsharp
  #r "FakeLib.dll"

  open Fake
  open System

  Target "Default" (fun _ ->
    ExecProcess (fun psi ->
      // I'm using node from PATH here, but that shouldn't matter
      psi.FileName <- "node"
      psi.Arguments <- "test.js")
      (TimeSpan.FromSeconds 2.0)
    |> ignore
  )

  Run "Default"
  ```

- Create `test.js`:

  ``` js
  console.log("© ACME 2017");
  ```

- Run `fake build.fsx`

### Expected behavior

FAKE should launch Node.js, which then should print `© ACME 2017` to the console.

### Actual behavior

```
Building project with version: LocalBuild
Shortened DependencyGraph for Target Default:
<== Default

The running order is:
  - Default
Running build with 1 worker
Starting Target: Default
node app.js
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: write EIO
    at exports._errnoException (util.js:1018:11)
    at WriteStream.Socket._writeGeneric (net.js:711:26)
    at WriteStream.Socket._write (net.js:730:8)
    at doWrite (_stream_writable.js:331:12)
    at writeOrBuffer (_stream_writable.js:317:5)
    at WriteStream.Writable.write (_stream_writable.js:243:11)
    at WriteStream.Socket.write (net.js:657:40)
    at Console.log (console.js:43:16)
    at Object.<anonymous> (D:\projects\FAKE\temp\test.js:1:71)
    at Module._compile (module.js:570:32)
Finished Target: Default

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Default    00:00:00.1977308
Total:     00:00:00.2934179
---------------------------------------------------------------------
Status:    Ok
---------------------------------------------------------------------
```

### Known workarounds

Use an older FAKE version, prior to merging #1215.

### Related information 

* Operating system: Windows (I think that Linux or macOS aren't affected)
[...]

PR #1606, opened 2017-06-29
Allow to set process encoding

forki:
fixes #1215


inosik:
That should work. Would also fix #1605, I guess.

PR #1636, opened 2017-08-20
Restore VS2015 support to open FAKE.sln

dedale:
Because we may not want to use VS2017 yet...

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+#### 5.0.0-alpha015 - 27.08.2017
+* Update Paket.core
+
 #### 5.0.0-alpha014 - 28.07.2017
 * Disable MSBuild NodeReuse by default - https://github.com/fsharp/FAKE/pull/1600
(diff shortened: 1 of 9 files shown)


matthid:
Oh nice, thanks

dedale:
CI fails (even without my change) with 

> Unable to cast object of type 'Newtonsoft.Json.Linq.JValue' to type 'Newtonsoft.Json.Linq.JArray'.

(Well I see other error message, but I guess this is the problem).
Will try to repro on my machine, but I'm a newbie in the magic world of .NET core...

matthid:
Yes your change looked fine. Will take a look soon when. I merge and test the open PRs. Sorry maybe the new sdk release broke something here as well...

dedale:
Thx, no hurry

matthid:
Hm, seems like github couldn't handle what I was doing. Basically, I removed your revert commit and merged the PR. But it was too much for github, sorry

dedale:
No problem, thx

PR #1658, opened 2017-09-11
Target: update DoNothing signature

iainnicol:
Update the signature to match the new signature of Target.Create.


iainnicol:
I wouldn't object to deleting it, given that FAKE 5 is in alpha or beta.

However a straw man in favor of keeping it: ignore might sound to beginners like we're ignoring the target entirely. But we don't ignore it: dependent targets are processed.

PR #1663, opened 2017-09-17
Fixed the tooltip position and improved the styling in the api reference pages

seanamos:
The tooltip position was broken in the API reference docs and tables didn't have styling:
Current

I've made some changes to fix the tooltip position and styled the `code` and `table` elements:

Changes

This was just something small that was bothering me when searching the docs.

### DIFF
--- a/help/content/assets/css/style.css
+++ b/help/content/assets/css/style.css
@@ -208,7 +208,13 @@ td.title, thead tr td {
 .github-link .normal { display: block; }
 .github-link:hover .normal { display: none; }
 
-.member-name code {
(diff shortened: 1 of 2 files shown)


matthid:
Help here is VERY welcome, thanks!

seanamos:
I've pushed a small change because it was a quick win while I was making changes here.

It reduces the size or the "Parameters" and "Samples" headings. It also styles the actual parameters which were in "code" elements.
Original:
Original

Updated:
Updated

matthid:
Thanks again.

PR #1754, opened 2018-01-03
GetDotNetSDKVersionFromGlobalJson

forki:


tlycken:
How will this handle the (pretty common) case when no `global.json` exists (because the user wants to just use latest)?

matthid:
@tlycken I guess you check if the file exists before using this new API? Or you assume the file is there when using the API?

tlycken:
Ah, yeah, that works. But if so, it would probably be more usable if it returned an `option` with the version (if a `global.json` was found, and `None` if it wasn't) rather than failing. That way one could call this, and provide a fallback (e.g. a hard-coded value) if no `global.json` is found.

Also, should this traverse parent directories? According to [the docs](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json), valid locations are the current directory or any ancestor directory up to the drive root.

forki:
I was told by the dotnet cli folks that it is good practice to have that
global.json file. But yes it should only be used in projects that actually
use global.json

Am 03.01.2018 17:59 schrieb "Tomas Lycken" <notifications@github.com>:

tlycken:
The `global.json` acts (sort-of) like `paket.lock` for the .NET SDK; with it, you can be sure that all developers (and build servers, etc) use the same .NET version, while without it, it just uses the latest.

Even though pinning it is probably a good idea as soon as a project is moderately sized with a team of more than one, i think FAKE should support not having one, with a nice and simple API, since otherwise we're imposing arbitrary requirements on your `dotnet` configuration that `dotnet` doesn't impose.

matthid:
> Even though pinning it is probably a good idea as soon as a project is moderately sized with a team of more than one, i think FAKE should support not having one, with a nice and simple API, since otherwise we're imposing arbitrary requirements on your dotnet configuration that dotnet doesn't impose.

@tlycken Yes, however an API called `GetDotNetSDKVersionFromGlobalJson` can only throw an exception if the requirement (`global.json`) is not fullfilled. IMHO there is nothing else meaningful such an API can do. Which version has any meaning in that scenario? People without this requirement just don't call this function.

I feel like either people want to lock down the version and have an error if non is given (this API) or they don't.

tlycken:
@matthid I guess what I'm after is an api that lets me figure out which dotnet version to install, that would respect global.json if it exists, but not blow up if it doesn't (rather figure out which is the latest, and return that). But you're right; that would probably be better to implement with the help of this function, rather than as part of it.

matthid:
@tlycken Yes everything is a mess - feel free to suggest an API for this via PR ;)

PR #1811, opened 2018-03-06
Added RunCodeunitWithSettings for DynamicsNAV

part-timeDev:
Added the ability to use a specific ClientUserSettings.config to avoid the warning message on client start if the parameters in the url  differ from the settings of the default config


forki:
do we need it for OpenPage as well?

part-timeDev:
Well ... OpenPage uses the dynamicsnav-protocol, so the protocolhandler kicks in    
the protocol handler does not support a config file, no chance to use it without some kind of registry magic beforehand

PR #1871, opened 2018-04-15
Fake.DotNet.Testing.Expecto

jackfoxy:
Does not support netstandard1.6
netstandard1.6 does not support ProcessStartInfo.EnvironmentVariables, and I don't think the effort at rewrite is justified


jackfoxy:
@matthid  how exactly do you envision that working? run returns ().

jackfoxy:
@matthid  if you like this https://github.com/jackfoxy/FAKE/blob/8405a1621ec2d9d5521bc31abcb8dc3b4a7fa51a/src/app/Fake.DotNet.Testing.Expecto/Expecto.fs#L96 better, I will merge into the PR.

Add a timeout TimeSpan to the ExpectoParams?

matthid:
Yes I think it is better. I don't think we need timeout. I just wanted to say that using the Fake.Core.Process module has some advantages over raw framework APIs.

Btw do we need to redirect output of expecto? I think maybe `Process.execSimple` is enough here?

jackfoxy:
@matthid  PR updated
I'm suffering under the slow intellisense resolution of VS, which makes API discovery clunky. I may have to switch to Ionide, like everyone else.

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
```diff
@@ -0,0 +1,135 @@
+/// Contains tasks to run [Expecto](https://github.com/haf/expecto) unit tests.
+module Fake.DotNet.Testing.Expecto
+
+open Fake.Core
+open Fake.Core.Environment
+open Fake.Core.StringBuilder
```
matthid: `RequireQualifiedAccess` was added to StringBuilder in rc005

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
```diff
@@ -0,0 +1,135 @@
+/// Contains tasks to run [Expecto](https://github.com/haf/expecto) unit tests.
+module Fake.DotNet.Testing.Expecto
```
matthid: I think we should add `RequireQualifiedAccess` here

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
```diff
@@ -0,0 +1,135 @@
...
+
+open Fake.Core
+open Fake.Core.Environment
```
matthid: `RequireQualifiedAccess` was added to Environment in rc005

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
```diff
@@ -0,0 +1,135 @@
...
+                    then args.WorkingDirectory else Fake.IO.Path.getDirectory testAssembly
+                (fun (info: ProcStartInfo) ->
+                    { info with 
+                        FileName = testAssembly
+                        Arguments = string args
+                        WorkingDirectory = workingDir 
+                        UseShellExecute = false 
+                        // Pass environment variables to the expecto console process in order to let it detect it's running on TeamCity
+                        // (it checks TEAMCITY_PROJECT_NAME <> null specifically).
+                        Environment = environVars() |> Map } )
```
matthid: Can we just leave it out? I think this is the default.

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
```diff
@@ -0,0 +1,135 @@
...
+module Fake.DotNet.Testing.Expecto
+
+open Fake.Core
+open Fake.Core.Environment
+open Fake.Core.StringBuilder
+open Fake.Testing.Common
+open System
+
+/// CLI parameters available if you use Tests.runTestsInAssembly defaultConfig argv in your code:
+type ExpectoParams =
```
matthid: As it is part of the `Expecto`-module we could 
- either call it `Params` (if we want to, users would then use `Expecto.Params`)
- move it out of the `Expecto` module and leave it `ExpectoParams`

What do you think? We probably even should have some guideline around this and make it more uniform in fake 6

jackfoxy: Shortening to Params.

src/app/Fake.DotNet.Testing.Expecto/Expecto.fs
jackfoxy: I think more than happened to ````RequireQualifiedAccess```` Environment. 

I now get this runtime error

````Your environment variables look like they are set manually, but you are missing the default variables. Use the `Process.` helpers to change the 'Environment' field to inherit default values! See https://github.com/fsharp/FAKE/issues/1776#issuecomment-365431982````

Comment (matthid) [reply src/app/Fake.DotNet.Testing.Expecto/Expecto.fs]: This was introduced somewhere in the beta phase. This will prevent incorrect usage of the API. Basically you almost always want to inherit the default environment variables, but you replaced them with your own map -> this error occurs.

File: src/app/Fake.DotNet.Testing.Expecto/Expecto.fs [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+                    if String.isNotNullOrEmpty args.WorkingDirectory
+                    then args.WorkingDirectory else Fake.IO.Path.getDirectory testAssembly
+                (fun (info: ProcStartInfo) ->
+                    { info with 
+                        FileName = testAssembly
+                        Arguments = string args
+                        WorkingDirectory = workingDir } 
+                    // Pass environment variables to the expecto console process in order to let it detect it's running on TeamCity
+                    // (it checks TEAMCITY_PROJECT_NAME <> null specifically).
+                    |> Process.setEnvironmentVariables (Environment.environVars()) )
```
matthid: It should be working by just deleting this like. AFAICS you don't add any special environment variables or remove them, correct?

jackfoxy: I just copied over this from the legacy code based on the comment. Removed.

jackfoxy:
@matthid perhaps the comment about travis and environment variables is to be taken seriously?

jackfoxy:
I really could not tell from the raw travis log why it failed.

jackfoxy:
@matthid  the best I can determine from the Travis log is it stalled for unknown (to me) reason.

````
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
````

matthid:
I'll try to figure out what's wrong

PR #1876, opened 2018-04-19
Port ChangeLogHelper to Fake.Core.Changelog

magicmonty:
I just ported my ChangeLogHelper to FAKE 5 and made a new Package Fake.Core.Changelog for it.
I also fixed some bugs I encountered while using it (I'm the original author of the ChangeLogHelper)


matthid:
I'm in principle OK with adding this, but first I'd like to ask why we cannot unify with Fake.Core.ReleaseNotes. What is this doing differently?

kblohm:
It would also be nice to be able to configure what indicates an added item or a changed item etc. That would be useful if you write your changelogs in another language. The same goes for the kind of header u use. Just today i could not use the releasenotes module because we format our changelogs with h1 headers and not h2. 
Just an idea if you have time, otherwise i might do that later.

magicmonty:
The format is a kind of fixed format which is specified at https://keepachangelog.com/en/1.0.0/

So the supported ReleaseNotes have a much simpler format than this one.
In theory I'm fine with merging this with Fake.Core.ReleaseNotes but to distinguish the formats I would prefer to leave it in its own module/namespace so we could use `Changelog.load` instead of `ReleaseNotes.loadChangelog` what seems to be against the new API guidelines IMHO.

@kblohm how should this look like? Do you want a ChangelogParams type here, which can be filled with different headers? What would be the use case here? I could only imagine translation. Also there is also a `Custom` ChangelogEntry, which contains a custom header and its description.

matthid:
@magicmonty I was more asking about what is actually different between the formats, couldn't we detect both and handle them via the same api?

kblohm:
@magicmonty Yes i did mean for translation. Maybe that is not too important but i could not use the module right now because our changelogs are in german.

magicmonty:
@matthid The format is much more detailed than the already supported release notes as it  subsums the "release notes" in different categories. So a use in the same API is IMHO not possible (without losing information). Also the Changelog module is not only able to load and parse a Changelog but also to save it back and to promote an unreleased section to a new version.

matthid:
Thanks for looking into that :)

magicmonty:
Awesome, thanks

PR #1878, opened 2018-04-20
Ported Fake.Sql.DacPac to FAKE 5

josiahdj:
I did my best to follow the guidelines and instructions for contributing. If I've missed something or done something that needs correcting me please let me know. This is my first "substantial" PR to FAKE.


src/legacy/FakeLib/Sql.DacPac.fs
```diff
@@ -1,4 +1,5 @@
 ﻿/// Contains helpers around deploying databases.
+[<System.Obsolete("Please open Fake.Sql instead and use the Fake.Sql.DacPac module (FAKE0001 - package: Fake.Sql.DacPac, module: DacPac)")>]
```
matthid: This is a small problem. Because now you cannot actually fix that warning according to the migration guide (because old and new namespace are the same and we cannot include the new module into fakelib, because of naming conflict).

So we either remove this file and add only the new module (and have a breaking change because of `RequireQualifiedAccess`) or find a different solution

matthid:
I'm trying to find a solution. Thanks for the PR!

PR #1890, opened 2018-04-26
Fake 5 InnoSetup Module

kblohm:
Hi,
i created a module to support [Inno Setup](http://www.jrsoftware.org/isinfo.php), a tool to build installers for windows. 
Since there are other installer tools not yet ported i put the module under Fake.Installer.
If u want to use that namespace for an actual Fake-Installer, i can also put it under Fake.InstallerTools or something similair.

I also have a few questions:
* should new modules still be added into FakeLib.fsproj?
* if not, the build script probably needs to be modified. GenerateDocs docs only looks into the "./build" folder and as far as i can see FAKE.sln is only used to generate nuget packages. At least i suppose thats why i do not get any documentation generated :)

Thanks!


matthid:
Thanks! I think the namespace is nice. I don't expect a critical conflict even if we do some kind of installer as this will be in a separate `InnoSetup` module (or something like that)

> should new modules still be added into FakeLib.fsproj?

not necessarily, the general guideline is that all new stuff needs to be available as module. If people need stuff in the old fakelib adding it is fine for me.

> if not, the build script probably needs to be modified. GenerateDocs docs only looks into the "./build" folder and as far as i can see FAKE.sln is only used to generate nuget packages. At least i suppose thats why i do not get any documentation generated :)

I think you found a huge pain point ;).
Actually currently we only generate docs for "fakelib" as it contains all the new apis as well (for migration purposes). If we now have a new module outside of fakelib we probably need to tackle that issue ...

kblohm:
Hi,
i will try to fix the documentation issue for new modules. But that should probably be another PR.
If there is anything else i can do here, just tell me.

matthid:
Hi, yes I think this is ready (I actually already changed to rc_9 in order to release this) you just slightly missed the release window ;)

help/templates/template.cshtml
```diff
@@ -173,6 +173,12 @@
 																<li><a href="@(prefix)apidocs/fake-documentation-docfx.html">DocFx</a></li>
 															</ul>
 														</li>
+                                                        <li>
+															<a href="@(prefix)apidocs/index.html#Fake.Installer">Installer</a>
+															<ul>
+																<li><a href="@(prefix)apidocs/fake-installer-innosetup.html">InnoSetup</a></li>
```
matthid: This will afaik be a dead link, but yeah

matthid:
@kblohm Yes that is quite an unfortunate bootstrapping problem, I can tackle that if needed. Usually I just `#load` the required `.fs` files temporarily in that situation.

matthid:
Thanks!

PR #1919, opened 2018-05-09
WIP Fsc module with fix for #689

vanceism7:
This is fschelper ported to a new fake5 module. Not sure if this is the correct way to call fsc.exe. Also, when calling fsc.exe through a Process call, you lose distinction between errors and warnings that you normally get back in the form of `FSharpErrorInfo` when you call FSharpChecker's compile method. I thought of using regex to scan the error messages for the keywords "warning" or "error", but I figured i'd ask about it first to see if there's a better way to handle this. 
Also, I'm still unsure if I've referenced FSharp Compiler Service the correct way, although I suspect this setup should work.

Lastly, did we need to apply this fix to the fake4 module as well?


src/app/Fake.DotNet.Fsc/Fsc.fs
```diff
@@ -0,0 +1,449 @@
...
+///                 Debug false 
+///             ]
+let Compile (fscParams : FscParam list) (inputFiles : string list) : unit = 
+    let res = compile fscParams inputFiles
+    if res <> 0 then raise <| BuildException("Fsc: compile failed with exit code", [ string res ])
+
+
+/// An externally referenced compiler
+let private externalCompile (fscTool: string) (optsArr: string []) = 
+    let args = String.Join(" ", optsArr)
```
matthid: I think we should use `(Arguments.OfArgs optsArr).ToWindowsCommandLine` instead of `String.Join` the former will handle arguments with spaces.

src/app/Fake.DotNet.Fsc/paket.references
```diff
@@ -0,0 +1,4 @@
+group netcore
+
+FSharp.Core
```
matthid: Please add `FSharp.Compiler.Service` here.

src/app/Fake.DotNet.Fsc/Fake.DotNet.Fsc.fsproj
```diff
@@ -0,0 +1,39 @@
...
+    <Compile Include="Fsc.fs" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Fake.Core.Process\Fake.Core.Process.fsproj" />
+    <ProjectReference Include="..\Fake.IO.FileSystem\Fake.IO.FileSystem.fsproj" />
+    <ProjectReference Include="..\Fake.Core.Trace\Fake.Core.Trace.fsproj" />
+    <ProjectReference Include="..\Fake.Core.String\Fake.Core.String.fsproj" />
+    <ProjectReference Include="..\Fake.DotNet.MSBuild\Fake.DotNet.MSBuild.fsproj" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net46'">
```
matthid: Please remove this item group for `FSharp.Compiler.Service` (see comment in paket.references)

src/app/Fake.DotNet.Fsc/Fsc.fs
```diff
@@ -0,0 +1,449 @@
...
+/// ## Sample
+///
+///     ["file1.fs"; "file2.fs"]
+///     |> Compile [Out "" 
+///                 Target Exe
+///                 Platform AnyCpu
+///                 References []
+///                 Debug false 
+///             ]
+let Compile (fscParams : FscParam list) (inputFiles : string list) : unit =
```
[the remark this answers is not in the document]
vanceism7: `Compile` and `compile` were both functions from the fake4's FscHelper. Initially I was thinking I needed to keep them the same so it wouldn't introduce a breaking change but I just realized this module doesn't exist in fake5 yet so it doesn't matter haha. I'll just remove one of them I think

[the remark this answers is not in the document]
matthid: Yes we have indeed the opportunity to improve the API and should use it :)

src/app/Fake.DotNet.Fsc/Fsc.fs
vanceism7: Hey I'm getting a weird name conflict error when trying to use Arguments. In vscode, everything is fine, but when running build.fsx, and in visual studio, the compiler thinks I'm trying to use the discriminated union case `DocoptResult.Arguments` instead of the one defined in `Fake.Core.Process`.
  
The error im getting specifically is
>  error FS0072: Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may
allow the lookup to be resolved.

Both of these Argument types exist in Fake.Core so I'm not really sure how I can inform the compiler of which one it needs to use. Any ideas on how to fix this?

vanceism7: Nevermind, figured it out. The file with Arguments defined in Fake.Core.Process was not included in the fakelib.fsproj config

matthid: I just fixed the same in release/rc_12 branch

vanceism7:
@matthid - If this seems ok to you then I think this pull request is good to go

src/app/Fake.DotNet.Fsc/Fsc.fs
```diff
@@ -0,0 +1,480 @@
...
+/// ## Sample
+///
+///     ["file1.fs"; "file2.fs"]
+///     |> CompileWithResult [Out "" 
+///                 Target Exe
+///                 Platform AnyCpu
+///                 References []
+///                 Debug false 
+///             ]
+let CompileWithResult (fscParams : FscParam list) (inputFiles : string list) : int =
```
matthid: Can you please use lower-cased first letters for public APIs (these are our current api guidelines), see also `Compile`, `CompileExternalWithResult`, `CompileExternal`

help/templates/template.cshtml
```diff
@@ -121,6 +121,7 @@
 															<ul>
 																<li><a href="/dotnet-cli.html">Cli</a></li>
 																<li><a href="/dotnet-assemblyinfo.html">AssemblyInfoFile</a></li>
+                                                                <li><a href="/apidocs/v5/fake-dotnet-fsc.html">MSBuild</a></li>
```
matthid: You might want to change the text here to `Fsc` :)

src/app/Fake.DotNet.Fsc/Fsc.fs
vanceism7: Oh gotcha, ok, sure thing!

help/templates/template.cshtml
vanceism7: Doh! My bad haha, I had a feeling I was forgetting something. Thanks

vanceism7:
Oh yea, one last issue I remembered: When I added FCS to paket references and tried to restore, I was getting an error saying something about System.Runtime is only compatible with .net 4.6.2. So I changed the target from net46 to net462. Also I don't believe it was compatible with netstandard1.6 either. It seems a little out of alignment with the rest of the modules so not sure if there's something wrong with that or not

matthid:
Thanks a lot for taking care of this!

matthid:
@vanceism7 https://github.com/fsharp/FAKE/pull/1920/commits/41f9142e1f36dc0aa3b0289f0ae51c9deb6d825e This is why you had problems with FSharp.Compiler.Service. Basically we had to update the dependencies in order to be able to compile for net462

vanceism7:
@matthid Oh gotcha, I wonder if it will work with net46 though? When I used fcs through nuget, it was working for net46 if I remember correctly.

Btw, is it safe for me to delete this branch now?

matthid:
@vanceism7 Because net46 was already in place.

> Btw, is it safe for me to delete this branch now?

Yes

PR #1923, opened 2018-05-11
Update documentation

kblohm:
Hi,
* i updated FSFormatting and added warnings in all the templates to show obsolete modules, types etc.
* i shortened the obsolete message so that it fits on one line (looks better)
* i changed all of the blockquotes to proper warnings/info blocks
* i added prism.js to format non fsharp code (mostly bash, but also some json and xml). I did not manage to use it to format the console output for example [here](https://fake.build/fake-commandline.html). But i am pretty sure that is due to a bug in FSFormatting. I am not able to get a `<pre>` without the markdown parser converting some of the elements. As far as i know markdown should not touch stuff in html-blocks so i guess that is a bug.


help/markdown/contributing.md
```diff
@@ -31,15 +31,23 @@ It turns `*.md` (Markdown with embedded code snippets) and `*.fsx` files (F# scr
 
 * Add a new git remote in order to retrieve upstream changes.
 
-        git remote add upstream https://github.com/fsharp/FAKE.git
+<pre>
+    <code class="lang-bash">
```
matthid: This is because 
````` 
```bash
`````

is not working, correct?

File: help/markdown/core-process.md [hunk 2]
```diff
@@ -1,6 +1,9 @@
 # Starting processes in "FAKE - F# Make"
 
-**Note:  This documentation is for FAKE 5! **
+<div class="alert alert-info">
```
matthid: Shouldn't we update the css for `**`? (Sorry I don't have any idea what's the best way)

help/templates/template.cshtml
```diff
@@ -97,7 +100,7 @@
 															<a href="/core.html">Core</a>
 															<ul>
 																<li><a href="/core-targets.html">Targets</a></li>
-																<li><a href="/apidocs/v5/fake-core-environment.html.html">Environment</a></li>
+																<li><a href="/apidocs/v5/fake-core-environment.html">Environment</a></li>
```
matthid: good eyes :)

help/markdown/contributing.md
matthid: Ah okey that is due to the bug you reported in fsf

kblohm: Yes. I have to add those by hand, otherwise i get somethink like `<pre class="fsnip"><code lang="bash">` and prism.js does not like those.
The ones for the commandline output are even worse. I did not manage to get them working at all so i had to leave them as is.

help/markdown/core-process.md
kblohm: I use both alert info and alert warning. Restyling bold everywhere is probably not a great idea. DocFx markdown has an extension to use notes and warnings. I could add that to FSFormatting in theory i guess.

help/templates/template.cshtml
kblohm: I had to click on every link to see if i missed something ;)

kblohm:
Contributing Page:
Contributing Page

API-Docs:
API-Docs

API-Docs2:
APID-Docs2

kblohm:
One think that is also kinda nice with prism.js: i could just add one more javascript file and get a copy-code button.

matthid:
Looks very nice! Good job.
One thing I noticed in the screenshots (Contribution Page) Is that the nesting of the code-blocks is no longer correct. AFAIK they should be part of the bullet points and are no longer. I'm not sure if we can fix that with the markdown/html combination. It's not a show-stopper.

kblohm:
They are inside the `<li>`. They were not [before](https://fake.build/contributing.html) (atleast one of the two) :p. The whole block just looks bigger because of the header. I can remove the header and place somethink on the left.

kblohm:
Ok i fixed the ones i could find. I will add the clipboard stuff tomorrow.
Have a good night :)

matthid:
Ok thanks a lot! I'm trying to get this out...

PR #1930, opened 2018-05-15
Better deprecated message.

JohBa:
I think the message for this obsolete could be missleading, as it could mean you should not open an issue.


matthid:
Thanks, welcome on board!

PR #1932, opened 2018-05-16
WIP: global dotnet cli-tool

kblohm:
Hi,
since dotnet core 2.1 is in RC now, i created a global FAKE cli tool, based on the 'old' cli-tool.
I build fake itself with the tool, so it should be working fine (at least on my windows machine).
There are still some things we need to decide/do:
* As far as i know a global tool can not be installed as a `DotNetCliToolReference`. So should there be 2 tools now? I guess the `DotNetCliToolReference` makes it easier for CI-Server so thats probably good. So i vote for 2 tools :)
* The name of the project. I named it Fake-Cli for now. Could also be dotnet-fake-global or fake.cli or whatever. The name of the project does not have to influence the name of the command, which is just `fake`.
* I did not update the docs yet. 
* Your bootstrapping-repo should probably be updated afterwards.


build.fsx
```diff
@@ -224,7 +224,16 @@ Trace.setBuildNumber nugetVersion
 //if current |> Seq.contains CoreTracing.defaultConsoleTraceListener |> not then
 //    CoreTracing.setTraceListeners (CoreTracing.defaultConsoleTraceListener :: current)
 
-let dotnetSdk = lazy DotNet.install DotNet.Release_2_1_4
+
+let dotnetSdk = lazy DotNet.install (fun p ->
```
matthid: Should we add that version to `Fake.DotNet.Cli` as well?

matthid:
yes yes yes. Thanks for looking into this :)
Yes finding a reasonable name is the hardest thing here:
 - `fake-bin`
 - `fake-cli`
 - `fake-global-tool`
 - `dotnet-fake-global`
 - ... or with `.` instead of `-`?

Honestly I don't really know (which is probably the reason why I have not looked into this yet)

build.fsx
kblohm: I would have, but i was not sure about the build version. We could just add '2.1.300-rc1' without the build version and just update that 'silently' if it changes. But i would not pollute the module with several 2.1.300-rc1-versions.

kblohm:
just 'fake' would be nice ;).
Could also be fake-runner.
Shouldn't travis use the version that was just downloaded?

matthid:
Seems like slack likes `fake-cli`, which is actually quite nice

kblohm:
fake-cli <- all lowercase?

matthid:
I would write it all lower case, but I don't think casing actually matters when used on the command line later on.

matthid:
> Shouldn't travis use the version that was just downloaded?

Yes it does, but it seems like environment variables are set such that the new version uses the old sdk-files....

matthid:
I think we need something similar to what we have in the msbuild helper:

https://github.com/fsharp/FAKE/blob/d099facbe19ed1676e66c74fef295baa43a73c66/src/app/Fake.DotNet.MSBuild/MSBuild.fs#L231-L233

kblohm:
Probably not, but its also consistent with dotnet-fake. I will change it to lower.

Exactly the same lines are in the DotNet module. Thats probably a bug.

matthid:
You can still force-push, I just changed to base because this was on the wish list of rc13 :)

kblohm:
At least for me, PATH just points to the root dotnet folder and not a specific version. So i guess its the travis.yml that sets the dotnet environment variable to 2.1.4.

kblohm:
Do u prefer updating the travis.yml or removing 'dotnet' from the environment in addition to MSBuild stuff (I have MSBUILD and MSBUILD_PATH on my machine and none of the ones listed. So they are missing aswell maybe?)

matthid:
Does that mean you have found the root cause? I'm not sure I understand. I feel like we should find the root cause but chaning travis.yml is a workaround if we can't find it.

kblohm:
So build.fsx changes the workingDirectory to the directory of the just downloaded sdk. It does not change the dotnet executable though. That one is still the one that DotNet.exec finds. As in home/.dotnet i suppose. Do u know if the installer script both installs to home/.dotnet/sdk/2.1.300-rc1-008673 AND updates the executable that is in home/.dotnet/ ? Because it does not look like it does that on linux.

kblohm:
Ok, i have changed the casing and i added the RC to the DotNet module.
I also changed the travis version. But just to see if that even matters. I will look into a proper fix later.

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -10,30 +10,30 @@ module DotNet =
 
     // NOTE: The #if can be removed once we have a working release with the "new" API
     // Currently we #load this file in build.fsx
-    #if NO_DOTNETCORE_BOOTSTRAP
+    // #if NO_DOTNETCORE_BOOTSTRAP
```
matthid: Maybe just remove. I think you are correct and we don't need this anymore.

matthid:
Do we need https://github.com/fsprojects/Paket/pull/3208 or how did you make it work?

paket.dependencies
```diff
@@ -95,7 +95,7 @@ group Build
     nuget SourceLink.Fake
     nuget ILRepack
     nuget RoslynTools.ReferenceAssemblies
-
+    nuget Newtonsoft.Json
```
matthid: Where does this come from?

kblohm: I referenced the #load DotNet module directly so i can use DotNet.Release_2_1_300_RC1. If i dont add the Newtonsoft.Json the Bootstrap test fails. I can remove it and just add the new version directly and update that later.

kblohm:
Did u read the reason why travis build is probably using the wrong version? Do u have any opinions on how to fix that? I could 
* set the right path in the build script
* i could add a new method to the DotNet module that can be used to try and find a specific version of Dotnet
* i could add a new property to the Dotnet options that specifies the prefered version. Since the current toolpath is set in the `Options.Create()` i would have to search for the tool twice if the property is set, so that might be kinda strange.

kblohm:
It did use the right version, but probably only because i changed the travis.yml. I wanted to see if that helped.
The build script still just sets the working directory and not the actual dotnet executable to the downloaded version.
So i dont think that is a bug. I am pretty sure Fake.Dotnet.Cli just uses the version it finds, and that is probably the one travis puts there. But i do not really know what happens exactly in travis when u set a dotnet version in travis.yml.

matthid:
> So i dont think that is a bug. I am pretty sure Fake.Dotnet.Cli just uses the version it finds

Maybe we talk through each other but if you use `DotNet.install` according to the docs:

https://fake.build/dotnet-cli.html

Ie combine the `DotNet.install` and `DotNet.<exec>` calls then it uses exactly the installed binary (this is ensured by the function returned by `DotNet.install` and afaik that works). 

You can see it in the logs `/usr/bin/dotnet` is the one installed by travis. In the previous log it started `/home/<something>/dotnet`

kblohm:
I read the docs, now i know what you mean, sorry:)
I added some logging and reverted the travis.yml change. Maybe that helps to see whats happening.

Another thing:
Do you also sometimes have the problem that you need to manually update your git index to get the assemblyinfo to update or is that just me? If not, maybe we should add something to the docs. The first time i had the issue it took quite some time as i did not know that the fake script is skipping them.

matthid:
> Do you also sometimes have the problem that you need to manually update your git index to get the assemblyinfo to update or is that just me? If not, maybe we should add something to the docs. The first time i had the issue it took quite some time as i did not know that the fake script is skipping them.

Yes I know that bit is quite a nasty "feature" the only idea is to never commit assemblyinfos to prevent merge conflicts between branches. I only have this issue when aborting forcefully the fake process, because usually the fake script should restore the state. If you have a scenario where it doesn't properly restore the state please let me know :)

matthid:
Now you have run into 

https://github.com/fsharp/FAKE/blob/d099facbe19ed1676e66c74fef295baa43a73c66/src/app/Fake.DotNet.Cli/DotNet.fs#L602-L611

Welcome to my world ;)
The "best" way to fix it is to use `Args.ToWindowsCommandLine` but then you might run into problems with `#load` because of the introduced dependency. Feel free to hack around as needed :)

kblohm:
> Do we need fsprojects/Paket#3208 or how did you make it work?

I just followed a tutorial. I thought it might not work because the tutorial [mentioned](https://natemcmaster.com/blog/2018/05/12/dotnet-global-tools/#publish-output-into-pack) that some stuff might not be copied so i just tried it out. I am currently using the global tool so i think it is working fine. Do you know what could go wrong?

matthid:
Basically this looks like a bug anyway, or I had some clever thought when implementing this I forgot already. I'd assume that the `NO_DOTNETCORE_BOOTSTRAP` is defined in the project and previously it had to work with mono (as at the time the script was still run with "old" fake to bootstrap. However I'm not sure why I did not just switch to `"` for all platforms, but I guess that doesn't work either for some random reason

matthid:
At least now you know why I hesitated to start doing this :P

matthid:
Also we probably need to report the dotnet-cli bug you found when updating travis.yml, because otherwise they will probably break us with the next release :/

kblohm:
Args.ToWindowsCommandLine <- for the download on linux? Or was that a typo?

matthid:
I think yes basically netcore uses windows escaping rules and internally converts to platform specific stuff, which is a difference to mono. On the other hand differences between the full framework and mono are usually considered a bug in mono. Therefore ToWindowsCommandLine should work everywhere. But honestly reality is a bitch and I wouldn't really wonder if  this breaks on mono. But as long as nobody reports it / uses fake 5 it should be fine ;)

matthid:
You can ignore the VSTS CI, it's something I'm experimenting with.

kblohm:
Not sure if i would call that progress, but at least i am back to breaking because the dotnet core version is not picked up and not because it cant parse arguments. :p.
On the arguments part: there are like 5 different approaches to do process arguments in FAKE. Maybe we should look into unifying them somehow? I am using the Process.xxxParam functions now. The string-module is very inconsistent.

matthid:
Ok we can see in https://travis-ci.org/fsharp/FAKE/builds/381013452#L7037
That /usr/bin/dotnet is 2.1.4, which makes sense as it is the installation from travis

We also see that https://travis-ci.org/fsharp/FAKE/builds/381013452#L7085-L7087 installed in /home/travis/.dotnet/dotnet the correct version and that we start the correct version

Now we need to figure out why the new version runs with old msbuild SDK files. I think the issue might be similar to https://github.com/fsharp/FAKE/issues/1776 where some environment variables are used to point the new dotnet-installation to the msbuild-sdk files from the 2.1.4 installation.
This is because on travis the complete build is run by using `dotnet fake` (which internally sets various variables).
Next step would be to just output all variables, for example by just adding

```fsharp
seq { for d in System.Environment.GetEnvironmentVariables() do let de = d :?> System.Collections.DictionaryEntry in yield (de.Key.ToString(), de.Value.ToString()) }
    |> Seq.sortBy fst
    |> Seq.iter (fun (key, value) -> printfn "'%O' -> '%O'" key value)
```

and letting it run on travis.

kblohm:
At least on windows, there is an sdk folder inside my dotnet folder. Shouldnt the installer return the dotnet executable from the dotnet/sdk folder it just installed instead of the one in dotnet/ ?  Or is this different on linux? Same goes for checking if the requested version is already installed. Just checking the main dotnet folder might get the wrong version request an older one for example.

matthid:
> Shouldnt the installer return the dotnet executable from the dotnet/sdk folder it just installed instead of the one in dotnet/ ?

It does. Problem is that in msbuild you can change so much behavior  via environment variables. And on travis we run within `dotnet fake`, because of `dotnet build build.proj /nologo /verbosity:minimal` so we already run WITHIN an msbuild context when running the new dotnet-cli. Basically default  behavior when running new processes (on any platform) is that it will inherit all environment variables. I'd assume (from previous issue) that we need to filter one of the inherited environment variables before calling our new dotnet-cli

>  Same goes for checking if the requested version is already installed. Just checking the main dotnet folder might get the wrong version when requesting an older one for example.

Question is: Where else should we search but in `PATH`? I'm not sure it is so easy

matthid:
In short: try to add the lines I send above at the start of `build.fsx` and let's compare the build outputs. If we really want to do an comparison we could add another travis build which executes `fake <>` directly without `dotnet build` as wrapper, similar to 

https://github.com/fsharp/FAKE/blob/d099facbe19ed1676e66c74fef295baa43a73c66/.circleci/config.yml#L14-L30

Because I'm assuming that this would work.

matthid:
Ok circle-ci added an option recently to build pull requests, so next push should have an circleci build (to ease it a bit)

matthid:
> On the arguments part: there are like 5 different approaches to do process arguments in FAKE. Maybe we should look into unifying them somehow? I am using the Process.xxxParam functions now. The string-module is very inconsistent.

Yes I agree, basically all but one are broken for some kind of input. We probably should all mark as obsolete.

matthid:
Added an issue https://github.com/fsharp/FAKE/issues/1940, will look at it eventually

kblohm:
I am not using Args.toWindowsCommandLine right now. At least for travis the Process.quote is working fine.

matthid:
Ok I think we probably should filter
```
  'MSBuildLoadMicrosoftTargetsReadOnly' -> 'true'
  'MSBuildSDKsPath' -> '/usr/share/dotnet/sdk/2.1.4/Sdks'
```

in addition to `MSBuildExtensionsPath` and `MSBUILD_EXE_PATH`. I'm pretty sure it will work then :)
Can you add that to the `msbuild` and `dotnet-cli` modules?

matthid:
And there is  ` 'DOTNET_HOST_PATH' -> '/usr/share/dotnet/dotnet'` which at least looks suspicious

matthid:
Google says it might be a good idea to unset that as well (at least when running the dotnet-cli)

kblohm:
Google never lies! I added them.

kblohm:
Should we remove all of the environment vars by default or just in the installer case? Somebody might set those intentionally.

matthid:
Actually it should be fine, I designed the setup in a way that if you set those manually it will "just work"

matthid:
Not sure why circleci was failing, retried the build.

matthid:
I think the current circleci error goes away after merge/rebase.

kblohm:
Ahh you are too fast ;). Should i let these run or rebase again?

kblohm:
The "UnskipAndRevertAssemblyInfo" target. Could that be a final target instead? This will not solve all the problems, as i sometimes just skip a build if a realise i forgot something, but at least for the cases where it just fails it should be in a valid state afterwards.

matthid:
Yes good idea. maybe we could detect ctrl+c and run final targets anyway no idea...

kblohm:
It would probably be a good idea if you could test the global tool on your machine. I am still kinda sceptical because of 
> Do we need fsprojects/Paket#3208 or how did you make it work?

As i said, its working for me, but there must be a reason why he needed to make the changes to paket.

matthid:
OK lets try this out on our new shiny staging instance :)

matthid:
Our new tool

kblohm:
Nice :). Ping me if something is not working.

matthid:
Just for reference:  The failure we saw on travis was a bug in travis, because it didn't actually install all dotnet sdk requirements. I reported it here: https://github.com/travis-ci/travis-ci/issues/9644

matthid:
It's released :)

PR #1953, opened 2018-05-21
List targets quietly

baronfel:
This PR fixes a build break I got while building the repo, and silences the output of the warnings if the `RunContext` verbosity is set to Silent.

### DIFF
--- a/src/app/Fake.Runtime/FakeRuntime.fs
+++ b/src/app/Fake.Runtime/FakeRuntime.fs
@@ -611,7 +611,8 @@ let prepareAndRunScript (config:FakeConfig) : RunResult =
 let retrieveHints (config:FakeConfig) (runResult:Runners.RunResult) =
     match runResult with
     | Runners.RunResult.SuccessRun _ -> []
-    | Runners.RunResult.CompilationError err -> [
+    | Runners.RunResult.CompilationError err ->
+      [
         // Add some hints about the error, for example
         // detect https://github.com/fsharp/FAKE/issues/1783
         let containsNotDefined = err.Errors |> Seq.exists (fun er -> er.ErrorNumber = 39)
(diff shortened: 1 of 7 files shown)


baronfel:
Result after this change:
```
bash-3.2$ dotnet ~/oss/FAKE/src/app/Fake.netcore/bin/Debug/netcoreapp2.0/fake.dll -s build --list
The following targets are available:
   AssemblyInfo
   KillStarted
   Clean
   YarnInstall
   DotnetRestore
   DotnetBuild
   DotnetTest
   WatchUnitTests
   BrunchProduction
   DotnetPublish
   BuildWebsiteDebug
   BuildCollectorDebug
   BuildBrunch
   WatchWebsite
   RunWebsite
   CheckGitlab
   ReleaseGitlab
   DebianPackage
   CopyDeployScripts
   CopyDeployScriptsDebug
   CreateUpdatePackage
   CreateUpdatePackageDebug
   UploadToUpdateServer
   Debug
   Release
   JavascriptBuild
   Build
   RunTests
   Version
   Xml
   BuildProduction
```
which any completion script could easily skip the first line of.
I'd looked at disabling that first line of output as well, but the RunContext isn't something that Targets have access to from what I can tell?

baronfel:
Added sorting of the target names, but it's harder to test that because my script is using proper nugets.

src/app/Fake.Runtime/FakeRuntime.fs
```diff
@@ -611,7 +611,8 @@ let prepareAndRunScript (config:FakeConfig) : RunResult =
 let retrieveHints (config:FakeConfig) (runResult:Runners.RunResult) =
     match runResult with
     | Runners.RunResult.SuccessRun _ -> []
-    | Runners.RunResult.CompilationError err -> [
```
matthid: Actually that syntax will no longer yield a warning with next F# version ;)

src/app/Fake.netcore/Program.fs
```diff
@@ -156,7 +156,7 @@ let runOrBuild (args : RunArguments) =
       let result =
         match runResult with
         | Runners.RunResult.SuccessRun warnings ->
-          if warnings <> "" then
+          if warnings <> "" && args.VerboseLevel.PrintNormal then
             traceFAKE "%O" warnings
```
matthid: Actually shouldn't this print into standard-error anyway? So your tool can just ignore stderr?

baronfel: that's a good point, let me try redirection and see what I get

baronfel: Yes you're right of course :D

baronfel: we can roll back this change then, I think.

matthid: Yes, this basically means it already works are you are not blocked, correct (besides the sort)?

baronfel:
we may not even need the sorting, at least for zsh, because it seems to have sorted it itself:
image

src/app/Fake.netcore/Program.fs
matthid: Just remove that line change, please

matthid:
Thanks! I think travis is just choking right now as all builds seem to be red right now.

PR #1973, opened 2018-05-31
Update CI script to latest .NET Core

teo-tsirpanis:


matthid:
@teo-tsirpanis We have to differentiate between `dotnet-fake` (https://docs.microsoft.com/en-us/dotnet/core/tools/extensibility#per-project-based-extensibility) and `fake-cli` (https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools)
- `dotnet-fake` which is used by travis (and other CIs) will still work
- `fake-cli` will probably no longer work as it was build with `2.1.300-rc` and not with the stable release. However we don't use it in any CI yet.

teo-tsirpanis:
Why not build it with the stable .NET Core 2.1 then?

Sent from Mail for Windows 10

matthid:
https://github.com/fsharp/FAKE/pull/1972 :)

teo-tsirpanis:
I see...

matthid:
Problem seems that they don't have the version on travis yet

PR #1978, opened 2018-06-04
FakeVar enhancements

BlythMeister:
Added:

- Get with type
- Get or default
- Get or fail

Also added tests to the above


BlythMeister:
Have also updated use of Expecto.Flip as this is quite confusing and subtle when used as a file level "open" statement

src/app/Fake.Core.Context/Context.fs
```diff
@@ -112,10 +112,20 @@ let forceFakeContext () =
...
   | RuntimeContext.Fake e -> e
 
-let getFakeVar name =
+let getFakeVar<'a> name =
   forceFakeContext()
   |> getFakeContext name
   |> Option.map (fun o -> o :?> 'a)
+
+let getFakeVarOrFail<'a> name =
+  match getFakeVar<'a> name with
```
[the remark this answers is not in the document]
BlythMeister: ohhh really :s

i wonder if it's possible to remove all the variable stuff?

matthid:
Looking at this, we probably should have basic integration-tests to test if we are still compatible with the old runner

BlythMeister:
Agreed...not sure how to implement...

src/app/Fake.Core.Context/Context.fs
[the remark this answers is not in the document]
matthid: > i wonder if it's possible to remove all the variable stuff?

We maybe could move it somewhere else but have to leave an obsolete attribute for now (we can do the breaking change sometime after the release)

matthid:
> not sure how to implement...

The tests? Basically like the current integration tests, but using the "installed" version if fake and not the built one (for example we could use `dotnet fake`)

src/app/Fake.Core.Context/Context.fs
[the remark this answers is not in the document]
BlythMeister: Exactly what I've started working on 😂
All internal usage I'll migrate away from obsolete.

BlythMeister:
@matthid done the refactoring and also fixed a few other things which were build warnings to make it easier to see the wood for the trees when building locally

src/app/Fake.Core.Variables/Variables.fs
```diff
@@ -6,7 +6,11 @@ open Fake.Core.Context
 let get<'a> name =
   forceFakeContext()
   |> getFakeContext name
-  |> Option.map (fun o -> o :?> 'a)
+  |> Option.map (fun o -> try
+                            o :?> 'a
+                          with e -> 
+                            failwithf "Variable '%s' - %s" name e.Message
```
matthid: please use `raise <| exn(sprintf "", e)` instead (at the very least)

src/app/Fake.Core.Variables/Variables.fs
```diff
@@ -21,29 +25,35 @@ let getOrDefault<'a> name defaultValue =
 let remove name =
   forceFakeContext()
   |> removeFakeContext name
-  |> Option.map (fun o -> o :?> 'a)
+  |> ignore
```
matthid: Sometimes you might want to know which value has been removed? Anyway I'm not sure if it is worth the potential breaking change ;) (On the other hand this one isn't too bad - I don't think anyone uses this directly at the moment.

BlythMeister: I was working on this being brand new, so not breaking change.

matthid: Ah sorry, I'm probably a bit over cautious on this thing :)
Problem is this is the "connection" between runtime and script and breaking stuff there is a huge pain to cleanup :P

src/app/Fake.DotNet.AssemblyInfoFile/AssemblyInfoFile.fs
```diff
@@ -213,7 +213,6 @@ module AssemblyInfoFile =
 
     /// Creates a C# AssemblyInfo file with the given attributes and configuration.
     /// The generated AssemblyInfo file contains an AssemblyVersionInformation class which can be used to retrieve the current version no. from inside of an assembly.
-    [<System.Obsolete "Please use 'create' instead">]
```
matthid: You don't like create anymore or what is the rationale?

src/app/Fake.Core.Variables/Variables.fs
BlythMeister: I completely agree, I was unaware of the link originally

src/app/Fake.DotNet.AssemblyInfoFile/AssemblyInfoFile.fs
BlythMeister: I thought I was a big agressive on the obsolete.  What if someone wanted to control themselves?
Since we don't plan to delete, only make private i see no harm in leaving without obsolete

matthid:
Generally I really like the direction of this. Good idea to make this a thing and simplify the context module.

> Have also updated use of Expecto.Flip as this is quite confusing and subtle when used as a file level "open" statement

I always thought that's how to use it :/. I personally don't really like the default signature, but whatever

src/app/Fake.DotNet.AssemblyInfoFile/AssemblyInfoFile.fs
matthid: > Since we don't plan to delete

;)

src/app/Fake.Core.Variables/Variables.fs
```diff
@@ -0,0 +1,59 @@
+﻿/// This module contains helpers for managing build time variables
+module Fake.Core.Variables
```
matthid: Should we add `[<RequireQualifiedAccess>]`?

BlythMeister: Agreed

BlythMeister:
Re expecting, I agree it seems wrong, but personnally spent an hour confused as the code didn't match the docs 😂

matthid:
Seems like AppVeyor died :/

BlythMeister:
I noticed it vanished... I didn't do it. 😂
Last commits made from phone as I'm away from laptop until the morning... I hope they compile!

src/app/Fake.Core.Variables/Variables.fs
```diff
@@ -0,0 +1,60 @@
+﻿/// This module contains helpers for managing build time variables
+[<RequireQualifiedAccess>]
+module Fake.Core.Variables
```
matthid: I just fixed the compilation error and I'm not sure about the plural form. But `Variable.set` doesn't really sound better. Maybe explicitly call it `FakeVar` or `FakeContext` for example? Any idea?

BlythMeister: I like FakeVar

BlythMeister: Though that means there is a function called fakeVar so would be FakeVar.fakeVar

How about ContextVariable ?

matthid: Yes but `FakeVar.fakeVar` ;) 
maybe `FakeVar.create` instead of that and
`FakeContext.set` for the raw stuff (or we make that internal)

BlythMeister: Oooo FakeVar.create love it!

BlythMeister:
Shame it has build error.
Will pick it up tomorrow 🤦🏼‍♂️

BlythMeister:
image

src/app/Fake.Core.Tasks/Fake.Core.Tasks.fsproj
```diff
@@ -21,9 +21,7 @@
     <ProjectReference Include="..\Fake.Core.Process\Fake.Core.Process.fsproj" />
     <ProjectReference Include="..\Fake.Core.String\Fake.Core.String.fsproj" />
     <ProjectReference Include="..\Fake.IO.FileSystem\Fake.IO.FileSystem.fsproj" />
-    <ProjectReference Include="..\Fake.Core.Context\Fake.Core.Context.fsproj">
-      <FromP2P>true</FromP2P>
-    </ProjectReference>
+    <ProjectReference Include="..\Fake.Core.Context\Fake.Core.Context.fsproj" FromP2P="true" />
```
matthid: What does P2P mean here? I hope this has no effect on dependencies.

src/app/Fake.Core.Trace/Fake.Core.Trace.fsproj
```diff
@@ -19,6 +19,7 @@
     <ProjectReference Include="..\Fake.Core.Environment\Fake.Core.Environment.fsproj" />
     <ProjectReference Include="..\Fake.Core.String\Fake.Core.String.fsproj" />
     <ProjectReference Include="..\Fake.Core.Context\Fake.Core.Context.fsproj" />
```
matthid: We probably could remove Fake.Core.Context now (but I guess it doesn't hurt)

matthid:
Thanks a lot for going all the way with this :) Let's see how it works on staging

src/app/Fake.Core.Tasks/Fake.Core.Tasks.fsproj
BlythMeister: It was already here, i just made the projects nicer.
From what i can tell, it enables the transitive project dependencies to be pulled

BlythMeister: I actually think all of the references should have it...based on this: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference

BlythMeister: P2P = project-to-project

PR #1987, opened 2018-06-08
Add Fake.Rsync

MangelMaxime:
I started adding Fake.Rsync module.

I am creating this PR to check if this is building and if I didn't forget to do something. Taking too much time on my machine. :)


MangelMaxime:
@matthid 
Do I need to create an empty version of `AssemblyInfo.fs`

Error:
```
fatal: Unable to mark file src/app/Fake.Rsync/AssemblyInfo.fs
=Finished (Failed) 'SetAssemblyInfo' in 00:00:01.4031626
```

matthid:
Yes or copy one from another project. Not exactly sure why. I guess we first try to ignore changes with git and then create the file afterwards. I guess we need to re-order that or do it twice (and check if the file exists first)

matthid:
@MangelMaxime Don't worry and no need to be sorry. Have fun! :)

MangelMaxime:
@matthid Ok the CI is green now :)

src/app/Fake.Rsync/Rsync.fs
```diff
@@ -0,0 +1,415 @@
...
+open System
+open System.IO
+
+/// Helpers for running rsync tool
+///
+/// Under windows you will need to add it yourself to your system or use something like cygwin/babun
+///
+/// ## Sample
+///
+/// let result =
```
matthid: I believe it needs to be indented further in order to be formatted as code in the docs.

src/app/Fake.Rsync/Rsync.fs
```diff
@@ -0,0 +1,415 @@
...
+        static member WithActions actions options =
+            { options with Actions = actions }
+
+        static member WithSources sources options =
+            { options with Sources = sources }
+
+        static member WithDestination dest options =
+            { options with Destination = dest }
+
+    let rec actionToString =
```
matthid: could be private/internal?

src/app/Fake.Rsync/Rsync.fs
```diff
@@ -0,0 +1,415 @@
...
+
+    let exec (buildOptions: Options -> Options) args =
+        let results = new System.Collections.Generic.List<Fake.Core.ConsoleMessage>()
+        let timeout = TimeSpan.MaxValue
+
+        let errorF msg =
+            Trace.traceError msg
+            results.Add (ConsoleMessage.CreateError msg)
+
+        let messageF msg =
```
matthid: I always thought we have a helper collecting the messages for you already. But I guess you wanted to have "live"-output?

MangelMaxime: Yes ,I think having them live make more sense. And also, I believe I did that to get the right output color too.

MangelMaxime:
@matthid I took your comments in account.

For the `> ` synthax, this is the only way I found to have a nice display.

matthid:
Yes I might take a look at the docs on staging, lets see...

matthid:
Thanks for this new feature :)

MangelMaxime:
For the need of ‘> this is probably due to a bug in Ionide display. 

I am working on fixing it. When released I will take a look back to this module documentation. 

Releasing in current stage still make sense 😊

matthid:
Ugs no idea how I missed it, but there are seem to be some problems with this:

- It is missing from the API-Docs completely https://staging.fake.build/apidocs/v5/index.html
- It is missing in the menu: 
image
- It should be in the `Fake.Tools` (ie `Fake.Tools.Rsync`) namespace instead of `Fake,Rsync` (IIRC the first problem is a direct consequence of this because of globbing)

matthid:
You can take a look at https://staging.fake.build

MangelMaxime:
Is it because the namespace is: `namespace Fake.Rsync` compared to `module Fake.Tools.Git` ?

matthid:
No I think it is because:

- I think the API reference is generated via globbing over `Fake.Tools.*` and `Fake.Core.*` (and some others). But we have to take a look to be sure.
- The menu is actually created by hand in `help/templates/template.csproj` and I would have added it myself, but I couldn't add the entry because I couldn't find the API reference to link to ;)

MangelMaxime:
I was thinking you were looking into it :)

Also, I have no idea how the docs site of FAKE works. If you don't take a look I will try to a look at it tomorrow or after tomorrow

matthid:
Nevermind, it looks like it was easy enough :). My analysis was incorrect - probably all that was missing was adding it to the Fake.sln solution...

MangelMaxime:
Ah yes, correct I don't use the `*.sln` file in general so I forgot to add it :)

Thanks you for taking a look at it. I would have search for a long time :)

PR #1994, opened 2018-06-12
Print target.Name if target.Description is empty

zakaluka:


zakaluka:
For #1993

zakaluka:
Incorrectly fixed this on a very old fork.

PR #1996, opened 2018-06-12
Changed NoDescription to target.Name. Updated links to use HTTPS per …

zakaluka:
…GitHub changes.

For #1993


matthid:
Which output are you referring to? Because afaik we first print the target name and then on the next line the description.
With this change we now print the target name twice. 

The relevant place is:

https://github.com/fsharp/FAKE/blob/fe316d3355e93b116e9536ad261ecc292d706554/src/app/Fake.Core.Trace/TraceListener.fs#L246-L250

Maybe in your change we just use `null` which will skip printing?

matthid:
Btw: Thanks for your first contribution!

zakaluka:
After reading your response, I see that you are right!  However, at least in Ionide on Windows 7, here's what it looks like:

image

The description is quite prominent and helps break up the different targets and the visual flow, whereas the `Start target 'XYZ'` looks like any other text and is difficult to notice at a glance.

Putting a null may make more sense, but maybe possibly even better would be to emphasize the line `Starting target ...` instead of the description.

zakaluka:
The latest version has 2 changes in total:
- Return `null` if there is no description
- If there is a description, print it to the right of the target, after a colon `:`

zakaluka:
@matthid I think I've gone as far as is needed to get the output a little better.  (surprisingly small amount of change)

I'm still seeing outputs of "NoDescription" in the tests, but can't figure out where they are coming from.  That string doesn't occur anywhere in my codebase.  If you have any insight, would love to fix it.

src/app/Fake.Core.Trace/TraceListener.fs
```diff
@@ -110,7 +110,7 @@ type TagStatus =
 
 /// Defines Tracing information for TraceListeners
 [<RequireQualifiedAccess>]
-type TraceData = 
+type TraceData =
     | ImportData of typ:ImportData * path:string
     | BuildNumber of text:string
     | ImportantMessage of text:string
```
matthid: Can you add a comment to `OpenTag` that `description` might be null in case it is missing (which is probably the best we can do for now, ideally we would change to `string option` but I don't want to make that breaking change here.

src/app/Fake.Core.Trace/TraceListener.fs
```diff
@@ -243,11 +243,12 @@ type ConsoleTraceListener(importantMessagesToStdErr, colorMap, ansiColor) =
...
             | TraceData.OpenTag(KnownTags.Target _ as tag, description) ->
-                write false color true (sprintf "Starting %s '%s'" tag.Type tag.Name)
-                if not (isNull description) then
-                    let msg = TraceData.TraceMessage("", true)
-                    let color2 = colorMap msg
-                    write false color2 true description
+                let msg = TraceData.TraceMessage("", true)
+                let color2 = colorMap msg
+                write false color2 false (sprintf "Starting %s '%s'" tag.Type tag.Name)
+                if not (String.IsNullOrWhiteSpace description) then
```
matthid: why not leave `isNull`?

Note that the current solution has a small problem when using `--parallel`. The "newline` character might be printed after another target description (ie you have something like `Starting T1 T1Starting T2 T2: descriptionT1: description T2` followed by two newline characters). It's probably not a huge deal, but if we use the same color anyway we should probably create the string and use a single write with newline character here.

matthid:
Regarding the tests: I think the output might come from testing the legacy fake version, which is fine (we don't really want to update that one).

matthid:
Ok let me finish this. Again thanks!

PR #2004, opened 2018-06-20
Fix an ambiguity in Octokit.fsx

teo-tsirpanis:
A version of Octokit has introduced an overload of the constructor of `Credentials` and has broken my build as a result.

### DIFF
--- a/modules/Octokit/Octokit.fsx
+++ b/modules/Octokit/Octokit.fsx
@@ -75,7 +75,7 @@ let private retryWithArg count input asycnF =
             return captureAndReraise ex
     }
 
-let createClient user password =
+let createClient user (password: string) =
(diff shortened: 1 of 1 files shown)


matthid:
Actually we should probably make people use `Fake.Api.GitHub` instead and add some warnings to this file

teo-tsirpanis:
That’s some very good news, since [Paket’s GitHub dependency support has issues](https://github.com/fsprojects/Paket/issues/3262). And I admit that this is not the most elegant way of releasing a plugin, especially in the age of FAKE 5.

Sent from Mail for Windows 10

teo-tsirpanis:
Any hope to integrate this in the `master` branch?

PR #2044, opened 2018-07-30
Add function to add additional command-line arguments in Fake.DotNet.DotNet.Options as a sequence of strings.

### LINKED ISSUE #2042: Shouldn't Fake.DotNet.DotNet.Options.CustomParams be a list of strings?
In FAKE 4, types in `Fake.DotNetCli.###Params` had an `AdditionalArgs` member.

But now, the closest thing of a replacement is a `string option`. I think a string list would ease migration to FAKE 5.

teo-tsirpanis:
This funcion internally converts them to a single string.
It does not break anything.
The downside is that the arguments cannot be _read_ as a sequence of strings.

This PR was made to address the hastily closed #2042.

### DIFF
--- a/src/app/Fake.DotNet.Cli/DotNet.fs
+++ b/src/app/Fake.DotNet.Cli/DotNet.fs
@@ -498,6 +498,10 @@ module DotNet =
             lift (fun o -> { o with Verbosity = verb}) x
         let inline withCustomParams args x =
             lift (fun o -> { o with CustomParams = args}) x
+        /// Sets custom command-line arguments expressed as a sequence of strings.
+        /// This function overwrites and gets overwritten by `withCustomParams`.
+        let inline withAdditionalArgs args x =
+            withCustomParams (args |> Args.toWindowsCommandLine |> (function | "" -> None | x -> Some x)) x
         let inline withDotNetCliPath path x =
             lift (fun o -> { o with DotNetCliPath = path}) x
 


src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -498,6 +498,8 @@ module DotNet =
             lift (fun o -> { o with Verbosity = verb}) x
         let inline withCustomParams args x =
             lift (fun o -> { o with CustomParams = args}) x
+        let inline withAdditionalArgs args x =
+            withCustomParams (args |> Args.toLinuxShellCommandLine |> Some) x
```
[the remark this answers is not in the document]
matthid: I think you need to use toWindowsCommandLine as this is the one netcore uses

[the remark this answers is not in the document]
teo-tsirpanis: Done.

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -498,6 +498,8 @@ module DotNet =
             lift (fun o -> { o with Verbosity = verb}) x
         let inline withCustomParams args x =
             lift (fun o -> { o with CustomParams = args}) x
+        let inline withAdditionalArgs args x =
+            withCustomParams (args |> Args.toWindowsCommandLine |> Some) x
```
matthid: Should we use `None` if the list is empty?

matthid:
This is actually a nice idea. Only thing is we should probably add a comment explaining that this uses the other field (such that a user knows that using `withCustomParams` and `withAdditionalArgs` doesn't work at the same time. Or we change the naming accordingly (`withCustomParamsAsList` for example?)

matthid:
Wow you are fast

teo-tsirpanis:
I will do the second. Documentation in these files is already non-existent, and documenting just one function would stand out. 😂😂

> Wow you are fast

A meme that elaborates on your claim of my speed of response

###### This is why it took me so much. 😛

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -498,6 +498,8 @@ module DotNet =
             lift (fun o -> { o with Verbosity = verb}) x
         let inline withCustomParams args x =
             lift (fun o -> { o with CustomParams = args}) x
+        let inline withAdditionalArgsAsList args x =
```
teo-tsirpanis: It's a `seq`, but I'm pretty sure they would get the point.

matthid: What I meant was for `withX` helpers it is clear they write to `X` for the `withAdditionalArgs` helper it is not really clear what it is doing (well it is clear but doesn't match the scheme). That's why I would either opt for something with `withCustomParams*` prefix or a comment (yes it would stand out but the helper is a bit "special" so that would be fine imho) ;)

teo-tsirpanis: Done.

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -498,6 +498,11 @@ module DotNet =
             lift (fun o -> { o with Verbosity = verb}) x
         let inline withCustomParams args x =
             lift (fun o -> { o with CustomParams = args}) x
+        /// Sets custom command-line arguments expressed as a sequence of strings.
+        /// This function overwrites and gets overwritten by `withCustomParams`.
+        /// Furthermore, there is no way to get back the arguments as a sequence of strings.
```
matthid: > there is no way to get back the arguments as a sequence of strings

Technically, you can with `Args.fromWindowsCommandLine`, but fine by me ;)

matthid: > there is no way to get back the arguments as a sequence of strings

Technically, you can with `Args.fromWindowsCommandLine`, but fine by me ;)

matthid:
Merging once CI is green.

teo-tsirpanis:
I will delete the last comment line. I actually don’t think it’s needed.

Sent from Mail for Windows 10

matthid:
Why wait - I want to have this part of the release :P

PR #2127, opened 2018-10-07
Add funtion to update build status based on context

BlythMeister:
### Description

This this can be piped into from runAndGetContext

This extends #2111

I did not add the functionality into the other run functions to return context as the way the run or default works doesn't lend itself to returning TargetContext since there is functions in there which return unit.

I'm not sure if there is a nice way to get the run or default to allow the return of the context.


BlythMeister:
@matthid updated and also added support into all other run types

help/markdown/core-targets.md
```diff
@@ -126,6 +126,44 @@ Everything after the target will be interpreted as argument for the target:
...
+
+open Fake.Core.TargetOperators
+
+// *** Define Dependencies ***
+"Clean"
+    ==> "Build"
+    ==> "Deploy"
+
+// *** Start Build ***
+match Target.runOrDefault "Deploy" with
```
matthid: `runOrDefault` still returns `unit`, correct?

src/app/Fake.Core.Target/Target.fs
```diff
@@ -834,22 +840,36 @@ module Target =
                 if not allowArgs && arguments <> [] then
                     failwithf "The following arguments could not be parsed: %A\nTo forward arguments to your targets you need to use \nTarget.runOrDefaultWithArguments instead of Target.runOrDefault" arguments
                 match target with
-                | Some t -> runInternal singleTarget parallelJobs t arguments |> ignore
-                | None -> fDefault singleTarget parallelJobs arguments
+                | Some t -> Some(fun () -> runInternal singleTarget parallelJobs t arguments)
+                | None when defaultTarget = "***FAKE-ListAvailable***" -> listAvailable()
```
matthid: If we add something like this we would add some "prefix" and prevent users from using it when defining targets. Honestly though I don't think we need to do this and there should be another way to do this (as there was previously)

help/markdown/core-targets.md
BlythMeister: that;s an error, will fix

src/app/Fake.Core.Target/Target.fs
BlythMeister: Agreed.  Will work on this :)

BlythMeister:
@matthid agreed.  I think in exception cases it doesn't work because runInternal raises an exception out, so you don't get the context. :(

BlythMeister:
yeah good point.

using the standard (not with context) methods basically handle all this for you.
If you want to get the context, your gonna need to do a bit of work yourself ;)

BlythMeister:
We also already have tests covering these cases, but they were pulling the context off of the exception.

BlythMeister:
@matthid this has gone a bit mental now...

BlythMeister:
i thin i have avoided a breaking change...

BlythMeister:
image

Tested using TeamCity & got a proper error message :D

BlythMeister:
@matthid i think this is done now. - i can keep tinkering...but i don't think i'm adding much now.

Are you OK with the changes proposed?

BlythMeister:
Further screenshot of teamcity

image

Showing the failure and the reported problems.

At the moment, only TeamCity will support these extra message formats, others will just print whatever they have in the BuildState section of the listener.

BlythMeister:
@matthid this pr is a bit big and messy, would you prefer a couple of smaller changes?

src/app/Fake.BuildServer.TeamCity/TeamCity.fs
```diff
@@ -226,11 +232,11 @@ module TeamCity =
 
     /// Sends a warning message.
     let internal warning message =
-        TeamCityWriter.sendToTeamCity "##teamcity[message text='%s' status='WARNING']" message
+        TeamCityWriter.sendToTeamCity "##teamcity[message text='%s' status='WARNING']" (TeamCityWriter.encapsulateSpecialChars message)
```
matthid: I think escaping is done by sendToTeamCity already (arguably a bit unexpected....)

src/app/Fake.Core.Trace/TraceListener.fs
```diff
@@ -129,6 +129,7 @@ type TagStatus =
     | Success
     | Warning
     | Failed
+    | FailedWithMessage of message:string
```
matthid: maybe we could include the exception object.
Note that I'm a bit worried about this change... This is because the TagStatus is used on other places and I feel like adding a new state here needs more updates...

BlythMeister:
There are 4 sets of changes here.
1) TeamCity listener enhancements
2) ability to always get context, splitting the raise error
3) update build state
4) raise build error messages

I'll break this up, as some are less controversial than others :wink:

BlythMeister:
Closing for now, will replace with a few PRs.

PR #2246, opened 2019-02-01
Git lg command refactoring and fix

serjebulavsky:
### Description

- Documentation related to getAllFiles was incorrect, fixed. 
- Now there is additional command getAllModifiedFiles that returns all --modified files from current revision.

- related to #2244


serjebulavsky:
I've researched a little bit into 'ls-files' command and seems like it is not really correct. --modified also returns not just modified in working directory, but modified from current head. Later i'll create a small PR related to docs adjustment of getAllFiles.

PR #2277, opened 2019-03-15
Add support for MSBuild16

BlythMeister:
Adding support for the new MSBuild16 paths into MSBuildHelper


BlythMeister:
@forki are you still taking fixes to fake 4 or should I approach this in my script  to use msbuild 16.

PR #2363, opened 2019-08-01
Add check for attribute & make get attribute return option

BlythMeister:
### Description

Add check for attribute & make get attribute return option


BlythMeister:
Don't need this really, you can call getAttribute and check for null.

PR #2394, opened 2019-09-11
Limit msbuild escaped characters to the minimum

### LINKED ISSUE #2392: MSBuild property escaping is invalid
### Description

Essentially the fix of https://github.com/fsharp/FAKE/issues/2112 has been a too broad I think. Properties passed to MSBuild command line need `;` to be escaped as it's a list of properties separated by `;` but after https://github.com/fsharp/FAKE/commit/e7e9c101849e0c22f9bf75a36e2607467833a3ee it's also escaping other things like `\\` but MSBuild doesn't support that at all ending up with an invalid command line when a property contains a path for example.

### Repro steps + Expected & Actual behavior

Pass any path as a property to msbuild, for example to call Wix trying to generate `/p:WixTargetsPath=c:\Code\packages\build\WiX\tools\wix.targets` by you'll get `/p:WixTargetsPath=c:%5CCode%5Cpackages%5Cbuild%5CWiX%5Ctools%5Cwix.targets`

### Known workarounds

Copy paste MSBuild.fs in your project and keep only `;` being escaped in properties

### Related information 

* Version: *5.16.1*
* Snark: *MSBuild doing the logical thing here and unescaping all characters % escaped would be too logical it seem* ☹

vbfox:
It's not a perfect solution but it will already solve the biggest problem (paths)

The always-working-solution of generating `/p:X="foo bar"` would be nice but require access to the low level msvcrt escaping and of producing non escaped arguments (like https://github.com/vbfox/FoxSharp/blob/master/src/BlackFox.CommandLine/MsvcrCommandLine.fs#L65 ) and that would require some changes to the command line escaping API or ugly code.

Also it looks always working but I didn't test it on linux or via the dotnet.exe wrapper so one day maybe.

Fixes #2392

### DIFF
--- a/src/app/Fake.DotNet.MSBuild/MSBuild.fs
+++ b/src/app/Fake.DotNet.MSBuild/MSBuild.fs
@@ -513,7 +513,9 @@ module MSBuild =
         //
         // See https://github.com/fsharp/FAKE/issues/2112 and https://github.com/fsharp/FAKE/issues/2392
         // for some history on this problem
-        v.Replace(";", "%3B").Replace(",", "%2C")
+        v.Replace("%", "%25")
+         .Replace(";", "%3B")
+         .Replace(",", "%2C")
 
     let properties =
         p.Properties
--- a/src/app/Fake.DotNet.MSBuild/MSBuild.fs
+++ b/src/app/Fake.DotNet.MSBuild/MSBuild.fs
@@ -501,19 +501,19 @@ module MSBuild =
         | [] -> None
         | t -> Some("t", t |> Seq.map (String.replace "." "_") |> String.separated ";")
 
-    // see https://github.com/fsharp/FAKE/issues/2112
     let escapePropertyValue (v:string) =
-        // https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-special-characters?view=vs-2017
-        v.Replace("%", "%25")
-         .Replace("\\", "%5C")
-         .Replace("\"", "%22")
-         .Replace(";", "%3B")
-         .Replace(",", "%2C")
-         .Replace("$", "%24")
-         .Replace("@", "%40")
-         .Replace("'", "%27")
-         .Replace("?", "%3F")
-         .Replace("*", "%2A")
+        // This list has been found by trial and error but isn't perfect.
+        //
+        // Neither escaping everything (breaks <UsingTask AssemblyFile="$(Path)"" />)
+        // nor only what's documented by Microsoft at https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-special-characters?view=vs-2017
+        // (same problem with paths and ',' isn't covered but is still required to escape) works.
+        //
+        // This escape isn't perfect, mainly passing a a parameter with a ',' inside to a part of
+        // msbuild that doesn't support '%' escaping doesn't work but it's good for 99% of use cases.
+        //
+        // See https://github.com/fsharp/FAKE/issues/2112 and https://github.com/fsharp/FAKE/issues/2392
+        // for some history on this problem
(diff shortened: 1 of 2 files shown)


forki:
"I don't know MSBuild" - that's one of the reason why FAKE was born ;-)

Julien Roncaglia <notifications@github.com> schrieb am Mi., 11. Sep. 2019,
17:24:

matthid:
I warned you of the tests ;)

matthid:
@vbfox Just to let you know we even have some (undocumented) helpers to work with "old" command line parsing. This is due to a recent bug-report in SonarQube which seems to use the old-style argument parsing as well:

https://github.com/fsharp/FAKE/blob/921f1973c9f1f1b3d7ce14ef4cc4282b4e0b83ef/src/app/Fake.Core.Process/CmdLineParsing.fs#L188

To feel free to change stuff around a bit and use the raw helpers (note that you can only use them at the end, because once you use any non-raw-helper the "raw" part will just disappear)

See https://fake.build/apidocs/v5/fake-core-argumentsmodule.html

matthid:
Just one more warning. It took me quite a while to get the test green especially xplat makes this quite a pain. Given that msbuild implementations changed especially on unix/mono even if we figure it out it might not work there..

As the output is working properly maybe there is a workaround in msbuild itself, maybe something like
```
<MyProp>$(MyProp)</MyProp>
```
It could force msbuild to remove escapings?

Also I'm still unsure whether we should just report this to the msbuild people? This looks like a bug in msbuild to be honest.

matthid:
Potential workaround: `$([MSBuild]::Unescape(...))` found in https://github.com/microsoft/msbuild/issues/4086

vbfox:
> Just one more warning. It took me quite a while to get the test green especially xplat makes this quite a pain. Given that msbuild implementations changed especially on unix/mono even if we figure it out it might not work there..

Yeah I read their code and it's essentially a bunch of sadness with more sadness piled on top.

* MSBuild Full framework use it's own argument splitting AND argument parsing rules that are slightly different from the ones in msvcrt
* On coreclr  they can't get the raw command line so do only their own parsing but use the platform splitting.
* On windows it mean msvcrt reimplemented rules over the raw cmdline so that's kinda ok
* On unix it mean that the args are transformed to a string with windows like rules in https://github.com/dotnet/coreclr/blob/39f207ad8babe677029ddc2c72714a9959da2475/src/pal/src/init/pal.cpp#L1188 then they are parsed by the msvcrt like function before being passed to msbuild that does it's own custom parsing

The plan to remove one step in https://github.com/microsoft/msbuild/issues/839 but that mean that anything typed on msbuild dotnet core on unix is currently parsed/escaped by the shell, made into an array, that array is made into a string (with windows like transformation) then parsed into an array (with windows parsing rules) then parsed by msbuild.

Knowing that on unix any command line string we build will be parsed with windows like rules to be passed to `execv` doesn't make any of that simpler....

vbfox:
Following the discovery of https://github.com/Microsoft/msbuild/issues/885#issuecomment-252737626 where essentially it's ruled that `UsingTask` is buggy but considered legacy and will never be fixed, I'll continue on this PR finding a minimum set of character to escape to pass the command line and let it at that.

If only `UsingTask` and a few other legacy places are affected by some characters being impossible to put without manually modifying the msbuild file to force an escape so be it. We'll live with it.

vbfox:
Tried a few things but I don't see any way forward https://github.com/microsoft/msbuild/issues/885#issuecomment-252737626 say this bug is by design and there is no way to avoid it hat is fully cross platform and future-proof.

2 solutions for others:
* `$([MSBuild]::Unescape(...))` in the msbuild project
* Use forward slashes when possible

PR #2437, opened 2019-12-11
Change the README badges and their placement.

teo-tsirpanis:
GitLab CI and CircleCI are not maintained (the first at least is not visible).

AppVeyor's badge image link needs to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # FAKE - F# Make
 
 [![Join the chat at https://gitter.im/fsharp/FAKE](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/fsharp/FAKE?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


matthid:
Probably fair, while gitlab is still in use it is not publically visible for some reason (I just looked at it: There doesn't seem to be any setting for it. I always have to log in to see the build). CirlcleCI indeed stopped working and I never cared to much to enable it again I guess :)

teo-tsirpanis:
You have to update the link for AppVeyor's badge image; it still points to the old one, where the build is successful. Because it seemed to be private, I couldn't do it myself.

matthid:
@teo-tsirpanis Can you repeat what exactly is still missing with the AppVeyor badge? Currently it seems to be OK:

image

teo-tsirpanis:
It is the badge's _picture_. It always says "BUILD PASSING" even when it was failing in the past, because it still using the old project.

You have to get the new URL from AppVeyor's settings:

AppVeyor's settings pane

matthid:
@teo-tsirpanis Thanks for the fast response. I think I got it sorted out...

Issue #2559, opened 2020-11-23
Octo.createRelease with mulitple packages genenerates incorrect command line


michaelsmithson:
### Description

When using the Octo.createRelease with multiple packages, the windows command line generated wraps all the package arguments inside quotes rather than specifying them individually.

### Repro steps

This can be seen by tweaking the input to the unit test "Realistic Octo Args To Windows Command Line"

In this test if we add a second package:
```
            let setReleaseParams = (fun (ro:Fake.Tools.Octo.CreateReleaseOptions) ->
                { ro with ...  Packages=["MyPackage:1234567890.12.34"; "Package2:1.2.3"]; ... })
```

We should expect the following command line (with other arguments skipped clarity):

```
create-release ... --package=MyPackage:1234567890.12.34 --package=Package2:1.2.3 ...
```

But instead we get both the --package arguments included inside a set of quotes.
```
create-release ... "--package=MyPackage:1234567890.12.34 --package=Package2:1.2.3" ...
```

### Related information 

The is running on Windows using FAKE 5.20.3.

PR #2560, opened 2020-11-23
Octo.createRelease generates correct windows command line with multiple packages

michaelsmithson:
## Description

This is a fix for #2559. As noted in that issue, the Octo.createRelease command was inserting incorrect quotes around multiple package parameters

Instead of generating a string that contains multiple package arguments and passing this to the command line generate (which on windows was enclosing all these in quotes), this change appends a list of all the individual package arguments to the list of the other arguments. This stops the args.ToWindowsCommandLine function from inserting quotes around what should be seperate parameters.

This also affects the Octo.push command which I assume suffers a similar problem, and the generated command line for this looks correct, but I don't use that in my build script so haven't really run it.

### DIFF
--- a/src/app/Fake.Tools.Octo/Octo.fs
+++ b/src/app/Fake.Tools.Octo/Octo.fs
@@ -190,10 +190,8 @@ let private optionalObjParam p o =
     | None -> ""
 
 let private stringListParam p os =
-    let sb = Text.StringBuilder()
-    for o in os do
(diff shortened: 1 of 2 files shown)


raymens:
Would love to have this in there. I have fixed to one of the previous version of the octo package to have this working still (in my case it's the "push"). From what I can tell the PR is quite complete.

Can we get this merged? Or is there something that's missing?

PR #2562, opened 2020-12-23
Changing coverlet output format to receive a list of formats

### LINKED ISSUE #2563: Coverlet output format should receive a list of formats
### Description

Shoud be possible to pass multiples output formats to coverlet

See [here](https://github.com/coverlet-coverage/coverlet/issues/161)

### Repro steps

When create a `CoverletParams ` the type of `OutputFormat` accepts only one format

### Expected behavior

I shold be able to pass a list of output paramters

### Known workarounds

Set manually the OutputFormat of `DotnetTestOptions`

```fsharp
options
  |> fun p -> {
            p with MSBuildParams = {
                 p.MSBuildParams with
                Properties = p.MSBuildParams.Properties @ [("CoverletOutputFormat","cobertura,json")]
            }}
```
 
I did a PR https://github.com/fsharp/FAKE/pull/2562 for this, ~but I cant figure out why it broke~

lucasteles:
### Description

Shoud be possible to pass multiples output formats to coverlet

See [here](https://github.com/coverlet-coverage/coverlet/issues/161)

fixes #2563

### DIFF
--- a/help/markdown/dotnet-testing-coverlet.md
+++ b/help/markdown/dotnet-testing-coverlet.md
@@ -47,7 +47,7 @@ DotNet.test (fun p ->
     }
     |> Coverlet.withDotNetTestOptions (fun p ->
         { p with
-            OutputFormat = Coverlet.OutputFormat.OpenCover
+            OutputFormat = [Coverlet.OutputFormat.OpenCover]
             Output = "coverage.xml"
             Include = [
                 // Include all namespaces from assemblies whose name starts with MyProject
--- a/src/app/Fake.DotNet.Testing.Coverlet/Coverlet.fs
+++ b/src/app/Fake.DotNet.Testing.Coverlet/Coverlet.fs
@@ -27,7 +27,7 @@ type ThresholdStat =
 /// Coverlet MSBuild parameters. For more details see: https://github.com/tonerdo/coverlet/blob/master/Documentation/MSBuildIntegration.md
 type CoverletParams =
     { /// (Required) Format of the generated output.
-        OutputFormat : OutputFormat
+        OutputFormat : OutputFormat list
         /// (Required) Path to the generated output file, or directory if it ends with a `/`.
         Output : string
         /// Namespaces to include, as (AssemblyName, Namespace) pairs. Supports `*` and `?` globbing.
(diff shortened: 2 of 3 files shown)


lucasteles:
Cant figure why the build broke

yazeedobaid:
The build for this PR should be fixed now after merging #2574.

lucasteles:
@yazeedobaid it worked! thanks! can you review this?

yazeedobaid:
@lucasteles Thanks for the update!
Here my notes on the PR:
1. As mentioned in the issue for this PR, if output formats have multiple values it should enclose them in quotes and escape these quotes. Please see [Coverlet documentation ](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/MSBuildIntegration.md#coverlet-integration-with-msbuild) page. Also as I understand Coverlet in FAKE is used with `dotnet test` command, so the escaping should not affect Linux or build environments per this note in [Coverlet documentation](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/MSBuildIntegration.md#note-for-linux-users) also?
2. The API now accepts a list of output formats, so users who use the module need to update their code to make the output format argument passed as a list, even if they use one output format. I'm not sure but this introduces a breaking change to the API. What do you think?

Thanks

lucasteles:
Thanks @yazeedobaid !

1 - About the escape quotes, I thought that FAKE already do that [because other comma separated params are not escaped inside the command construction](https://github.com/fsharp/FAKE/pull/2562/files#diff-1e045ef9e52f8a01d8e6146bdc3dd0c76314f510225f89d2e91faeec1a278c7fR44), but I do not checked it in the code tho, am I wrong?

2 - Yeah, I see this as a breaking change, we could create other property which is a list of outputs types and keep the old way too, so when we have a list we prioritise it, when not, just use the old single output. But this can turn the API a bit confusing. WDYT?

yazeedobaid:
@lucasteles Thanks for the update!
For the first point did you see any need to add the escape quotes to the arguments? If it is working for you without them then I guess it is ok. For the breaking change, as you said it will make the API a bit confusing. So we can add this as a minor breaking change in the next FAKE minor release.

Also, could you please link this issue #2563 to the PR so that when it is merged the issue will be closed automatically.
Thanks

lucasteles:
@yazeedobaid  yeah, it worked

> Also, could you please link this issue #2563 to the PR so that when it is merged the issue will be closed automatically.
Thanks
Done!

yazeedobaid (commented):
Thanks

yazeedobaid (commented):
Thanks

yazeedobaid (approved):
Thanks

PR #2594, opened 2021-05-25
Update Octokit to 0.50, and pin to that version

enovales:
### Description

This updates to Octokit 0.50, and pins to that release. Previously, the `paket.dependencies` did not specify a version, which resulted in downstream breaks for things like [MiniScaffold](https://github.com/TheAngryByrd/MiniScaffold)'s release action. [Octokit 0.50](https://github.com/octokit/octokit.net/releases/tag/v0.50.0) added support for a CancellationToken in UploadAsset (https://github.com/octokit/octokit.net/pull/2267/files), which broke `Fake.Api.GitHub.uploadFile` since the interface changed.

This simply passes `Async.DefaultCancellationToken`, and pins the desired version of Octokit to 0.50 so that breaking interface changes like this aren't propagated to dependees. I updated the `paket.lock` file manually, rather than updating via `dotnet paket update Octokit`.

I ran `dotnet fake build` successfully with these changes, but it's not clear to me how I should be testing this further, as there don't seem to be integration tests available for this.

## TODO


yazeedobaid:
@enovales can you please pull from `release/next` branch and fix the conflicts. We did big changes to dependencies and the Octokit dependency was updated.

yazeedobaid:
Thanks for contributing to FAKE, 
we had to do the fix to proceed with the release. I have taken the changes in this PR and added them to `next/release`. Please see the [release notes](https://github.com/fsprojects/FAKE/releases/tag/5.21.0-alpha003), last line.

I'll close this, thanks again

PR #2631, opened 2022-01-07
Add support for 64 bit VS2022 path and version structure.

### LINKED ISSUE #2629: SqlPackage doesn't work with VS2022
### Description

OOTB SqlPackage can't be located with VS2022. This is because VS no longer resides in Program Files (x86) but Program Files.

### Known workarounds

Manually set the path yourself.

I'll raise a PR for this.

isaacabraham:
### Description
Fixes #2629 - now supports VS2022's installation of SqlPackage and gets the version using reflection against the executable itself.

## TODO

- [x] boy scout rule: "leave the code behind in a better state than you found it" (fix warnings, obsolete members or code-style in the places you worked in).

I've tested this locally - my machine is a clean build that only has VS2022 on it, hence why I noticed this issue in the first place.


yazeedobaid:
@isaacabraham no need to add the next version in `RELEASE_NOTES.md` file, it has been added for next release as `5.21.0-alpha005`. The version needs to follow the SemVer notation, this is why the build fails.

Thanks

isaacabraham:
@yazeedobaid Fixed, rerunning the build.

isaacabraham:
@yazeedobaid can you re-run? Looks like the build failed because of a timeout.

yazeedobaid:
@isaacabraham I have re-run the build again and now it fails due to this error:
```F#
Fsc: /home/runner/work/FAKE/FAKE/src/app/Fake.Sql.SqlPackage/Sql.SqlPackage.fs(201,22): error FS0001: All branches of an 'if' expression must return values of the same type as the first branch, which here is 'seq<string>'. This branch returns a value of type ''a list'.
```

Thanks

yazeedobaid:
It seems I had introduced a conflict on this pull request when I merged the refactor of `SqlPackage` pull request https://github.com/fsprojects/FAKE/pull/2634. Can you please fix the conflicts

Thanks

isaacabraham:
Hmm. I've merged and solved the conflict - that's not the issue, because it builds locally fine.

I think the problem is that it's compiling against F#6 here but the build agent is still going against F#5: Despite having a global.json of .NET 6 preview (BTW, I think you should upgrade to a non-alpha version of .NET 6 e.g. 6.0.101), it's still using [MSBuild 16](https://github.com/fsprojects/FAKE/runs/4838751329?check_suite_focus=true#step:10:282).

Maybe that's a red herring but it seems suspicious, and I can reproduce the exact compiler error by reverting back to building against .NET 5.

yazeedobaid:
@isaacabraham thanks for the update.
I have updated FAKE to latest .NET6 SDK version: 6.0.101 (https://github.com/fsprojects/FAKE/pull/2635)
As you pointed out the MSBuild version after the update is 17 now.
Could you please update your branch from release/next

Thanks

isaacabraham:
Done

yazeedobaid:
Thanks

PR #2634, opened 2022-01-14
Refactor of SqlPackage code.

isaacabraham:
### Description
This refactors some of the SqlPackage code both for readability as well as for correctness in terms of using pattern matching and (importantly) safely unwrapping options rather than checks of `IsSome` and then `.Value`, which caused a bug with Timeout previously.


yazeedobaid:
Thanks for the update on the `SqlPackage` module.

PR #2640, opened 2022-01-26
Fix FAKE template integration tests.

yazeedobaid:
### Description

Fix FAKE template integration tests. The tests started to fail after the upgrade of FAKE repository to NET 6.0.101.
The template tests were pinned to an old version of FAKE runner and from the logs, it seems that it is not compatible with NET 6

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "fake-cli": {
-      "version": "5.21.0-alpha004",
+      "version": "5.21.0",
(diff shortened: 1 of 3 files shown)


yazeedobaid:
@matthid Is there a reason why the template integration tests were pinned to version `5.16.2` ?
The template integration tests started to fail after the upgrade to Net 6.0.101, I have removed the version option in the tests and now they work again. But just want to make sure that this will not cause issues in the future that are not caught currently.

Thanks

PR #2655, opened 2022-02-07
Add known paths for more recent versions of msbuild

alcated:
### Description

Since Visual Studio 2017, two 2 major versions were released.
Here is the update to add the known paths for thoses two new versions.

## TODO


.config/dotnet-tools.json
```diff
@@ -15,10 +15,10 @@
       ]
     },
     "fake-cli": {
-      "version": "5.21.0",
+      "version": "1.0.0-1",
```
yazeedobaid, line 18: Can you please revert this change also? It seems this is an automatic commit
we will try to fix it and remove it.

alcated: I rebased onto release/next.
Is it acceptable to you ?

yazeedobaid: yes thanks

NogginBops:
This will actually not find the Visual Studio 2022 files as they are in "Program Files" but "Program Files (x86)" is the only folder searched.

Issue #2666, opened 2022-03-28
Trying to make sense of fake-cli syntax


mlaily:
Hello,

I'm new to FAKE, and I'm having a hard time trying to make sense of the FAKE CLI.  
After actually spending a few hours on this, I decided to open an issue here so maybe this can be improved. (And I can get answers :))

I don't find the `--help` very helpful on its own, maybe even convoluted to the point of being misleading, and https://fake.build/fake-commandline.html does not really answer my questions.

Here are a few things I would really appreciate to be explained better:

The help web page states that `run` and `build` are "basically equivalent", but the commands I'm trying to execute sometime succeed with one but fail with the other.  
e.g. `dotnet fake build --list` lists my targets, but `dotnet fake run --list` fails with a useless error message.  
- What is the actual difference between the two?
- Is one actually just calling the other with predefined arguments?

The `Usage:` section of the `--help` indicates that both `run` or `build` are basically identical, with everything following them being optional, but as per my previous example, it seems that's actually not true! (at least some of the time)

- Why is the ` -- SCRIPT ARGUMENTS SECTION --` showing usage with `fake-run`?
- Is `fake-run` an actual command? Something used in place of `dotnet fake run`?

I know that `--` is somewhat standard in many CLI tools, but neither the `--help` nor the docs explain what it does and when it is supposed to be used. My understanding is that it's to disambiguate that something is supposed to be a `scriptargs`, but can be omitted when the `scriptargs` does not conflict with a `fake-cli` existing option or command.  
- Am I correct in my understanding of `--`?

Maybe some basic examples of everyday commands would be useful in the `--help`?  
(instead of the current example of `fake -v build -- --list` — I can feel that it's trying to demonstrate an important point, but I'm not completely sure I get it, so it's actually more scary than useful...)

Thank you for your time.

matthid:
Yes the UX is not optimal and it is entirely my fault (I couldn't come up with something better). Rather than going full out technical let me try to motivate why it was implemented this way (we can decide to change that here and start a discussion).
Some relevant discussions (just if someone is interested in the archives):
- https://github.com/fsprojects/FAKE/issues/2151 (in particular [this comment](https://github.com/fsprojects/FAKE/issues/2151#issuecomment-429547166))
- https://github.com/fsprojects/FAKE/issues/2282 Where we already tried to improve the help string
- https://github.com/fsprojects/FAKE/issues/2125
[...]

PR #2667, opened 2022-04-19
Try to clarify CLI --help

mlaily:
### Description

Try to fix most of my concerns with the CLI --help (see #2666)

Here is the result of `fake --help` after this PR:
```
FAKE 5 - F# Make (5.22.0)


Usage:
  fake [fake_opts] run [run_opts] [<script.fsx>] [--] [<scriptargs>...]
  fake [fake_opts] build [build_opts] [--] [<scriptargs>...]
  fake --version
  fake --help | -h

Note `fake build` is basically equivalent to calling `fake run` with a script named `build.fsx`.

Fake options [fake_opts]:
  -v, --verbose [*]        Verbose (can be used multiple times)
                           Is ignored if -s is used.
                           * -v: Log verbose but only for FAKE
                           * -vv: Log verbose for Paket as well
  -s, --silent             Be silent.
                           Use this option if you need to pipe your output into another tool or need some additional processing.

Fake Run options [run_opts]:
  -d, --debug              Debug the script.
  -n, --nocache            Disable Fake cache for this run.
  -p, --partial-restore
                           Only restore the required group instead of a full restore,
                           can be set globally by setting the environment variable FAKE_PARTIAL_RESTORE to true.
  --fsiargs <fsiargs> [*]  Arguments passed to the F# interactive.

Fake Build options [build_opts]:
  -d, --debug              Debug the script.
  -n, --nocache            Disable Fake cache for this run.
  -p, --partial-restore
                           Only restore the required group instead of a full restore,
                           can be set globally by setting the environment variable FAKE_PARTIAL_RESTORE to true.
  --fsiargs <fsiargs> [*]  Arguments passed to the F# interactive.
  -f, --script <script.fsx>
                           The script to execute (defaults to `build.fsx`).


 ----- SCRIPT ARGUMENTS SECTION -----

Remaining arguments following the previously defined options are provided to the script.

Each script might understand different arguments,
but since there are good chances you are using the 'Fake.Core.Target' package,
its command-line is documented below.

THIS SECTION ONLY APPLIES IF YOU USE THE 'Fake.Core.Target' PACKAGE IN YOUR SCRIPT!

You can use the following arguments in place of `<scriptargs>`:

(`fake-run` refers to the Fake command and arguments defined above)

Usage:
  fake-run --list
  fake-run --write-info <file>
  fake-run --version
  fake-run --help | -h
  fake-run [target_opts] [target <target>] [--] [<targetargs>...]

Target Module options [target_opts]:
    -t, --target <target>    Run the given target (ignored if a target is already provided with '[target <target>]')
    -e, --environment-variable <keyval> [*]
                             Set an environment variable. Use 'key=val'.
                             Consider using regular arguments, see https://fake.build/core-targets.html
    -s, --single-target      Run only the specified target.
    -p, --parallel <num>     Run parallel with the given number of tasks.

 ----- END OF SCRIPT ARGUMENTS SECTION -----

Warning:

Ordering of arguments does MATTER.
`fake -v run script.fsx` executes `script.fsx` in verbose mode.
`fake run -v script.fsx` will try to execute a script named '-v' and fail.

If a script argument/option conflicts with any of the options allowed before,
you need to separate script options with `--`.
The reverse is also true: to print all targets, you can use
`fake build --list` instead of `fake build -- --list`
because `--list` doesn't conflict with any of the [build_opts].

Basic examples:

Specify script file and execute default script action:
    fake run mybuildscript.fsx

Specify script file and run the Clean target:
    fake run build.fsx --target Clean
or shorter version using the default build.fsx script:
    fake build -t Clean
```

### DIFF
--- a/help/markdown/core-commandlineparsing.md
+++ b/help/markdown/core-commandlineparsing.md
@@ -36,16 +36,18 @@ match DocoptResult.tryGetArgument "-m" results with
 | Some arg -> printfn "%s" arg
 ```
 
+*Note the following links are permalinks to old commits. They are intended only as examples. You are encouraged to look at the current version too, because it's likely to have been updated.*
+
 A more sophisticated example can be found in the fake runner: https://github.com/fsharp/FAKE/blob/64d871f5065412fe7b233025e454ccf3b89e46d7/src/app/Fake.netcore/Program.fs#L204-L259
 
 Or the target module:
 
-- https://github.com/fsharp/FAKE/blob/rc_1/src/app/Fake.Core.Target/Target.fs#L9-L26
+- https://github.com/fsharp/FAKE/blob/64d871f5065412fe7b233025e454ccf3b89e46d7/src/app/Fake.Core.Target/Target.fs#L9-L26
 - https://github.com/fsharp/FAKE/blob/64d871f5065412fe7b233025e454ccf3b89e46d7/src/app/Fake.Core.Target/Target.fs#L564-L619
 
 You can also take a look at the test-suite:
 
-- https://github.com/fsharp/FAKE/blob/rc_1/src/test/Fake.Core.CommandLine.UnitTests/Fake.Core.CommandLine.fs
+- https://github.com/fsharp/FAKE/blob/64d871f5065412fe7b233025e454ccf3b89e46d7/src/test/Fake.Core.CommandLine.UnitTests/Fake.Core.CommandLine.fs
 
 ## Differences to the python reference Docopt implementation
 
--- a/src/app/Fake.netcore/Cli.fs
+++ b/src/app/Fake.netcore/Cli.fs
@@ -6,8 +6,8 @@ open System
 open Fake.Core.CommandLineParsing
 
 let fakeArgsHint =
-  """
-FAKE 5
+  $"""
+{Fake.Runtime.Environment.fakeVersionStr}
 
 """
 
--- a/src/app/Fake.netcore/Cli.fs
+++ b/src/app/Fake.netcore/Cli.fs
@@ -7,62 +7,63 @@ open Fake.Core.CommandLineParsing
 
 let fakeArgsHint =
   """
-General:
-
-  The Fake command line is divided into runtime and script arguments.
-  Runtime arguments control compilation and processing of the script,
-  while script arguments are specific for the script or provided by
-  a NuGet package.
-  In most use cases you use the "Fake.Core.Target"-Package and therefore
-  inherit the corresponding command line interface. While these arguments
-  are not strictly part of the runtime we still show both below to
-  make it easier for newcomers.
-
- -- RUNTIME ARGUMENTS SECTION --
+FAKE 5
 
 """
 
 let fakeUsage =
   """
 Usage:
-  fake.exe [fake_opts] run [run_opts] [<script.fsx>] [--] [<scriptargs>...]
-  fake.exe [fake_opts] build [build_opts] [--] [<scriptargs>...]
-  fake.exe --version
-  fake.exe --help | -h
+  fake [fake_opts] run [run_opts] [<script.fsx>] [--] [<scriptargs>...]
+  fake [fake_opts] build [build_opts] [--] [<scriptargs>...]
+  fake --version
+  fake --help | -h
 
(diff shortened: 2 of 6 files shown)


src/app/Fake.netcore/Cli.fs
```diff
@@ -7,84 +7,100 @@ open Fake.Core.CommandLineParsing
...
-  Runtime arguments control compilation and processing of the script,
-  while script arguments are specific for the script or provided by
-  a NuGet package.
-  In most use cases you use the "Fake.Core.Target"-Package and therefore
-  inherit the corresponding command line interface. While these arguments
-  are not strictly part of the runtime we still show both below to
-  make it easier for newcomers.
-
- -- RUNTIME ARGUMENTS SECTION --
+FAKE 5
```
yazeedobaid, line 10: Thanks for the clean-up and organizing the help message!
I would like to suggest that we automate this and get the version of the current FAKE-CLI and print it instead of
hard-coding it as `FAKE 5`

yazeedobaid:
Also, the failing test in the build has been fixed now, could you please rebase your branch from `release/next`?

mlaily:
Hello, thank you for your feedback.

I used the existing dynamic version string (though FAKE 5 is still hardcoded in that part^^') as you suggested, and did the rebase.

I updated this PR's description with how the new `--help` looks with these changes.

yazeedobaid:
Hi thanks for the PR
Could you please rebase your branch from `release/next`? Now that the build has been fixed.

mlaily:
Hi! Done!

help/markdown/core-commandlineparsing.md
```diff
@@ -40,12 +40,12 @@ A more sophisticated example can be found in the fake runner: https://github.com
 
 Or the target module:
 
-- https://github.com/fsharp/FAKE/blob/rc_1/src/app/Fake.Core.Target/Target.fs#L9-L26
+- https://github.com/fsharp/FAKE/blob/64d871f5065412fe7b233025e454ccf3b89e46d7/src/app/Fake.Core.Target/Target.fs#L9-L26
```
yazeedobaid, line 43: These links point to a specific commit. Can we have a permanent link to the latest commit on the master branch maybe?

mlaily: That was on purpose, to avoid unintended 404s or wrong highlighted lines if the code is updated.  

On second thought, I'm not entirely sure what's preferable between a correct link to a potentially not up-to-date sample, or a might-be-broken link to an up-to-date sample...

Do you still think a link to master is better?

yazeedobaid: Ok, so as I saw, the purpose of the links is to provide the user with advanced examples of FAKE command line parsing. Why note we provide extract these examples and inline them in this MD file instead of linking it to. What do u think?

mlaily: Indeed, that would work for the example yes.

And I guess I can point the test suite link to master, as it's less a problem if the lines change.

mlaily: @yazeedobaid Hey, I was about to do what we said, but it doesn't feel right:
- The examples are quite big, as they are meant to be "more sophisticated" than the basic example inlined, and they are actually split in multiple parts. I'm not 100% sure what's important and what's not. I'd rather highlight the main part but keep the rest too as-is, but I think it would be too long to be inlined.
- There is already an inlined example to demonstrate basic usage of the parser, so adding another big one feels redundant.
- I think it's better to have a link to some "real" code on github where we can be reasonably sure that it worked at some point in time, in contrast to an artificial inlined example extracted from the sources that no one compiled before.
- For the same reason, I think it's better to have a permalink to some commit rather than point to master. I checked on master, and this code has moved quite a lot since the permalink, and it's hard to find the originally highlighted part. Since this is only for example purposes, I think it's better to have something that's not completely up to date but is easily understandable.

Sorry for the turnaround.  
I'll at least add a note indicating that the links are probably old and that it would be a good idea to check the current version too.

I'll rebase my branch too, since it's behind by a few commits. I hope it's ok with you.

yazeedobaid (approved):
Thanks for the PR

mlaily:
Thanks!

PR #2668, opened 2022-04-26
`5.23.0-alpha001` release

yazeedobaid:
`5.23.0-alpha001` release

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 5.23.0-alpha001 - 2022-04-26
+* ENHANCEMENT: Porting CMake module to FAKE 5, thanks @ziadadeela - https://github.com/fsprojects/FAKE/pull/2658
+* BUGFIX: Resolve symlinks while searching for dotnet, thanks @Smaug123 - https://github.com/fsprojects/FAKE/pull/2662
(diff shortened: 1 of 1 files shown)


mclark1129:
@yazeedobaid I'm unable to see the specific failure in the logs, not sure if it's a UI issue.  Not sure what would have changed between my branch and this one, but I can try to help figure it out!  I imagine its likely some change to stdout since the call to `dotnet --version` outputs the result, and some tests have fairly strict assertions around stdout content.

mclark1129:
Ok looks like `Error while retrieving StartTime of started process: System.ComponentModel.Win32Exception (3): Unable to retrieve the specified information about the process or thread.  It may have exited or may be privileged.`  Not clear if this is new due to the call to `dotnet --version`.  I'm going to try and run the checks against the current release as well to see.

mclark1129:
@yazeedobaid I still can't reproduce this on my fork, I should have the latest `release/next` so it seems like there may be something flaxy when running the build in Github.  Possibly some environmental difference between agents? :shrug:

https://github.com/mclark1129/FAKE/pull/3/checks

nojaf:
@yazeedobaid can we please just have an alpha of this?
I'm hitting this problem like 5 times a day and somebody finally fixed it.

PR #2681, opened 2022-06-07
Fix getLastTag

### LINKED ISSUE #2677: getLastTag should have --tags as describe argument
### Description

`Fake.Tools.Git.Information.getLastTag` should have --tags as describe argument.

### Repro steps

Run: `Fake.Tools.Git.Information.getLastTag ()` on a repository with tags.

### Expected behavior

Should return the last tag.

### Actual behavior

```
Script reported an error:
-> git describe failed: fatal: No annotated tags can describe '6e36165e8fa304de19d3743153696c517b16a888'.
   However, there were unannotated tags: try --tags.
```

### Known workarounds

`Fake.Tools.Git.CommandHelper.runGitCommand "" "describe --tags --abbrev=0"`

### Related information 

* Version of FAKE: 5.X

akhansari:
### Description

- fixes #2677

### DIFF
--- a/src/app/Fake.Tools.Git/Information.fs
+++ b/src/app/Fake.Tools.Git/Information.fs
@@ -1,4 +1,4 @@
-﻿/// Contains helper functions which can be used to retrieve status information from git.
+/// Contains helper functions which can be used to retrieve status information from git.
 module Fake.Tools.Git.Information
 
 open Fake.Tools.Git.CommandHelper
@@ -88,7 +88,10 @@ let shortlog repositoryDir =
     msg |> Seq.head
 
 /// Gets the last git tag of the current repository by calling git describe
-let getLastTag() = (describe "").Split('-') |> Seq.head
+let getLastTag () =
+    let _,msg,error = runGitCommand "" "describe --tags --abbrev=0"
+    if error <> "" then failwithf "git describe --tags failed: %s" error
+    msg |> Seq.head
 
 /// Gets the current hash of the current repository
 let getCurrentHash() =


yazeedobaid:
Hi thanks for the PR
Could you please rebase your branch from `release/next`? Now that the build has been fixed.

akhansari:
Hi @yazeedobaid, it should be ok now.

yazeedobaid:
Thanks for the fix

akhansari:
Maybe it could be some integration tests running inside a container.

Each test creates an unique temp folder and initializes git.
Then it arranges with some commits, for instance.
And finally it runs the test logic.

Wdyt?

yazeedobaid:
Yes, an integration test, the steps you mentioned will be sufficient to cover the logic. 
But one question, why u need a container for that? Why not initialize a temporary folders and repositories, run the tests and then delete them. This way is somehow similar to what a test against a database is like. You initialize the database before each test and after each test turn it down.
Can u give more info about the container approach?

akhansari:
Because usually CI runners don't have all necessary infrastructure with different setup or versions (here it's git and dotnet) to be able to run integration tests and to keep runners clean.
So [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) are used for these kind of cases.

But I'm not very familiar with github (I know gitlab better). Maybe what I'm saying is too much regarding our case.

yazeedobaid:
To my knowledge, service containers are used to hook up with another service instead of managing it in the CI pipeline. So for example running tests against multiple databases. Instead of manually installing the different database services and managing them, you can use service containers to do that for you.

We can use them for tests, but in my opinion, they will be overkill in our case. And if we went with this approach the service container will be used to test GIT-related code. And for other parts of the FAKE repo, we maybe need to set up another container.

Or maybe we can use a general container and test against it? I don't have much information/details about the best way to use service containers, but if we can use them in a way that will make it easy to develop/test/use/maintain then I'm ok with it.
What do you think?

akhansari:
You're right, when I read the documentation, what you said makes sense and service containers are not the right way. 💭

Issue #2683, opened 2022-07-12
ReleaseNotes.load does not parse SemVer correctly v2


Freymaurer:
### Description

Issue #2557 is still not fixed. I could propably fix it and do a PR, but i need to know if this is wanted, as i see multiple stale PRs.

@forki

yazeedobaid:
@Freymaurer yes, please. That would be great

Freymaurer:
Is there already a planned date when the next FAKE release will happen?

yazeedobaid:
I'm planning for another alpha release today or tomorrow.

PR #2686, opened 2022-07-18
Fix ReleaseNotes SemVer

### LINKED ISSUE #2557: ReleaseNotes.load does not parse SemVer correctly
### Description

I used ReleaseNotes.load to read in my markdown release notes and tried to add the github commit to the SemVer [metadata](https://semver.org/lang/de/#spec-item-10). As "#" is not a allowed symbol in metadata the version looks like this `0.1.1+09da94a`. 

```fsharp
let release = ReleaseNotes.load "RELEASE_NOTES.md"
```

`release.SemVer.BuildMetaData` returns in this case `09`
`release.SemVer.Original` returns in this case `0.1.1+09`

if i change the SemVer to `0.1.1+test` it returns nothing for meta data.

### Repro steps

1. Create RELEASE_NOTES.md with 

```markdown
### 0.1.1+09da94a (Released 2020-11-18)
* TestDes.:
    * TestSubDesc
```

2. Parse the markdown file with `ReleaseNotes.load "RELEASE_NOTES.md"`

3. Check SemVer.BuildMetaData/SemVer.Original of the ReleaseNotes record type.

### Expected behavior

As far as i understand the [specification](https://semver.org/lang/de/#spec-item-10) it should allow both numbers and letters for 
metadata.

> Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. 

> Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, **1.0.0-beta+exp.sha.5114f85**, 1.0.0+21AF26D3—-117B344092BD.

### Actual behavior

It stops parsing the metadata at the first letter.

Freymaurer:
### Description

- fixes #2557, #2683
- Added unit tests :white_check_mark:

@aggieben, @yazeedobaid

Had to make SemVer parsing a bit more complicated, but should still be clearly understandable. I also added comments to explain certain choices.


src/app/Fake.Core.ReleaseNotes/ReleaseNotes.fs
```diff
@@ -66,16 +66,40 @@ type ReleaseNotes =
     static member New(assemblyVersion,nugetVersion,notes) = ReleaseNotes.New(assemblyVersion,nugetVersion,None,notes)
 
 let private parseVersions =
-    let nugetRegex = String.getRegEx @"([0-9]+.)+[0-9]+(-[a-zA-Z]+\d*)?(.[0-9]+)?"
+    // https://github.com/fsprojects/FAKE/issues/2557
+    let nugetRegexLegacy = String.getRegEx @"([0-9]+.)+[0-9]+(-[a-zA-Z]+\d*)?(.[0-9]+)?"
+    let nugetRegex = 
+        /// From Fake.Core.SemVer
+        let pattern =
```
yazeedobaid, line 73: Can u please re-use patterns from `Fake.Core.SemVer` instead of duplicating them here?

src/app/Fake.Core.ReleaseNotes/ReleaseNotes.fs
```diff
@@ -66,16 +66,40 @@ type ReleaseNotes =
...
     let assemblyVersionRegex = String.getRegEx @"([0-9]+.)+[0-9]+"
     fun line ->
         let assemblyVersion = assemblyVersionRegex.Match line
         if not assemblyVersion.Success
         then failwithf "Unable to parse valid Assembly version from release notes (%s)." line
 
-        let nugetVersion = nugetRegex.Match line
-        if not nugetVersion.Success
-        then failwithf "Unable to parse valid NuGet version from release notes (%s)." line
+        let nugetVersion =
```
yazeedobaid, line 86: Can we try to decrease the nesting level in this code block? Maybe try to replace if..else with `Option` API or match expression? Or do u have any other suggestions?

src/app/Fake.Core.ReleaseNotes/ReleaseNotes.fs
Freymaurer: Can i pull `pattern` out of `(|SemVer|_|)` to use it? Like such?

```fsharp
let pattern =
      @"^(?<major>\d+)" +
      @"(\.(?<minor>\d+))?" +
      @"(\.(?<patch>\d+))?" +
      @"(\-(?<pre>[0-9A-Za-z\-\.]+))?" +
      @"(\+(?<build>[0-9A-Za-z\-\.]+))?$"

let (|SemVer|_|) version =
        match version with
        | ParseRegex pattern [major; minor; patch; pre; build] ->
            Some [major; minor; patch; pre; build]
        | _ ->
            None
```

yazeedobaid (commented):
Review changes

src/app/Fake.Core.ReleaseNotes/ReleaseNotes.fs
yazeedobaid: yeah sure

src/app/Fake.Core.ReleaseNotes/ReleaseNotes.fs
Freymaurer: This should decrease nesting, but evaluates `nugetRegexLegacy.Match line` even if not necessary.

```fsharp
  let nugetVersion = 
      /// Must split by whitespace to try match start of line and end of line in SemVer regex pattern
      let nugetVersion = 
          line.Split(' ') 
          |> Array.tryPick (fun segment -> 
            let m = segment.Trim() |> nugetRegex.Match
            if m.Success then Some m else None
          )
      // Add support for "nugetRegexLegacy" after change to correct SemVer parsing.
      // This should lead to the least disruption to users.
      let nugetVersionLegacy = nugetRegexLegacy.Match line
      // if isSome then it must be Success, so no need to check for that
      match nugetVersion, nugetVersionLegacy.Success with
      | Some nugetVersionValue, _ -> nugetVersionValue
      | None, true -> nugetVersionLegacy
      | None, false ->
          failwithf "Unable to parse valid Nuget version from release notes (%s)." line
```

yazeedobaid: I agree, but this is more readable than if..else version. So I think readability wins here.
Also, can u convert the legacy value to an option? So that in match expression u will be matching on option values and not booleans and options.

Thanks

Freymaurer: Yeah matching on option and bool was bugging me too. Cleaned it up, will push after running tests.

Freymaurer:
Done @yazeedobaid 👍

yazeedobaid:
Thanks a lot.
One last request, can u please remove the changes in `Fake.sln` file from PR to make the PR focused on issue at hand?

Freymaurer:
Done @yazeedobaid 👍 Hope it worked out.

yazeedobaid (approved):
Thanks a lot

aggieben:
🎉 Thanks @Freymaurer !

PR #2698, opened 2022-09-07
Fake.Tools.Rsync: fix quiet flag typo

SimenLK:
### Description

Using the rsync package I noticed it threw an error when trying to use rsync with quiet. Saw that there was a typo in the flag


yazeedobaid:
Thanks for sending the PR!

PR #2699, opened 2022-09-09
Enhanced search path for Visual Studio which could be installed under…

plainionist:
… "Program Files" instead of "Program Files (x86)" with VS 2022

### Description

When installing VS 2022 Professional it is no longer located under "Program Files (x86)" but "Program Files".
Therefore the search paths for msbuild.exe had to be adapted.


yazeedobaid:
Thanks for the PR

PR #2700, opened 2022-09-09
On Unix like OSs, look for sqlpackage in directories specified in $PATH

nickgravgaard:
This PR addresses issue https://github.com/fsprojects/FAKE/issues/2701 of sqlpackage not existing in the same place on MacOS and Linux distros by following the canonical Unix practice of searching the directories specified by the `PATH` environment variable instead of using a hard-coded path.

### DIFF
--- a/src/app/Fake.Sql.SqlPackage/Sql.SqlPackage.fs
+++ b/src/app/Fake.Sql.SqlPackage/Sql.SqlPackage.fs
@@ -49,7 +49,7 @@ module SqlPackage =
     let internal validPaths =
         let paths = [
             if Environment.isUnix then
-                Environment.pathDirectories |> Seq.map (fun dir -> !!(dir </> "sqlpackage")) |> Seq.concat |> Seq.map (fun path -> path, 15)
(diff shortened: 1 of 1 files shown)


mattgallagher92:
I've tested this out in my Ubuntu environment, where I have a symlink in `/usr/local/bin` to my `sqlpackage` install location, by running the expression Nick has added in F# interactive. It resolves correctly:

```fsharp
    Environment.pathDirectories
    |> Seq.map (fun dir -> !!(dir </> "sqlpackage"))
    |> Seq.concat
    |> Seq.map (fun path -> path, 15)

val it: seq<string * int> = seq [("/usr/local/bin/sqlpackage", 15)]
```

yazeedobaid:
Thanks for the PR,
I would like to suggest keeping both ways. The old hard-coded way and the commonplace (best practice) as you mentioned. From what I see, now it will search in environment variables only. I'm thinking about a case when a user doesn't have SQL Package in the environment variables but is added to the directory that was hard-coded.

yazeedobaid:
@mattgallagher92 thanks for testing and confirming the fix

nickgravgaard:
Thanks @yazeedobaid, that sounds reasonable - I've just pushed up a new commit. First it checks `$PATH` as that's where users can define their own locations for binaries, falling back to `/usr/local/bin` (normal user binaries not managed by the distro's package manager, eg. locally compiled packages), and finally `/usr/bin` (distro-managed normal user binaries).

PR #2733, opened 2023-02-22
Use RollForward project property

amis92:
### Description

Use project property documented here:
https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#rollforward

The rollForwardOnNoCandidateFx in runtimeconfig.template.json is the old way. Since .NET Core 3.0, it was merged into RollForward: https://github.com/dotnet/runtime/issues/43492

rollForwardOnNoCandidateFx is no longer documented on microsoft docs. Controlling behavior suggests RollForward project property: https://learn.microsoft.com/en-us/dotnet/core/versions/selection#control-roll-forward-behavior

The mapping from rollForwardOnNoCandidateFx=2 to Major is based on: https://github.com/dotnet/runtime/issues/43492
(and the design document linked in last comment there).

### DIFF
--- a/src/app/fake-cli/fake-cli.fsproj
+++ b/src/app/fake-cli/fake-cli.fsproj
@@ -4,12 +4,13 @@
         <OutputType>Exe</OutputType>
         <ToolCommandName>fake</ToolCommandName>
         <TargetFramework>net6.0</TargetFramework>
+        <RollForward>Major</RollForward>
         <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
         <!-- Debian packaging -->
         <Prefix>/opt/fake</Prefix>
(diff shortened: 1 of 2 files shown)


yazeedobaid:
@amis92 thanks for the PR. 
Can you please describe the effect of this change and what it addresses?

amis92:
The effect of this change is just using modern and documented solution, but the behavior stays exactly the same, as the current value of `rollForwardOnNoCandidateFx=2` is currently mapped to the RollForward=Major by the SDK anyway.

So this PR addresses tech debt, by removing usage of a legacy, undocumented solution, and replacing it with a current, documented one.

yazeedobaid:
Thanks

Issue #2735, opened 2023-02-26
dotnet local tool fails when run via FAKE, but succeeds when run via command line


mattgallagher92:
### Description

I'm seeing different behaviour when running a dotnet local tool ([grate](https://erikbra.github.io/grate/)) via FAKE than when running it directly. In particular, it fails when run via FAKE, but works without error when run directly from the command line.

### Repro steps

Details in https://github.com/mattgallagher92/fake-dotnet-local-tool-bug, but the gist is that the following command doesn't work when run via FAKE:

``` sh
dotnet grate
    --noninteractive
    --databasetype postgresql
    --connectionstring 'Host=localhost;Database=my_database;Username=postgres;Password=aVeryStrongPassword!'
    --adminconnectionstring 'Host=localhost;Database=postgres;Username=postgres;Password=aVeryStrongPassword!'
```

### Expected behavior

The command to work when executed via FAKE.

### Actual behavior

The command fails when run via FAKE, but succeeds when run directly from the command line.

### Known workarounds

Do not use FAKE for this command.

### Related information 

* Ubuntu 22.04
* dotnet 6.0.406
* Severity: annoying
* FAKE 5.23.1

yazeedobaid:
Thanks for reporting, can you please changing the command to be like follows:

```diff
Target.create "ApplyLocalDatabaseMigrations" (fun _ ->
    let grateCommand =
        "grate"
        + " --noninteractive"
        + " --databasetype postgresql"
-       + $" --connectionstring '%s{localDbConnectionString}'"
-       + $" --adminconnectionstring '%s{localDbAdminConnectionString}'"
+       + $" --connectionstring {localDbConnectionString}"
+       + $" --adminconnectionstring {localDbAdminConnectionString}"
    // This works locally but not via FAKE
    run Tools.dotnet grateCommand dbSrcPath
)
```

_Removing the quotations around the connection string._ 

I have executed the command directly on the terminal with quotations and it gave me the same error you reported. So seems quotations are not acceptable around connection strings for `grate` tool.

Also, on the re-produce repository, you mentioned in the prerequisites section that you need .NET 6 SDK to run FAKE. However, you are running the build script using `dotnet run` as a project, not an interactive script through FAKE CLI `dotnet fake build`. In that case, you need .NET 6 since the runner uses SDK assemblies to compile the script.

mattgallagher92:
Thanks for your response Yazeed 🙂 I've clarified why .NET 6 is needed in the repo.

[...]

PR #2739, opened 2023-03-10
[MSBuild] Fixed trailing garbage

0x53A:
fixes https://github.com/fsprojects/FAKE/issues/2738

---------

The previous use of ``appendQuotedIfNotNull`` doesn't make any sense, it passes ``Some`` as a function as the first parameter, which then get's formatted through ``%A`` into a string like ``<fun:quoteString@686>``. The intention was probably ``(Some str)``, but this again doesn't work.

image

Just wrapping it in quotes is probably not strictly correct, we would also need to escape existing quotes inside the string, but it quickly fixes the code without a regression.

### DIFF
--- a/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
+++ b/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
@@ -5,7 +5,6 @@ open System.Text
 open Fake.Core
 open Fake.IO
 open Fake.Net
-open Fake.Core
 
 /// [omit]
 module internal AppVeyorInternal =
@@ -54,18 +53,8 @@ module internal AppVeyorInternal =
                 // because otherwise there might be recursive failure...
                 eprintfn "AppVeyor 'AddMessage' failed: %O" e
 
-    /// Adds quotes around the string
-    /// [omit]
-    let private quote (str: string) = "\"" + str.Replace("\"", "\\\"") + "\""
-
-    /// Adds quotes around the string if needed
-    /// [omit]
-    let private quoteIfNeeded str =
-        if String.isNullOrEmpty str then ""
-        elif str.Contains " " then quote str
-        else str
-
-    let private quoteString str = quoteIfNeeded str
+    /// quote and escape the single argument, if required.
+    let private quoteString str = Args.toWindowsCommandLine [str]
 
     /// Starts the test case.
     let StartTestCase testSuiteName testCaseName =
--- a/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
+++ b/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
@@ -56,7 +56,7 @@ module internal AppVeyorInternal =
 
     /// Adds quotes around the string
     /// [omit]
-    let private quote (str:string) = "\"" + str.Replace("\"","\\\"") + "\""
+    let private quote (str: string) = "\"" + str.Replace("\"", "\\\"") + "\""
 
     /// Adds quotes around the string if needed
     /// [omit]
--- a/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
+++ b/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
@@ -53,10 +53,7 @@ module internal AppVeyorInternal =
                 // because otherwise there might be recursive failure...
                 eprintfn "AppVeyor 'AddMessage' failed: %O" e
 
-    let internal quoteString str =
-        StringBuilder()
-        |> StringBuilder.appendQuotedIfNotNull Some str
-        |> StringBuilder.toText
+    let internal quoteString str = sprintf "\"%s\"" str
 
     /// Starts the test case.
     let StartTestCase testSuiteName testCaseName =
--- a/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
+++ b/src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
@@ -5,6 +5,7 @@ open System.Text
 open Fake.Core
 open Fake.IO
 open Fake.Net
+open Fake.Core
 
 /// [omit]
 module internal AppVeyorInternal =
@@ -53,7 +54,18 @@ module internal AppVeyorInternal =
                 // because otherwise there might be recursive failure...
                 eprintfn "AppVeyor 'AddMessage' failed: %O" e
 
-    let internal quoteString str = sprintf "\"%s\"" str
+    /// Adds quotes around the string
+    /// [omit]
+    let private quote (str:string) = "\"" + str.Replace("\"","\\\"") + "\""
+
+    /// Adds quotes around the string if needed
+    /// [omit]
+    let private quoteIfNeeded str =
+        if String.isNullOrEmpty str then ""
+        elif str.Contains " " then quote str
+        else str
+
+    let private quoteString str = quoteIfNeeded str
 
     /// Starts the test case.
     let StartTestCase testSuiteName testCaseName =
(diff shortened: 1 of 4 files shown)


yazeedobaid:
Thanks for the PR. I think append with an object formatting gives extra options for parameters passed. For example, passing objects with custom `toString` methods. I'm not aware of the history of this function. But as it is used in MSBuild and AppVeyor modules yes it doesn't make sense. In TypeScript module it is used with a string passed as the first parameter.

For the use case, as you said it should also account for nested quotes, so can you please include that in PR. I digged more into legacy process module and found a similar helper that did this ([legacy process module](https://github.com/fsprojects/FAKE/blob/4c22f110105018f85c359c179007c402e4d9d1a1/src/legacy/FakeLib/ProcessHelper.fs#L322-L333)):
```f#
/// Adds quotes around the string
/// [omit]
[<System.Obsolete("FAKE0001 Use the Fake.Core.Process module instead")>]
let quote (str:string) = "\"" + str.Replace("\"","\\\"") + "\""

/// Adds quotes around the string if needed
/// [omit]
[<System.Obsolete("FAKE0001 Use the Fake.Core.Process module instead")>]
let quoteIfNeeded str =
    if isNullOrEmpty str then ""
    elif str.Contains " " then quote str
    else str
```

0x53A:
Unfortunately, just quoting isn't enough, see https://learn.microsoft.com/en-us/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way for the mess that is command line parsing.

Fortunately, the process module of Fake already implements this, and by using ``fromRawCommand`` instead of ``fromRawCommandLine`` we can pass the arg array down to that level instead of having to escape it ourselves.

0x53A:
I'll leave it at this for today - I believe it should be more or less correct, **but I have not done any manual tests**

0x53A:
uhhh, how much do y'all trust the unit/integration test coverage? From my side I believe it's ready, but I have (still) not done any manual testing.

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -1156,18 +1165,18 @@ module DotNet =
         let argString = MSBuild.fromCliArguments msBuildArgs
 
         let binLogPath, args =
-            addBinaryLogger msBuildArgs.DisableInternalBinLog (args + " " + argString) common
+            addBinaryLogger msBuildArgs.DisableInternalBinLog (args @ argString) common
```
yazeedobaid, line 1168: Why space has been removed here?

src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
```diff
@@ -53,10 +54,18 @@ module internal AppVeyorInternal =
                 // because otherwise there might be recursive failure...
                 eprintfn "AppVeyor 'AddMessage' failed: %O" e
 
-    let internal quoteString str =
-        StringBuilder()
-        |> StringBuilder.appendQuotedIfNotNull Some str
-        |> StringBuilder.toText
+    /// Adds quotes around the string
+    /// [omit]
```
yazeedobaid, line 58: Can you please change the function documentation to XML format?

src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
```diff
@@ -53,10 +54,18 @@ module internal AppVeyorInternal =
...
+    let private quote (str: string) = "\"" + str.Replace("\"", "\\\"") + "\""
+
+    /// Adds quotes around the string if needed
+    /// [omit]
```
yazeedobaid, line 62: also here

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -769,11 +769,20 @@ module DotNet =
         let cmdArgs =
             buildCommand
                 (command |> Args.fromWindowsCommandLine |> Seq.toList)
-                (args |> Args.fromWindowsCommandLine |> Seq.toList)
+                (Args.fromWindowsCommandLine args |> Seq.toList)
                 options
 
         run cmdArgs options
 
+    /// the 'exec' above with args as a single string is public and therefore can't be changed, create a second overload with takes a string list
```
yazeedobaid, line 777: Can you please change the function documentation to XML format?

src/app/Fake.DotNet.MSBuild/MSBuild.fs
```diff
@@ -872,24 +867,29 @@ module MSBuild =
...
         |> Seq.map (fun (k, v) -> "/" + k + (if String.isNullOrEmpty v then "" else ":" + v))
-        |> Args.toWindowsCommandLine
+        |> Seq.toList
 
     /// [omit]
     let buildArgs (setParams: MSBuildParams -> MSBuildParams) =
+        let p = MSBuildParams.Create() |> setParams
+        p, fromCliArguments p.CliArguments |> Args.toWindowsCommandLine
+
+    /// [omit]
```
yazeedobaid, line 877: no need to `omit` comment, if the function will not make it to documentation then u can use visibility modified, maybe change it to internal? Also, can you come up with another name instead of `buildArgs2`. While reading this, I had to search if `buildArgs` exists and compare both of them to see the difference.

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -769,11 +769,20 @@ module DotNet =
...
+    let private exec2 (buildOptions: Options -> Options) (command: string) (args: string list) =
```
yazeedobaid, line 778: same as with `buildArgs2`, can you come up with a descriptive name instead of `exec2`

yazeedobaid:
Thanks for the new changes. I have added some comments about them.

Regarding handling the other cases in addition to just escaping quotes, that was discussed recently in [#2735](https://github.com/fsprojects/FAKE/issues/2735). 
@matthid could you please advise on this and the changes made in this PR to replace `fromRawCommandLine` with `fromRawCommand`?

src/app/Fake.DotNet.Cli/DotNet.fs
0x53A: Before, ``args`` and ``argString`` were plain strings, which were just concated with a space between them.

Now, ``args`` and ``argString`` are both ``string list``, which get combined into a single list with the ``@`` operator.

The later code then uses either ``Args.toWindowsCommandLine`` (or ``CreateProcess.fromRawCommand``) to convert the string list into a single, properly quoted, string.

0x53A: with all that said I should also rename the variable from argString to argList

src/app/Fake.Core.Process/CmdLineParsing.fs
```diff
@@ -140,7 +140,7 @@ type FilePath = string
 module Args =
     /// <summary>
     /// Convert the given argument list to a conforming windows command line string, escapes parameter in quotes if
-    /// needed (currently always but this might change).
+    /// needed.
```
0x53A, line 143: this just corrects the comment, the function already did only quote and escape if required.

src/app/Fake.BuildServer.AppVeyor/AppVeyorInternal.fs
```diff
@@ -53,10 +53,8 @@ module internal AppVeyorInternal =
                 // because otherwise there might be recursive failure...
                 eprintfn "AppVeyor 'AddMessage' failed: %O" e
 
-    let internal quoteString str =
-        StringBuilder()
-        |> StringBuilder.appendQuotedIfNotNull Some str
-        |> StringBuilder.toText
+    /// quote and escape the single argument, if required.
```
0x53A, line 56: all other functions in this file use simple non-xml comments, and it's private anyway

voronoipotato:
Who has the rights to merge this? @yazeedobaid do you?

kantora:
Any hints when this could be merged? Can something to be done to push this? As I can understand - without this fix Fake.MSBuild is totally unusable.

voronoipotato:
maybe we should fork and call it fmake.

xperiandri:
Have we finished all the discussions and are confident to merge?

xperiandri (approved):

xperiandri:
@0x53A you may press a merge button now, I suppose it is available

0x53A:
image

PR #2740, opened 2023-03-10
msbuild: include preview in search path

0x53A:
https://github.com/fsprojects/FAKE/issues/1729#issuecomment-352076885

> I probably would accept a PR adding Preview to the search path (after no stable has been found).
Question is whether we should prefer 2017 Preview over stable 2015 for example.

This uses the ordering as proposed by @aaronpowell: VS2022 Preview, VS2019 Release, VS2019 Preview, VS2017 Preview, VS2017 Release, ...


yazeedobaid:
Thanks

Issue #2744, opened 2023-04-21
Update MSBuild.StructuredLogger to latest


KirillOsenkov:
I'm not sure how this works, but this seems to be using an old version of the 
MSBuild.StructuredLogger package:

https://github.com/fsprojects/FAKE/blob/13e30330cae0597aed6154a95a06d21716b18de3/paket.lock#L805

The latest version is:
https://www.nuget.org/packages/MSBuild.StructuredLogger/2.1.814

Without the latest version, builds fail with:

```
Unhandled exception. Fake.Core.BuildFailedException: Target 'PrepareMSBuildTests' failed.
 ---> System.AggregateException: One or more errors occurred. (Unsupported log file format. Latest supported version is 14, the log file has version 15.)
 ---> System.NotSupportedException: Unsupported log file format. Latest supported version is 14, the log file has version 15.
   at Microsoft.Build.Logging.StructuredLogger.BinLogReader.ReadRecordsFromDecompressedStream(Stream decompressedStream)+MoveNext() in C:\MSBuildStructuredLog\src\StructuredLogger\BinaryLogger\BinLogReader.cs:line 218
   at Microsoft.FSharp.Collections.Internal.IEnumerator.choose@171.System.Collections.IEnumerator.MoveNext() in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\seq.fs:line 178
   at Microsoft.FSharp.Collections.SeqModule.ToList[T](IEnumerable`1 source)
   at Fake.DotNet.MSBuildBinLog.getErrorsAndWarnings(String binLogFilePath)
   at Fake.DotNet.MSBuild.handleAfterRun(String command, FSharpOption`1 binLogPath, Int32 exitCode, String project)
   at Fake.DotNet.DotNet.pack(FSharpFunc`2 setParams, String project)
   at Build.prepareMSBuildTests[a](a _arg1) in C:\Farkle\eng\build.fs:line 217
   at Build.initTargets@417-6.Invoke(TargetParameter arg00@)
   at Fake.Core.TargetModule.runSimpleInternal(TargetContext context, Target target)
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace ---
   at Fake.Core.TargetModule.raiseIfError(OptionalTargetContext context)
   at Fake.Core.TargetModule.runOrDefaultWithArguments(String defaultTarget)
   at Build.main(String[] argv) in C:\Farkle\eng\build.fs:line 506
```

[...]

PR #2746, opened 2023-05-16
Bump MSBuild.StructuredLogger to 2.1.820

gtbuchanan:
### Description

Updated with `paket update MSBuild.StructuredLogger --group fakemodule`

Resolves: #2744

## TODO

- [ ] (if new module) the module has been linked from the "side navigation" menu, edit `docs/data.json`.
- [ ] (if new module) the module is in the correct namespace.
- [x] Fake [API guideline](https://fake.build/guide/contributing.html#API-Design-Guidelines) is honored


paket.lock
```diff
@@ -794,19 +799,17 @@ NUGET
...
-    Microsoft.Win32.SystemEvents (6.0) - restriction: || (== net6.0) (&& (== netcoreapp2.1) (>= netcoreapp3.1)) (&& (== netstandard2.0) (>= netcoreapp3.1))
+    Microsoft.Win32.SystemEvents (6.0) - restriction: || (&& (== net472) (>= net7.0)) (== net6.0) (&& (== netcoreapp2.1) (>= netcoreapp3.1)) (&& (== netstandard2.0) (>= netcoreapp3.1))
     Mono.Cecil (0.11.4)
     Mono.Posix.NETStandard (1.0)
-    MSBuild.StructuredLogger (2.1.545)
-      Microsoft.Build (>= 16.10)
-      Microsoft.Build.Framework (>= 16.10)
-      Microsoft.Build.Tasks.Core (>= 16.10)
-      Microsoft.Build.Utilities.Core (>= 16.10)
+    MSBuild.StructuredLogger (2.1.820)
```
cataggar, line 810: 2.1.545 August 2021 to 2.1.820  May 2023. https://www.nuget.org/packages/MSBuild.StructuredLogger

cataggar (approved):

paket.lock
xperiandri: Can we make it wildcard? `2.1.*`

xperiandri: Looks like it changes pretty often

xperiandri: Should be `(>= 2.1.0)` right?

xperiandri:
@gtbuchanan I'm a new maintainer. Do you want to finish your PR?

Numpsy:
So what needs to be done to get this finished?
(There are a few more versions post the one updated to here, and the release notes for the latest one says that it adds support for another new format version, so maybe it should be jumped up to that version)

Numpsy:
An observation:
paket.dependencies currently contains
```
nuget MSBuild.StructuredLogger
        nuget Microsoft.Build
        nuget Microsoft.Build.Framework
        nuget Microsoft.Build.Tasks.Core
        nuget Microsoft.Build.Utilities.Core
```
and those sub packages match the deps of the 2.1.545 version currently in use:
image

However, the current structured loggers versions have reduced deps of 
image

so does paket.dependencies also need changing to match?

Numpsy:
> maybe it should be jumped up to that version

Or not, seems that updating the microsoft.build deps to 17.5 causes the integration tests to fail with the error described in https://github.com/fsprojects/FAKE/issues/2722, so maybe leaving the min version at 2.1.815 that works with the older MSBuild versions would be safest (that should still support the v16 binlog format)

xperiandri:
@Numpsy OK, will you submit a PR?

PR #2743, opened 2023-03-24
Fix dotnet fsi call

pchinery:
### Description

The call to `dotnet fsi` had a typo

PR #2747, opened 2023-05-19
test: fix typo in getting-started.md

dviererbe:
### Description

Just fixed a small typo in the documentation. There is no `dotnet si` command. Just checked it if I did not knew that command:
```
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-si does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```


cataggar:
Duplicate of #2743

Issue #2760, opened 2023-09-02
Security warnings about usage of NuGet.Protocol v6.0


Numpsy:
### Description

I created a CI build using FAKE 6 which also gets run through a Mend analysis, and it raised a warning about references to NuGet.Protocol v 6.0 which has known security vulnerabilities.

Looking at the listing for NuGet.Protocol on nuget.org, it seems that the 6.0.0 versions of all those libraries have actually been delisted due to issues, and several of the updates versions are listed as having issues themselves.

Given the delisting, I think it would be good to bump the version used?

### Repro steps

Version 6.0 seems to be specified at https://github.com/fsprojects/FAKE/blob/13e30330cae0597aed6154a95a06d21716b18de3/paket.lock#L825C1-L825C9

### Known workarounds

As i'm running the build via a .fsproj file, I can locally update the referances to a newer version if I have to.

### Related information 

* Indications of severity
Nuget says the vulnerability is 'high severity'

* Version of FAKE (4.X, 5.X, 6.x)
6.0

xperiandri:
@Numpsy will you prepare a PR?

Numpsy:
That was the intent of #2761 / #2764

xperiandri:
Approved

Numpsy:
This stuff is never ending, versions 6.7.0 is showing as having issues now: https://github.com/advisories/GHSA-68w7-72jg-6qpp :-(

Numpsy:
The NuGet dependencies have been updated to the latest versions now, so I'll close this.

PR #2761, opened 2023-09-02
Update the Nuget.* package versions

Numpsy:
### Description

Updates references to Nuget.Protocol and related libraries to version 6.7.0, to avoid versions with known security vulnerabilities.

refs #2760 

## TODO

- I don't know if this should be updated to the current version (6.7.0) or the most-minimal update that is listed without vulnerabilities (6.0.5)?
- I know pretty much nothing about Paket, so I hope this is the right way to do an update
- It's possible that it'll conflict with the other pending PRs to update dependencies


xperiandri:
Are only `packet.lock` changes required? Not `paket.dependencies` also?

Numpsy:
It didn't update anything else when I did the update, but I do just notice that I updated the fakemodule group and not the fakerunner group, and I suppose those should be in step

xperiandri:
Please, try updating one more time on top of the latest master from scratch

Numpsy:
I tried updating and got merge conflcts, so instead of trying to update here I made a clean PR at #2764 2764

PR #2762, opened 2023-09-23
Run Fantomas on DotNet.fs

Numpsy:
### Description

Just an observation that the CI builds are failing with a Fantomas error, which might be down to this


xperiandri:
Thanks, I fixed that

PR #2767, opened 2023-12-04
Update Github actions

Numpsy:
### Description

When looking at some build results I noticed a number of warnings about old versions of github actions using deprecated versions of node:
```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-dotnet@v1, actions/upload-artifact@v2
```
so I thought i'd have a go at updating them.


xperiandri (approved):

xperiandri:
Thnaks!

PR #2752, opened 2023-07-13
Added shorthash to git functions

voronoipotato:
### Description

Grabs the shorthash

PR #2763, opened 2023-10-01
Project maintenance updates

xperiandri:
### Description

* Migrate to default .gitignore
* Updated Paket to 7.2.1
* Update Fantomas to 6.2.1
* Update FsDocs to 19.1.0
* Update .NET SDK on CI to 6.0.414
* Make .editorconfig more safe
 
- [X] boy scout rule: "leave the code behind in a better state than you found it" (fix warnings, obsolete members or code-style in the places you worked in)     
- [X] Fake [API guideline](https://fake.build/guide/contributing.html#API-Design-Guidelines) is honored

PR #2770, opened 2024-01-18
Added shorthash to git functions

Numpsy:
As #2752 but with an extra commit to try to fix the Fantomas complaints about the formatting.

Doing this here to try to get a CI build run done on the changes.

### DIFF
--- a/src/app/Fake.Tools.Git/Information.fs
+++ b/src/app/Fake.Tools.Git/Information.fs
@@ -80,6 +80,7 @@ module Information =
             match Environment.environVarOrNone "BUILD_SOURCEVERSION" with
             | None -> reraise ()
             | Some s -> s
+
     /// <summary>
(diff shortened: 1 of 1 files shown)


xperiandri:
Maybe we should update to 8.0?
https://github.com/fsprojects/FAKE/blob/82e38df01e4b31e5daa3623abff57e6462430395/.github/workflows/build_and_test.yml#L18

Numpsy:
Is it worth trying to pull the update to 6.0.414 from #2763 into a separate PR? (or may be the latest, 6.0.418) first?

Numpsy:
Hmm, tried doing a build with a newer version at https://github.com/Numpsy/FAKE/actions/runs/7585225090/job/20660587816 and the integration tests failed, but then those tests seem to be downloading a copy of the older version for test use, and I don't know anything about how they work

xperiandri (approved):

PR #2773, opened 2024-02-15
Try running the macOS CI builds on macOS 13

Numpsy:
Just seeing if it has any effect on the intermittent macOS CI issues really - as per the github docs at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources 'macOS-latest' actually uses macOS 12, and the macOS 13 runner also has more CPU cores than the 12 runner

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -9,7 +9,7 @@ jobs:
         runs-on: ${{ matrix.os }}
         strategy:
             matrix:
-                os: [ubuntu-latest, windows-latest, macOS-latest]
+                os: [ubuntu-latest, windows-latest, macOS-13]
         steps:
             - uses: actions/checkout@v4
             - name: Setup .NET 6


Numpsy:
I see that the macos-latest runners are now using macOS 14 / ARM cpus, so not sure if this is helpful any more or not

Thorium:
The current Mac-CI seems to fail...

Numpsy:
The errors in the latest builds sound like https://github.com/dotnet/runtime/issues/64103 ?
That does say it effects ARM cpus though, so I guess the builds having jumped to macOS 14 / ARM runners would be the cause of that,

Numpsy:
The changes in #2763 including updating the .NET 6.0 SDK in the CI builds to ```6.0.414``` which acording to the issue I linked fixes the problem on ARM macs, but that has been WIP for a while.
It does look like it either needs to have the SDK updated, or be changed to use the Intel macOS runners though.

Numpsy:
As it stands, this is the only change that seems to have got the macOS build to pass :-(

Thorium:
Can we get this merged? "Review required", who is the maintainer here?

I think this project would have other useful PRs to merge too. Before those are merged, it's quite pointless to compile projects using FAKE (many F# projects) with .NET 8.0, better just keep Global.json to point at .NET6. Which is good and .NET8 compatible, but sometimes .NET 8 could give some more performance and new features.

Numpsy:
@xperiandri any thoughts?

Thorium:
There are 10+ PRs in line (which most fail CI because of this issue here), and not many commits in recent 6 months. FAKE is still lot used project, and if the maintainers feel here that FAKE is not anymore their interest, could they reach e.g. Amplifying F# to get more resources allocated here?

xperiandri (approved):

Issue #2708, opened 2022-10-26
Update Paket.Core to support .NET 7


mk185147:
### Description

Today Fake uses Paket.Core 6.2.1 that doesn't support .NET 7. There already exists a stable package of Paket.Core 7.1.5 that works fine with .NET 7.

### Repro steps

If you have paket.lock file that was created for .NET 7 running dotnet fake build can fail with error:
```
here was a problem while setting up the environment:
-> Error during parsing of 'C:\Users\mk185147\source\repos\ncr-swt-cfr\site-trust\paket.lock'.
-> invalid parameter 'net7.0' after >= or < in '== net7.0'
```

Thorium:
I have the same error with .NET 8:

```
There was a problem while setting up the environment:
-> Error during parsing of 'C:\git\MyProject\paket.lock'.
-> invalid parameter 'net8.0' after >= or < in '== net8.0) (== netstandard2.1)'
```

PR #2776, opened 2024-05-26
Update Paket.Core to support .NET 8.0

Thorium:
### Description

Updated Paket reference.

- fixes new instance of #2708


xperiandri (approved):

xperiandri:
Do we need to mention this in the release notes? I suppose it does not affect anyone except maintainers and contributors

Thorium:
This does affect if you run Paket restore via Fake. But I'd bundle with it under general .NET8 compatibility updates.

xperiandri:
@Numpsy what do you think?

Numpsy:
There are quite a few ```Could not detect any platforms from 'net8.0' in %package%``` warnings in the CI builds and it'd be nice to get that fixed, but i'm not using Paket myself so I don't know if there are any other ramifications of changing it

Numpsy:
I see that Paket 8 also added support for .NET Framework 4.8.1, so maybe this change would be helpful to anyone using that as well?

NogginBops:
This is blocking us from upgrading to `.net8.0`.
Would it be possible to release a prerelease version of FAKE with these fixes as a temporary measure?

xperiandri:
Let me get back home next Monday and I will figure out how to release

NogginBops:
@xperiandri Just a reminder about this, just in case. :slightly_smiling_face:

NogginBops:
@xperiandri if there is anything I can help with that could speed up the release of the pre-release package I would be willing to give it a try. Also should I open a separate issue for this?

PR #2777, opened 2024-05-26
Update assembly-resolver to not hardcode .NET 6.0

Thorium:
### Description

Some workarounds to issue [mentioned in https://github.com/fsprojects/FAKE/issues/2713]:(https://github.com/fsprojects/FAKE/issues/2713#issuecomment-1312560991)
The .NET version was for no reason hard-coded to 6.0 and not coming from an environment variable that can be overwritten by the user. There is no reason why Fake needs to be updated every time .NET version updates.

1. This commit adds new `FAKE_SDK_RESOLVER_CUSTOM_DOTNET_VERSION` which can be used to change the default "6.0" to whatever other version. So you set it to .NET "18.5" if you want.

2. Also, this changes assembly-resolver to use lists instead of single strings: `SdkVersionRaws` `SdkVersions` and `PaketFrameworkIdentifiers`. The default is still list of single item ["6.0"] to not break any backward-compatibility. However, in many cases the user has multiple NET versions installed in the computer and is just interested of whatever defined in wherever "globals.json" and not what Fake is compiled with. Currently the system just throws build errors, meanwhile user would just want to old VB style on-error-resume-next: "come on, build it now!", e.g. user tries to builds .NET 8 software. So you can define "6.0" and "8.0" and then the assembly-resolver will check what it can find. There are still backward compatible versions kept, `SdkVersionRaw` `SdkVersion` and `PaketFrameworkIdentifier` which just take the first from the list.


xperiandri:
@Thorium formatting required

Thorium:
Thanks, done...

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -308,6 +335,9 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
             paketDependenciesFile: Lazy<Paket.DependenciesFile>
         ) =
         if this.LogLevel.PrintVerbose then
-            Trace.tracefn $"Using .Net {this.SdkVersion.Major} assemblies"
+            let versions =
+                String.Join(", and", this.SdkVersions |> List.map (fun v -> $" .Net{v.Major}"))
```
xperiandri, line 339: Should `net` be from capital?

xperiandri (commented):
Do we need to add `FAKE_SDK_RESOLVER_CUSTOM_DOTNET_VERSION` to the documentation somewhere?

Thorium:
> Do we need to add `FAKE_SDK_RESOLVER_CUSTOM_DOTNET_VERSION` to the documentation somewhere?

yes, and probably to the error message!

xperiandri:
And update release notes, please
https://github.com/fsprojects/FAKE/pull/2778/files

RELEASE_NOTES.md
```diff
@@ -1,5 +1,8 @@
 # Release Notes
 
+## 6.0.1 - 2024-07-11
+* ENHANCEMENT: Fixes for usage in .NET 8.0 enviroment projects.
+
 ## 6.0.1 - 2024-06-04
 * BUGFIX: MSBuild.build adds a bad string at the end of properties, thanks @0x53A - https://github.com/fsprojects/FAKE/issues/2738
 * ENHANCEMENT: Added shorthash to git functions, thanks @voronoipotato - https://github.com/fsprojects/FAKE/pull/2752
```
xperiandri, line 8: I think it must go together as one. @Numpsy right?
```suggestion
## 6.0.1 - 2024-07-11
* BUGFIX: MSBuild.build adds a bad string at the end of properties, thanks @0x53A - https://github.com/fsprojects/FAKE/issues/2738
* ENHANCEMENT: Added shorthash to git functions, thanks @voronoipotato - https://github.com/fsprojects/FAKE/pull/2752
* ENHANCEMENT: Fixes for usage in .NET 8.0 enviroment projects.
```

xperiandri (changes requested):

RELEASE_NOTES.md
Numpsy: I think so

Thorium:
This should be good now.

PR #2768, opened 2023-12-11
add support for /tl:[auto|on|off] msbuild flag

smoothdeveloper:
### Description
see `--tl:[auto|on|off]` on `dotnet build` documentation: https://learn.microsoft.com/en-gb/dotnet/core/tools/dotnet-build

Since this only works with dotnet sdk >= 8, the default case (specifying `Auto`) will not emit the tag.

## TODO

Feel free to open the PR and ask for help

- [ ] New (API-)documentation for new features exist (Note: API-docs are enough, additional docs are in `help/markdown`)
- [x] unit or integration test exists (or short reasoning why it doesn't make sense)
  
  > Note: Consider using the `CreateProcess` API which can be tested more easily, see https://github.com/fsharp/FAKE/pull/2131/files#diff-4fb4a77e110fbbe8210205dfe022389b for an example (the changes in the `DotNet.Testing.NUnit` module)
  
- [ ] boy scout rule: "leave the code behind in a better state than you found it" (fix warnings, obsolete members or code-style in the places you worked in)     
- [ ] (if new module) the module has been linked from the "side navigation" menu, edit `docs/data.json`.
- [ ] (if new module) the module is in the correct namespace.
- [ ] (if new module) the module is added to Fake.sln (`dotnet sln Fake.sln add src/app/Fake.*/Fake.*.fsproj`)
- [ ] Fake [API guideline](https://fake.build/guide/contributing.html#API-Design-Guidelines) is honored

PR #2778, opened 2024-06-04
Update release notes

Numpsy:
refs https://github.com/fsprojects/FAKE/issues/2738#issuecomment-2146284145

This is just grabbing a couple of notable changes, and I'm not sure what the next version will be, so maybe it's a placeholder than a complete change

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## 6.0.1 - 2024-06-04
+* BUGFIX: MSBuild.build adds a bad string at the end of properties, thanks @0x53A - https://github.com/fsprojects/FAKE/issues/2738
+* ENHANCEMENT: Added shorthash to git functions, thanks @voronoipotato - https://github.com/fsprojects/FAKE/pull/2752
+
 ## 6.0.0 - 2023-02-20
 * ENHANCEMENT: Site UI fixes and documentation link fixes.
 


Numpsy:
Also, the macOS CI builds might fail unless we do https://github.com/fsprojects/FAKE/pull/2773 or similar

voronoipotato:
Looks great! Thanks @Numpsy !! 🎉🎉🎉

voronoipotato (approved):

xperiandri (approved):

xperiandri:
Does anyone know how to fix the build?

Numpsy:
> Does anyone know how to fix the build?

If the issue is https://github.com/dotnet/runtime/issues/64103 then it looks like either change the macOS runner from macOS-latest to macOS(12/13) (e.g. #2773) or update the .NET SDK used in the build to 6.0.300 or newer.

smoothdeveloper:
@xperiandri I noticed in #2768 (which also updates release notes) that you rebased / repushed commits I did in that branch, I think it is better to just append commits and let the branch author do rebase in case it is needed, there is also concern it can mess the contribution history in github (which is nice for people that begin contribution).

#2768 should be merged and released, IMO, it is not because macOS CI fails in that particular branch, that it is related to it, most FAKE users can't leverage terminal logger goodness until it ships.

xperiandri:
@smoothdeveloper let's fix the build in a separate PR and rebase this one on top. Anyway, if a build does not happen I will not be able to release anything. A release also requires building 🙂

voronoipotato:
@xperiandri how do we update .net core to 6.0.5 for that the macOS-latest target so we can keep moving?

xperiandri:
Submit another PR with a fix

Numpsy:
> @xperiandri how do we update .net core to 6.0.5 for that the macOS-latest target so we can keep moving?

I tried doing it in #2780 and got a different error :-(

Numpsy:
Grasping at straws a bit here, but  - https://github.com/fsprojects/FAKE/pull/2781.
(Unclear if there is actually a real FAKE issue here on ARM cpu macs that might effect end users as well)

voronoipotato:
@Numpsy  maybe we should reach out to people in the slack and discord and see if anyone has an arm macbook that they can help test this with. I have an arm surface pro x, but I doubt that will be helpful.

Numpsy:
At this point, I think it might be nice to just try to get the build pipeline running successfully on the Intel Mac runners, and have 'fix ARM problems' as a seperate work item once there's a more stable base

Numpsy:
Hurrah, build succeeded

PR #2780, opened 2024-06-16
Update the .NET SDK in the CI builds to 6.0.300

Numpsy:
### Description

A minimal update to the .NET 6 SDK version, to see if it fixes the macOS/ARM CI failures discussed in https://github.com/fsprojects/FAKE/pull/2778

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -15,7 +15,7 @@ jobs:
             - name: Setup .NET 6
               uses: actions/setup-dotnet@v4
               with:
-                  dotnet-version: '6.0.101'
+                  dotnet-version: '6.0.300'
             - name: Restore dotnet tools
               run: dotnet tool restore
             - name: remove current fake runner tool
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
             - name: Setup .NET 6
               uses: actions/setup-dotnet@v4
               with:
-                  dotnet-version: '6.0.101'
+                  dotnet-version: '6.0.300'
             - name: Restore dotnet tools
               run: dotnet tool restore
             - name: remove current fake runner tool
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
-    "sdk" : {
-        "version": "6.0.101",
+    "sdk": {
+        "version": "6.0.300",
         "rollForward": "latestMinor"
     }
 }


Numpsy:
Hmm, the macOS CI failed and the built output contains

```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.TypeInitializationException: The type initializer for '<StartupCode$Fake-DotNet-Cli-IntegrationTests>.$Fake.DotNet.Cli.IntegrationTests.TemplateTests' threw an exception.
 ---> System.DllNotFoundException: Unable to load shared library 'MonoPosixHelper' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libMonoPosixHelper, 0x0001): tried: 'libMonoPosixHelper' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibMonoPosixHelper' (no such file), '/usr/lib/libMonoPosixHelper' (no such file, not in dyld cache), 'libMonoPosixHelper' (no such file), '/usr/local/lib/libMonoPosixHelper' (no such file), '/usr/lib/libMonoPosixHelper' (no such file, not in dyld cache)
```

So now I wonder if the old version of Mono.Posix it's using maybe only has Intel CPU support?

Numpsy:
There are some integration tests in https://github.com/fsprojects/FAKE/blob/master/src/test/Fake.Core.IntegrationTests/Fake.DotNet.sdkAssemblyResolver.fs which are doing explicit things with the 6.0.10X SDK, and I start to wonder if that simply isn't going to work on ARM macs?

Thorium:
This could help the CI failure: https://github.com/fsprojects/FAKE/pull/2783

xperiandri:
But it is already rebased

Thorium:
Maybe the failure is this asking 6.0.100:
https://github.com/fsprojects/FAKE/blob/master/integrationtests/core-reference-assemblies-net60101-rollforward/before/global.json

...as the CI reports it here:
https://github.com/fsprojects/FAKE/actions/runs/10092902189/job/27907519123?pr=2780#step:8:9360

Numpsy:
I tried updating some of those in https://github.com/Numpsy/FAKE/commits/6.0.131/ a while ago and hit some other failures.
At this point, I don't know if it's worth spending time updating the 6.0 SDKs or if it's better to put the effort into switching everything to .NET 8.0 before 6.0 goes out of support.

xperiandri:
Probably 8 is better

Numpsy:
The build at https://github.com/Numpsy/FAKE/actions/runs/10161143249 from the branch at https://github.com/Numpsy/FAKE/tree/bump_flail seems to have passed with the 6.0.30x SDK, though there are still a couple of changes n top of the SDK update in there

PR #2782, opened 2024-07-11
Potential fix for Mac build

Thorium:
Testing if we could fix the build on mac-latest:

It fails for .NET 6 installation:
https://github.com/dotnet/runtime/issues/64103


Thorium:
Not working.

PR #2783, opened 2024-07-14
Update FSharp.Core dependency

Thorium:
This updates FSharp core to latest stable, but nothing else.

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -4,7 +4,7 @@ GROUP Build
 CONTENT: NONE
 NUGET
   remote: https://api.nuget.org/v3/index.json
-    FSharp.Core (6.0.6)
+    FSharp.Core (8.0.301)
     ILRepack (2.0.18)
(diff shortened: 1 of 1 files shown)


Thorium:
This is because some existing unit-tests seems to do add reference to FSharp.Core without declaring a version, so it downloads a latest, which used to be F# 6 but is now F# 8, so then the unit-tests raise: 

"Paket resolved a FSharp.Core with version 8, but fake runs with a version of '6'. This is not supported. Please either lock the version via 'nuget FSharp.Core <nuget-version>' or upgrade fake. Read https://github.com/fsharp/FAKE/issues/2001 for details."

FSharp.Core dependency version is independent of any target-frameworks, so this is not touching that, target is still .NET6.

xperiandri (approved):

xperiandri (approved):

PR #2784, opened 2024-07-15
Update the upload-artifact CI build action to version 4

Numpsy:
In the current CI builds, there are a number of instances of this warning:

image

In addition, the documentation at https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/ documents that the v3 action will be deprecated and not useable later this year.

That doc also says ```While v4 of the artifact actions improves upload and download speeds by up to 98% ``` which may or may not effect the intermittent artifact upload timeout that has caused the CI builds to fail a few times recently.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -27,7 +27,7 @@ jobs:
             - name: Build
               run: dotnet fake build --parallel 3
             - name: publish build artifacts
-              uses: actions/upload-artifact@v3
+              uses: actions/upload-artifact@v4
               with:
-                  name: fake-artifacts
+                  name: fake-artifacts-${{ matrix.os }}
                   path: release/artifacts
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             - name: Build
               run: dotnet fake build --parallel 3
             - name: publish build artifacts
-              uses: actions/upload-artifact@v3
+              uses: actions/upload-artifact@v4
               with:
-                  name: fake-artifacts
+                  name: fake-artifacts-${{ matrix.runs-on }}
```
Numpsy, line 32: The docs at https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md state that artifacts need unique names with the v4 action

xperiandri (approved):

PR #2795, opened 2024-08-05
Fixed null reference exception on compilation results moving

Thorium:
As the tooltip says, the array may contain null-references:

image

Now, if it does, the Seq.filter below without null-check dies and kills the whole run for "Null reference exception".
Check added to skip nulls.

Also when "file already exists", added trace message which file.

### DIFF
--- a/src/app/Fake.Runtime/CompileRunner.fs
+++ b/src/app/Fake.Runtime/CompileRunner.fs
@@ -79,7 +79,7 @@ let tryRunCached (c: CoreCacheInfo) (context: FakeContext) : RunResult =
 
                 match
                     types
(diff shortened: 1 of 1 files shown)


xperiandri (approved):

xperiandri:
Hm https://github.com/fsprojects/FAKE/actions/runs/10631745314/job/29473273685

Thorium:
I don't think this failure is related to this commit, rather it's still some unstability of integration test.
In this case, I've only run this locally on Windows that does work well.

I expect this is something about the test that runs `integrationtests/core-no-dependencies-hello-world/before` folder build, not having any restriction settings (like global.json), so it tries to run it on .Net 8.0 and it fails for something?

PR #2796, opened 2024-08-06
Update fakemodule MSBuild.StructuredLogger

Thorium:
Updated latest MsBuild StructureLogger (and its dependencies) to avoid issues like https://github.com/fsprojects/FAKE/issues/2748


Thorium:
Hmm, this is not solving the issues properly.

Thorium:
The log file format keeps updating so #2744 is relevant, now the latest version is already 18.

PR #2797, opened 2024-08-06
SdkAssemblyResolver improvements

Thorium:
It still seems we have unreliable integration tests.

I still blame the SdkAssemblyResolver not being able to pick proper .NET versions.

There were some "try finds" which just took first ever found version, and I think it should rather collect a list of versions to make the decision of proper version later.

Also added verbosity to intergration-test-log so we could better see what goes wrong, if something goes wrong.

### DIFF
--- a/src/app/Fake.Runtime/SdkAssemblyResolver.fs
+++ b/src/app/Fake.Runtime/SdkAssemblyResolver.fs
@@ -143,23 +143,28 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
         if this.LogLevel.PrintVerbose then
             Trace.tracefn "Trying to resolve runtime version from network.."
 
-        try
-            let sdkVersionReleases =
-                ProductCollection.GetAsync()
-                |> Async.AwaitTask
-                |> Async.RunSynchronously
-                |> List.ofSeq
-                |> List.filter (fun product ->
-                    this.SdkVersionRaws
-                    |> List.exists (fun raws -> product.ProductVersion.Equals raws))
+        let sdkSet = this.SdkVersionRaws |> Set.ofList
 
-            sdkVersionReleases
-            |> List.collect (fun rel ->
-                rel.GetReleasesAsync()
-                |> Async.AwaitTask
-                |> Async.RunSynchronously
-                |> List.ofSeq)
+        try
+            task {
+                let! productCollection = ProductCollection.GetAsync()
+
+                let sdkVersionReleases =
+                    productCollection
+                    |> Seq.filter (fun product -> sdkSet |> Set.exists (fun raws -> product.ProductVersion.Equals raws))
+
+                let! releases =
+                    sdkVersionReleases
+                    |> Seq.map (fun rel -> rel.GetReleasesAsync())
+                    |> System.Threading.Tasks.Task.WhenAll
+
+                return releases |> Seq.collect (fun sdks -> sdks |> Seq.toList)
+            }
+            |> Async.AwaitTask
+            |> Async.RunSynchronously
+            |> Seq.toList
             |> Some
+
         with ex ->
             Trace.traceError $"Could not get SDK runtime version from network due to: {ex.Message}"
             None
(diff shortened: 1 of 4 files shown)


src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -149,14 +149,16 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
-            sdkVersionReleases.GetReleasesAsync()
-            |> Async.AwaitTask
-            |> Async.RunSynchronously
-            |> List.ofSeq
+            sdkVersionReleases
+            |> List.collect (fun rel ->
+                rel.GetReleasesAsync()
+                |> Async.AwaitTask
+                |> Async.RunSynchronously
+                |> List.ofSeq)
```
xperiandri, line 161: why not?
```suggestion
            |> Seq.map (fun rel -> rel.GetReleasesAsync())
            |> Task.WaitAll
            |> Seq.collect
            |> Seq.toLost
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -194,9 +196,14 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
             |> List.ofSeq
             |> List.exists (fun sdk -> sdk.Version.Equals(version))
```
xperiandri, line 197: I guess this is enough
```suggestion
            |> Seq.exists (fun sdk -> sdk.Version.Equals(version))
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
Thorium: True that could be better. I expect the list to be very small so no firewall blocks on parallelism. The Seq.toList's side-effected brother Seq.toLost :-D

Thorium: This needs a bit more refactoring, because Tasl.WaitAll is not not returning the result and WhenAll is not waiting the task.

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -143,23 +143,28 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
+                let sdkVersionReleases =
+                    productCollection
+                    |> Seq.filter (fun product -> sdkSet |> Set.exists (fun raws -> product.ProductVersion.Equals raws))
+
+                let! releases =
+                    sdkVersionReleases
+                    |> Seq.map (fun rel -> rel.GetReleasesAsync())
+                    |> System.Threading.Tasks.Task.WhenAll
+
+                return releases |> Seq.collect (fun sdks -> sdks |> Seq.toList)
```
xperiandri, line 161: Maybe without lists allocation?
```suggestion
                return releases |> Seq.collect id
```

Thorium: The issue it's a "list of arrays" so either Seq.toArray |> Array.collect |> Array.toList or this. As a list is small, like max 10, I don't see a difference.

xperiandri (approved):

xperiandri:
I would prefer release pipeline to be fixed and run to completion before we merge new fixes

Thorium:
This one may actually help the pipeline. :-)

xperiandri:
For now the issue is that pushing the same version of the package to NuGet does not skip it but fails

src/app/Fake.Runtime/SdkAssemblyResolver.fs
xperiandri: Will not it implicitly convert arrays to seqs?

Thorium:
That's intentional I guess, NuGet packages cannot be updated.

src/app/Fake.Runtime/SdkAssemblyResolver.fs
Thorium: it didn't

xperiandri (approved):

PR #2799, opened 2024-08-06
Some code clean-ups and performance improvements

Thorium:
Some trivial code clean-up and performance improvements:

- isNull instead of "= null": https://latkin.org/blog/2015/05/18/null-checking-considerations-in-f-its-harder-than-you-think/
- Structs/ValueTuples on active patterns that doesn't carry lot of data
- Seq.map + Seq.concat = Seq.collect

### DIFF
--- a/src/app/Fake.Core.Target/Target.fs
+++ b/src/app/Fake.Core.Target/Target.fs
@@ -861,7 +861,7 @@ module Target =
             let isValidTarget name = targetLeftSet.Contains(name)
 
             let canBeExecuted (t: Target) =
-                t.Dependencies @ t.SoftDependencies |> Seq.filter isValidTarget |> Seq.isEmpty
+                t.Dependencies @ t.SoftDependencies |> Seq.exists isValidTarget |> not
 
             let map =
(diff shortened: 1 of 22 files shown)


src/app/Fake.Core.Target/Target.fs
```diff
@@ -861,7 +861,7 @@ module Target =
             let isValidTarget name = targetLeftSet.Contains(name)
 
             let canBeExecuted (t: Target) =
-                t.Dependencies @ t.SoftDependencies |> Seq.filter isValidTarget |> Seq.isEmpty
+                t.Dependencies @ t.SoftDependencies |> Seq.exists isValidTarget |> not
```
xperiandri, line 864: maybe for readability?
```suggestion
                t.Dependencies @ t.SoftDependencies |> (not << Seq.exists) isValidTarget
```

Thorium: You can't pipe it that way, doesn't compile.

xperiandri (approved):

PR #2802, opened 2024-08-09
Release-docs TOKEN_GITHUB to GITHUB_TOKEN

Thorium:
Fix release yaml docs token

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,5 +48,5 @@ jobs:
                   NUGET_SOURCE: ${{ secrets.NUGET_SOURCE }}
                   RELEASE_USER_GITHUB: ${{ secrets.RELEASE_USER_GITHUB }}
                   REPOSITORY_NAME_GITHUB: ${{ secrets.REPOSITORY_NAME_GITHUB }}
-                  GITHUB_TOKEN: ${{ secrets.TOKEN_GITHUB }}
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: ${{github.workspace}}/release/dotnetcore/Fake.netcore/win7-x64/fake.exe build -t FastRelease
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,5 +48,5 @@ jobs:
                   NUGET_SOURCE: ${{ secrets.NUGET_SOURCE }}
                   RELEASE_USER_GITHUB: ${{ secrets.RELEASE_USER_GITHUB }}
                   REPOSITORY_NAME_GITHUB: ${{ secrets.REPOSITORY_NAME_GITHUB }}
-                  TOKEN_GITHUB: ${{ secrets.TOKEN_GITHUB }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -48,5 +48,5 @@ jobs:
                   NUGET_SOURCE: ${{ secrets.NUGET_SOURCE }}
                   RELEASE_USER_GITHUB: ${{ secrets.RELEASE_USER_GITHUB }}
                   REPOSITORY_NAME_GITHUB: ${{ secrets.REPOSITORY_NAME_GITHUB }}
-                  TOKEN_GITHUB: ${{ secrets.TOKEN_GITHUB }}
+                  GITHUB_TOKEN: ${{ secrets.TOKEN_GITHUB }}
```
xperiandri, line 51: Actually it should have been like this. Maybe at the time of pipeline `GITHUB_TOKEN` not existed.
```suggestion
                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Probably we can even remove this line

xperiandri: It works like this here
https://github.com/fsprojects/FSharp.Data.GraphQL/blob/d96b065b95d553cdaac68cd2697490c5b618be44/.github/workflows/publish_ci.yml#L119

xperiandri: Can the token passed from environment be more restrictive then the one set manually?

baronfel: It can be, but even the token permissions can be described in the yaml, so it should be more clear what's required. See https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token for details.

Thorium:
Integration test failed but I expect it's nothing to do with this commit, they are just bit un-reliable, addressed at #2797 .

xperiandri (approved):

xperiandri:
```
remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
```

Thorium:
Ok so it gets further than before, that's good news!

The messages you pasted are just informational messages. It is doing token auth, not password auth.
I think what goes wrong is this:

image

`Authentication failed`

So CI is using that user to try to push the release-docs, but the token is for some other user?

Numpsy:
At a risk of not having a clue what I'm doing (never used gh-pages) - an observation - 

The build script at https://github.com/fsprojects/FAKE/blob/f07f74dd37bf6dad05dec31f1b3874339f3c5103/build.fsx#L997 uses ```x-oauth-basic``` when building the URL, but looking at the code for some github pages publishing actions, e.g. https://github.com/peaceiris/actions-gh-pages/blob/0b7567fde6f7517edcc13d8ffa2d89cd8734d47c/src/set-tokens.ts#L100, I see them using ```x-access-token``` instead.
Is that something that needs reviewing?

Issue #2803, opened 2024-08-11
[REGRESSION] FAKE 6.1.0 fails with dotnet preview versions


SteveGilham:
### Description

Upgrading the AltCover build from v6.0.0 to v6.1.0, the first use of `Fake.DotNet.DotNet.exec` fails when parsing the version string in `global.json`. The same is true for other APIs, like ` DotNet.restore`

### Repro steps

1. Install SDK 9.0.100-preview.6.24328.19
2. Invoke it with a global.json like
```
{
  "sdk": {
    "version": "9.0.100-preview.6.24328.19",
    "rollForward": "latestMinor"
  }
}
```
3. Run a command with `DotNet.exec`

### Expected behavior

It just works, i.e. unchanged since v 6.0.0

### Actual behavior

Exception thrown
```
Unhandled exception. System.TypeInitializationException: The type initializer for 'AltCover.Targets' threw an exception.
 ---> System.TypeInitializationException: The type initializer for '<StartupCode$build>.$Targets' threw an exception.
 ---> System.Exception: Could not parse `sdk.version` from global.json at 'C:\Users\email\Documents\Github\altcover\global.json': Version string portion was too short or too long. (Parameter 'input')
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1448.Invoke(String message)
   at Fake.DotNet.DotNet.tryGetSDKVersionFromGlobalJsonDir(String startDir) in D:\a\FAKE\FAKE\src\app\Fake.DotNet.Cli\DotNet.fs:line 77
   at Fake.DotNet.DotNet.Options.Create() in D:\a\FAKE\FAKE\src\app\Fake.DotNet.Cli\DotNet.fs:line 546
   at Fake.DotNet.DotNet.exec(FSharpFunc`2 buildOptions, String command, String args) in D:\a\FAKE\FAKE\src\app\Fake.DotNet.Cli\DotNet.fs:line 772
   at <StartupCode$build>.$Targets..cctor() in C:\Users\email\Documents\Github\altcover\Build\targets.fs:line 174
```

### Known workarounds

Revert to 6.0.0

### Related information 

* Operating system - Windows 11
* Branch - using the 6.1.0 release from NuGet
* .NET Runtime, CoreCLR or Mono Version - dotnet SDK 9.0.100-preview.6.24328.19
* Indications of severity - complete stopper for pre-release development
* Version of FAKE (4.X, 5.X, 6.x) - 6.1.0

[...]

PR #2757, opened 2023-08-07
Fixed `NullReferenceException` on missing `version` in `global.json`

xperiandri:
- fixes #2756

- [ ] unit or integration test exists (Is it needed? Where to write it?)
  
- [X] boy scout rule: "leave the code behind in a better state than you found it" (fix warnings, obsolete members or code-style in the places you worked in)     
- [X] Fake [API guideline](https://fake.build/guide/contributing.html#API-Design-Guidelines) is honored

PR #2771, opened 2024-01-29
Allow `RunSettingsArguments` to be set

TheAngryByrd:
### Description

This allows `RunSettingsArguments` to be set in `DotNet.test`.  This is useful for configuring such as in YoloDev https://github.com/YoloDev/YoloDev.Expecto.TestSdk?tab=readme-ov-file#configuration

If available, link to an existing issue this PR fixes. For example:

- fixes #2376 

## TODO

Feel free to open the PR and ask for help

- [ ] New (API-)documentation for new features exist (Note: API-docs are enough, additional docs are in `help/markdown`)
- [x] unit or integration test exists (or short reasoning why it doesn't make sense)
  
  > Note: Consider using the `CreateProcess` API which can be tested more easily, see https://github.com/fsharp/FAKE/pull/2131/files#diff-4fb4a77e110fbbe8210205dfe022389b for an example (the changes in the `DotNet.Testing.NUnit` module)
  
- [x] boy scout rule: "leave the code behind in a better state than you found it" (fix warnings, obsolete members or code-style in the places you worked in)     
- [ ] (if new module) the module has been linked from the "side navigation" menu, edit `docs/data.json`.
- [ ] (if new module) the module is in the correct namespace.
- [ ] (if new module) the module is added to Fake.sln (`dotnet sln Fake.sln add src/app/Fake.*/Fake.*.fsproj`)
- [ ] Fake [API guideline](https://fake.build/guide/contributing.html#API-Design-Guidelines) is honored

PR #2804, opened 2024-08-13
Remove `Version.Parse` check from `global.json` parsing

Numpsy:
refs #2803 - The check seems to break if the version is a pre-release version such as ```9.0.100-preview.6.24328.19```

The Parse call was added as part of the null change in #2757 but i'm not sure if there was any reason for it other than an attempt at a sanity check?

(needs release notes if it the change is ok)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 ## 6.1.1 - 2024-08-19
 * BUGFIX: Assembly resolver to pick a found SDK instead of first SDK - https://github.com/fsprojects/FAKE/pull/2797/files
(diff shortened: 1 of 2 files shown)


xperiandri (approved):

xperiandri:
Should this be mentioned in the release notes?

Numpsy:
Should probably be listed as a fix for #2803, but I left out the release notes at the time as #2797 and #2771 had both already added a 6.1.1 to the release notes and it might be best to merge those first to avoid creating more conflicts

xperiandri:
#2771 also conflicts.
Could you rebase or merge?

Numpsy:
Rebased and added a release note entry

PR #2805, opened 2024-08-13
Update System.Text.Json in the fakerunner group to 8.0.4 to fix vulnerability

Numpsy:
Update System.Text.Json in the fakerunner group to 8.0.4 to fix https://github.com/advisories/GHSA-hh2w-p6rv-4g7w

### Description

fake-cli has a transitive dependency on System.Text.Json, via Microsoft.Deployment.DotNet.Releases.

The fake-cli 6.1.0 nuget package contains a v7.0.x copy of that, which is listed as suffering from https://github.com/advisories/GHSA-hh2w-p6rv-4g7w

This is an attempt at fixing that, by updating to the fixed 8.0.4 version.

I'm not sure if this is the best way to do the update, as I'm still not getting to grips with managing transitive dependencies with Paket :-(


Numpsy:
CI build over at https://github.com/Numpsy/FAKE/actions/runs/10375639274 worked fwiw

xperiandri (approved):

xperiandri (approved):

PR #2809, opened 2024-08-26
Ignore errors in DotNetPushChocolateyPackage when IGNORE_CONFLICT is …

Numpsy:
…true

refs https://github.com/fsprojects/FAKE/issues/2800#issuecomment-2304386375 and following comments.

This was an attempt at getting the release build to work by ignoring the 'version already exists' error that it's getting when trying to push a build to Chocolately with a version number that already exists.
The issue is that it doesn't seem to have much context about what caused the error, so just ignoring all errors would ignore failures for other reasons as well.


xperiandri (approved):
Looks good

PR #2812, opened 2024-08-30
Remove the `System.Runtime.Loader` NuGet reference from `Fake.Runtime`

Numpsy:
### Description

When looking at the dependencies of the[ Fake.Runtime NuGet package on NuGet.org](https://www.nuget.org/packages/Fake.Runtime#dependencies-body-tab), I noticed that the Fake.Runtime package depends on System.Runtime.Loader:

image

And given that Fake.Runtime is only built as .NET 6.0, I wondered if it needs to use the NuGet package, or if it can use the in-box version instead?

Any thoughts?


xperiandri:
If it builds and tests pass I suppose it is a good modification

xperiandri (approved):

PR #2813, opened 2024-09-01
Remove `Microsoft.SqlServer.SqlManagementObjects` from the fakemodule group on `paket.dependencies`

Numpsy:
### Description

I don't know any history about these dependencies, so this is just anothet observation:

paket.dependencies has a reference to ```Microsoft.SqlServer.SqlManagementObjects```.
However, I don't see that attached to any of the projects.

Looking at the change history in that area, it looks like the reference was added in https://github.com/fsprojects/FAKE/pull/2074 as part of a module update for Fake 5, but according to https://github.com/fsprojects/FAKE/blob/b658df48c4012db211fdb2094d1fafca5c1e7755/docs/articles/fake-v6-release-announcement.md?plain=1#L1070 that whole library was removed in Fake 6.
The ```Fake.Sql.SqlPackage``` library that is mentioned as a replacement doesn't seem to reference any of these SQL libraries directly.

So - I wonder if the reference here was just left behind when the consumers were removed?

Reason for asking is that there is a substantial dependency chain which includes some old libraries with known CVEs (e.g. Microsoft.Data.SqlClient 4.0), and if it is unused then removing it removes a substantial amount of things from paket.lock.


xperiandri:
Looks good. Tests pass

xperiandri (approved):

PR #2818, opened 2024-09-03
Build with .NET 8.0

Thorium:
Changes with this PR:

- Change global.json .NET 6 to 8
- Search and replace targetframeworks from fsproj files to add net8.0
- Add net8.0 to paket.dependencies 
- dotnet paket install to find .NET8 compatible dependencies
- Expecto had to be hardcoded for now, because some tests are running on netstandard2.0 library (hopefully we can update this separately later)
- MSBuild.StructuredLogger problem: DisableInternalBinLog = true had to be added to build.fsx NuGet commands (hopefully we can update this separately later)
- A few places of code had new overrides so had to explicitly type to strings
- SdkAssemblyResolver to default .NET 8 as well
- Readme update

Possible issue:
If tests restore still previous fake-cli 6.1.1 and not this version, then they might fail.


paket.lock
```diff
@@ -613,692 +613,608 @@ NUGET
...
     Mono.Cecil (0.11.5)
     Mono.Posix.NETStandard (1.0)
-    MSBuild.StructuredLogger (2.1.815)
-      Microsoft.Build.Framework (>= 16.10)
-      Microsoft.Build.Utilities.Core (>= 16.10)
+    MSBuild.StructuredLogger (2.2.317)
+      DotUtils.StreamUtils.Sources (>= 0.0.7)
+      Microsoft.Build.Framework (>= 17.5)
+      Microsoft.Build.Utilities.Core (>= 17.5)
+      Microsoft.SourceLink.GitHub (>= 8.0)
```
Numpsy, line 797: fwiw I'm wondering if the latest StructuredLogger package [might have accidentaly made some dev dependencies into runtime dependencies](https://github.com/KirillOsenkov/MSBuildStructuredLog/pull/798#discussion_r1742775769) - maybe some of these new deps might not really be needed anyway

Thorium:
The real issue with the build.fsx is that it's recursive: In the beginning it says "use paket `source release/dotnetcore`" which is a local folder, that is having the latest build result. If there is nothing yet, then it uses latest release from the NuGet. 

But how can you build .NET 8 local package, when there is no .NET 8 compatible packages in NuGet?

Numpsy:
> But how can you build .NET 8 local package, when there is no .NET 8 compatible packages in NuGet?

How was it done when .NET 6.0 support was first added?

Do the Microsoft.Build.* package references in build.fsx need unpinning from 17.3.2 to support .NET 8 buiids?

paket.lock
KirillOsenkov: I fixed this in https://www.nuget.org/packages/MSBuild.StructuredLogger/2.2.337

Thorium:
If I build this locally with `dotnet build` or `dotnet pack` it works on both my Mac and Windows.
But if I try to run it with `dotnet fake run` it fails on "Fake is not defined" which doesn't make sense to me. Maybe I'm missing something trivial. At least it does read the dll so that is already further than the current Fake 6.1.1 which just fails for https://github.com/fsprojects/FAKE/issues/2314#issuecomment-2327156343

I have a bad hunch that version update could have been done by releasing some -alpha package without Fake, and then moved to build.fsx when it's uploaded to NuGet. But hopefully there is a way to avoid that, as it sounds slow and error-prone.

Numpsy:
> But if I try to run it with dotnet fake run it fails on "Fake is not defined" which doesn't make sense to me

Not sure if this is an issue or not, but in case it is - 

When testing a local build I noticed that ```build.fsx.lock``` has a ```RESTRICTION: || (== net6.0) (== netstandard2.0)``` at the top, and wondered if that could be an issue if dealing with the latest versions of libraries like Microsoft.Build which don't have either of those TFMs (just .NET 8.0 and FW 4.7.2)?

I tried adding .NET 8.0 in the fsx file at https://github.com/fsprojects/FAKE/commit/2852870eb7f942cb923442cecbee7926189727a4 as a test and it failed to load some v8.0.0 libraries - https://github.com/Numpsy/FAKE/actions/runs/10702782852/job/29671766410#step:10:950 - so not sure if that counts as an improvement.

Anyway - is it possible that the default frameworks at https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/src/app/Fake.Runtime/FakeHeader.fs#L182 should be different in a .NET 8.0 build?

Thorium:
if you run locally `dotnet fake --version` then you can observe that it has picked FakePath: .../net6.0/... for whatever reason even there should be both net6.0 and net8.0 in the tool package, and the global.json tells to use net8.0

Thorium:
lf I open \.fake\ dll with IlSpy, it seems to be compiled to .NET core 3.1, which could explain why it can't handle the libraries referenced: 
image

The reason is CompileRunner.fs calling FSharpChecker.Create()
and I expect this is related: https://github.com/dotnet/fsharp/issues/14243

Thorium:
> When testing a local build I noticed that `build.fsx.lock` has a `RESTRICTION: || (== net6.0) (== netstandard2.0)` at the top, 

Oh, but netstandard 2.0 is not compatible with .net8. Shouldn't it be 2.1?

src/app/Fake.Runtime/Environment.fs
```diff
@@ -73,5 +73,17 @@ let inline internal fakeContextAssembly () =
...
+        | null -> ""
+        | itms when itms.Length > 0 ->
+            (itms |> Seq.head :?> System.Runtime.Versioning.TargetFrameworkAttribute)
+                .FrameworkName
+        | _ -> ""
+
 /// Gets the FAKE Version string
-let fakeVersionStr = sprintf "FAKE 6 - F# Make (%s)" fakeVersion
+let fakeVersionStr =
+    sprintf "FAKE 6 - F# Make (%s) (running on %s)" fakeVersion (getPlatform ())
```
xperiandri, line 89:
```suggestion
    // TODO: Change to 8 for 8th release
    sprintf "FAKE 6 - F# Make (%s) (running on %s)" fakeVersion (getPlatform ())
```

build.fsx
```diff
@@ -3,13 +3,14 @@
...
 #r "paket:
+source release/temp-recursion-break
 source release/dotnetcore
 source https://api.nuget.org/v3/index.json
 nuget FSharp.Core
-nuget Microsoft.Build 17.3.2
-nuget Microsoft.Build.Framework 17.3.2
-nuget Microsoft.Build.Tasks.Core 17.3.2
-nuget Microsoft.Build.Utilities.Core 17.3.2
+nuget Microsoft.Build 17.11.4
```
Numpsy, line 10: Question: Does it still need to reference all 4 of these Microsoft.Build packages here?

Older versions of MSBuild.StructuredLogger used to depend on all of these, but newer versions (including the 2.1.815 build currently used by Fake) only depend on ```Framework``` and ```Utilities.Core``` so I wonder if the base one and ```Tasks.Core``` can be removed now?

Numpsy:
I made an attempt at debugging the cli tool locally to see if I could see what's going on with the failure to load System.Globalization.dll seen in the CI builds, and made an observation:

When debugging it, the call to LoadFromAssemblyName at https://github.com/fsprojects/FAKE/blob/3a3e00c73041162b83f878960189e5736e836dad/src/app/Fake.Runtime/CoreCache.fs#L331 manages to load the assembly -
image

However, the ```isFramework``` check a few lines down then discards the loaded dll. This is different from the current .NET 6.0 build, which keeps it.

Looking at the difference, it appears that the .NET 6.0 version of System.Globalization has the ```.NETFrameworkAssembly``` attribute, and the .NET 8 version doesn't.

Searching about a bit for that attribute value I spotted https://github.com/dotnet/runtime/pull/89490, where said property was removed from all in-box assembles in .NET 8.0.

So - I wonder if the logic there is just wrong in .NET 8 and newer?

Numpsy:
I guess the latest failure is becuse https://github.com/fsprojects/FAKE/blob/3a3e00c73041162b83f878960189e5736e836dad/build.fsx#L807 needs to specify whether to publish as .NET 6.0 or 8.0? (assuing it still needs to build as 6.0 rather than just using 8.0)

xperiandri (approved):

xperiandri (approved):

Thorium:
> > But how can you build .NET 8 local package, when there is no .NET 8 compatible packages in NuGet?
> 
> How was it done when .NET 6.0 support was first added?
> 

We might need to go with a route where there is "fake-cli 8.0.0-alpha -package" in NuGet to not break any existing fake implementations but still be able to call the global tool.

xperiandri (approved):

Thorium:
When trying to reason the build-log: There are 3 parallel threads all writing to same output, and many of the commands are just starting a process and then monitoring exit-code.

build.fsx
```diff
@@ -658,15 +655,15 @@ Target.create "DotNetCoreIntegrationTests" (fun _ ->
 
     runExpecto
         root
-        "src/test/Fake.Core.IntegrationTests/bin/Release/net6.0/Fake.Core.IntegrationTests.dll"
+        "src" </> "test" </> "Fake.Core.IntegrationTests" </> "bin" </> "Release" </> "net6.0" </> "Fake.Core.IntegrationTests.dll"
```
Numpsy, line 658: Do these need ```()``` round them?

Numpsy:
> When trying to reason the build-log: There are 3 parallel threads all writing to same output, and many of the commands are just starting a process and then monitoring exit-code.

Would it be worthwhlie trying to get something like https://github.com/EnricoMi/publish-unit-test-result-action plugged into the build and publish the test result xml files? Might be easier to see the results vs. trawling through all the text

Thorium:
I like the idea, but maybe separately in master-branch and I'll merge it here then?

Thorium:
It turns out the latest FAKE 6.1.3 builds .NET 8 projects just fine (after this PR https://github.com/fsprojects/FAKE/pull/2835), so I think the priority of this struggle just dropped.

Numpsy:
> I like the idea, but maybe separately in PR to master-branch and I'll merge it here then?

I'll have a look at what it would involve

Numpsy:
> It turns out the latest FAKE 6.1.3 builds .NET 8 projects just fine (after this PR #2835), so I think the priority of this struggle just dropped.

Ok, some test results are getting published now - e.g. https://github.com/fsprojects/FAKE/actions/runs/11105226220/job/30852070808

PR #2821, opened 2024-09-11
Change GitHub.createPullRequest to await the call to PullRequest.Create

### LINKED ISSUE #2820: Is the return type of Fake.Api.GitHub.createPullRequest supposed to be Async<Async<PullRequest>> ?
When looking through build.fsx, I noticed the comment at https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/build.fsx#L1055 about changing to ```GitHub.createPullRequest``` in the future, and thought I'd try making that change to simplify things.

Howevver, when I went to do it I noticed that the return type of ```createPullRequest``` is ```Async<Async<PullRequest>>```, which meant that the sample didn't quite work as the result of calling ```Async.RunSynchronously``` is still ```Async<PullRequest>```.

Ref the [public documentation](https://fake.build/reference/fake-api-github.html#createPullRequest):
image

Is this correct?

Otherwise, I wonder if the async handling at https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/src/app/Fake.Api.GitHub/GitHub.fs#L472 should do ```return! Async.AwaitTask``` or maybe return the ```Async.AwaitTask``` directly without the extra ```async {}``` block?

Numpsy:
- fixes #2820 

### Description

Referencing the comments in #2820, and assuming that the return type is actually wrong, I think it should be awating the ```PullRequest.Create``` so that it gets awaited inside the retry machinery?

Still a draft, as if it's a correct cfhange it needs release notes (and It'd be a breaking change to the public API)

### DIFF
--- a/src/app/Fake.Api.GitHub/GitHub.fs
+++ b/src/app/Fake.Api.GitHub/GitHub.fs
@@ -469,4 +469,4 @@ module GitHub =
     /// </example>
     let createPullRequest owner repoName (pullRequest: NewPullRequest) (client: Async<GitHubClient>) =
         retryWithArg 5 client
-        <| fun client' -> async { return Async.AwaitTask <| client'.PullRequest.Create(owner, repoName, pullRequest) }
+        <| fun client' -> async { return! Async.AwaitTask <| client'.PullRequest.Create(owner, repoName, pullRequest) }


xperiandri:
Indeed, that would be strange

xperiandri:
That would be more appropriate.
However 7.0 can be released if urgently needed

Numpsy:
Ok, I'll leave until the .NET 8 changes are done

xperiandri (approved):

xperiandri:
Should we already merge this?

Numpsy:
Is there going to be a 6.1.2 with any of the other more minor changes, or is just in to version 8 now?

src/app/Fake.Api.GitHub/GitHub.fs
```diff
@@ -469,4 +469,4 @@ module GitHub =
     /// </example>
     let createPullRequest owner repoName (pullRequest: NewPullRequest) (client: Async<GitHubClient>) =
         retryWithArg 5 client
-        <| fun client' -> async { return Async.AwaitTask <| client'.PullRequest.Create(owner, repoName, pullRequest) }
+        <| fun client' -> async { return! Async.AwaitTask <| client'.PullRequest.Create(owner, repoName, pullRequest) }
```
xperiandri, line 472: Do we really need to wrap that into `async` CE?

PR #2824, opened 2024-09-16
Fix a FS3548 'Pattern discard is not allowed for union case that takes no data.' compiler warning

Numpsy:
### Description

```ToolType.FullFramework``` has no payload, so it shouldn't need the discard

### DIFF
--- a/src/app/Fake.DotNet.Cli/CreateProcessExt.fs
+++ b/src/app/Fake.DotNet.Cli/CreateProcessExt.fs
@@ -156,7 +156,7 @@ module CreateProcessDotNetExt =
         /// </example>
         let withToolType (toolType: ToolType) (c: CreateProcess<_>) =
             match toolType with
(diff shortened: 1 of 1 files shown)


Numpsy:
refs

image

xperiandri (approved):

Thorium:
Mac build is still unstable. The message which Mac build fails make no sense to this PR.
It's even hard to reason about what it fails to, 
`no dependencies hello world and casing #2314 failed` so it should do "dotnet fake" for very simple file. Seems to work in my MacBook (old Intel).

Numpsy:
I put some observations about the macOS CI failures in https://github.com/fsprojects/FAKE/issues/2814, but I don't know if it's just a timing issue or something else (it doesn't seem to happen on the Windows or Linux riuns though)

xperiandri (approved):

PR #2825, opened 2024-09-19
[Experiment] Rework the dependencies in fake-cli / Fake.Runtime a little.

Numpsy:
### Description

I'm not sure if this is a sensible thing to try, without knowing the history of it or anything, but - 

I was looking at the failure in the .NET 8.0 update PR at https://github.com/fsprojects/FAKE/actions/runs/10942268784/job/30378864009#step:10:12614 and the issue described at https://github.com/fsprojects/FAKE/issues/2722 and was wondering - Is there actually a need to package the Microsoft.Build assemblies in the fake-cli nuget package - as in, are they actually used by any of the code in fake-cli or are they just pulled in as part of a dependency chain and not actually used?

So, I tried bending things a bit to include a couple of pieces of Fake.DotNet.Cli inline in Fake.Runtime (just the bits for getting global.json and executing 'dotnet --version' and excluding any dependencies on MSBuild) and the build appears to work.

It's not very pretty due to how the code is layed out in a single module, but here it is in acse the idea is of any interest.


src/app/Fake.DotNet.Cli/GlobalJson.fs
```diff
@@ -0,0 +1,60 @@
+﻿module internal GlobalJson
+
+open System.IO
+open Newtonsoft.Json.Linq
```
xperiandri, line 4: I would switch to `System.Text.Json` as the next step in the next PR

xperiandri (approved):

src/app/Fake.DotNet.Cli/GlobalJson.fs
Numpsy: I've had that idea on the thought pile as it'd be nice to use the inbox version of System.Text,Json instead of needing Newtonsoft, though I also need to see how you go about including the STJ NuGet package in the .NET Standard 2.0 build and using the inbox version in the .NET 6.0 build

Numpsy: rebased on top of the System.Text.Json change

src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -4,20 +4,25 @@ namespace Fake.DotNet
 /// .NET Core + CLI tools helpers
 /// </summary>
 [<RequireQualifiedAccess>]
+#if FAKE_INTERNAL_DOTNET_CORE_CLI
+module InternalDotNet =
```
Numpsy, line 8: This was because when I tried making the module internal inside Fake.Runtime I got 
image
so the name was an attempt at avoiding any clashes

xperiandri (approved):

Numpsy:
Is this something to look at in an update, or leave till the next major version?

xperiandri:
Probably the major update as it changes the public API, right?

Numpsy:
I was trying to avoid changing the public API for Fake.DotNet.Cli, but as long as the functions added to Fake.Runtime are public rather than internal that would be an additive change there

PR #2826, opened 2024-09-20
Some maintenance updates to help upcoming .NET 8 migration

Thorium:
Updated expired runtime-identifiers to current version:
https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#known-rids

Added few explicit types to places where .NET8 will have multiple overloads so compiler won't be able to pick correct types.

Addressed the issue detecting .NET-assemblies in the future, pointed out here: https://github.com/fsprojects/FAKE/pull/2818#issuecomment-2356617433

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -425,7 +425,7 @@ docker run --rm -v $PWD:$PWD -w $PWD linuturk/mono-choco $@
     altToolPath
 
 //TODO:: see if we need to update the runtimes to newer versions of OSs
-let runtimes = [ "win7-x86"; "win7-x64"; "osx.10.11-x64"; "linux-x64" ]
+let runtimes = [ "win-x86"; "win-x64"; "osx-x64"; "linux-x64" ]
 
 /// <summary>
 /// Publishes the build artifacts for the given runtime
@@ -809,7 +809,10 @@ runtimes
                     { c with
                         Runtime = Some runtime.Value
                         Configuration = DotNet.Release
-                        OutputPath = Some outDir }
+                        OutputPath = Some outDir 
+                        Framework = Some "net6.0"
+                        // DisableInternalBinLog: https://github.com/fsprojects/FAKE/issues/2722
+                        MSBuildParams = { MSBuild.CliArguments.Create() with DisableInternalBinLog = true }}
                     |> dotnetSimple)
                 proj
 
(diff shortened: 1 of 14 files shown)


Thorium:
Looking the build-logs, it was naive to expect only reference-assemblies have GitHub account under dotnet, because e.g. System.Text.Json can come either from .NET reference assemby or Nuget, and they both have same repository url. I'll try to invent something better.

xperiandri (approved):

xperiandri:
Could you rebase on top of the latest master?

build.fsx
```diff
@@ -876,7 +878,8 @@ Target.create "DotNetCreateNuGetPackage" (fun _ ->
                     { c.MSBuildParams with
                         Properties =
                             [ ("Version", nugetVersion)
-                              ("PackageReleaseNotes", release.Notes |> String.toLines) ] } }
+                              ("PackageReleaseNotes", release.Notes |> String.toLines) ] 
+                        DisableInternalBinLog = true } }
```
xperiandri, line 882: Why do we need this?
Maybe explanatory comment will be relevant here?

src/app/Fake.Runtime/CoreCache.fs
```diff
@@ -332,7 +332,9 @@ let findAndLoadInRuntimeDeps
 
                     let isFramework =
                         assembly.GetCustomAttributes<AssemblyMetadataAttribute>()
-                        |> Seq.exists (fun m -> m.Key = ".NETFrameworkAssembly")
+                        |> Seq.exists (fun m ->
+                            m.Key = ".NETFrameworkAssembly"
+                            || (m.Key = "RepositoryUrl" && m.Value.StartsWith "https://github.com/dotnet/"))
```
xperiandri, line 337: A would add an explanatory comment here too

build.fsx
Thorium: This was due to issues generated of updated MSBuild.StructuredLogger:
- If it's not updated, the run will fail with "Unsupported log file format. Latest supported version is"
- If you user the updated version without it, it says "DotNet.publish fails attempting to load Microsoft.Build.Framework"

But there is even more recent version now, so I'm not sure if it's needed anymore or is the issue fixed.

Thorium: I tested, yes it was needed. Comment added.

Thorium:
Rebased passing build, now failed for ".NET Core SDK install failed with code", maybe a network error.

xperiandri (approved):

Issue #2814, opened 2024-09-01
The continuing intermittent macOS CI build failures


Numpsy:
We seem to still have an intermittent CI build failure on macOS, with errors about not being able to get the start time of a process - 

https://github.com/fsprojects/FAKE/actions/runs/10654117014/job/29530030223#step:8:9395

Looking at the code, it seems to expect to fail to get the process start time sometimes, and responds by logging it and returning DateTime.Now - https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/src/app/Fake.Core.Process/Process.fs#L470

There are sometimes multiple instances of that error in the logs but the only one that seems to make it break is in the ```no dependencies hello world and casing #2314``` test case at https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/src/test/Fake.Core.IntegrationTests/SimpleHelloWorldTests.fs#L79 where it explicitly expects that there is no error output from the test process - https://github.com/fsprojects/FAKE/blob/fac72951389da2798f396e32239fb03a032f36ea/src/test/Fake.Core.IntegrationTests/SimpleHelloWorldTests.fs#L95

I'm not sure why the error occurs (maybe it is just because the test process exits too quickly and is gone by the time the runner tries to get the time), so I 'm not sure if there is a bug or if the test is a bit too strict?
The error only seems to happen on macOS, though there is a different code path for Windows vs. macOS/Linux that might effect things there.

Thorium:
Ok it seems the test is asserting that there is no output. But Fake.Core.Trace.TraceFAKE will be output. Now the question is, if proc.StartTime fails to get, is it just that this platform didn't work and we can ignore, or is it so important error that we should fail the test?

Numpsy:
> Now the question is, if proc.StartTime fails to get, is it just that this platform didn't work and we can ignore, or is it so important error that we should fail the test?
> 

The way it catches the exceptions and returns DateTime.Now suggests an explicit decision to not treat it as an error, but I don't know any history behind the decision

PR #2827, opened 2024-09-22
Mac CI debugging

Thorium:
Testing if adding missing potential exceptions or a little delay would help #2814


xperiandri (approved):

Thorium:
Ok it built, but that might be just pure luck...

xperiandri (approved):

PR #2829, opened 2024-09-22
Set PublishReferencesDocumentationFiles=false in fake-cli

Numpsy:
### Description

Just a thought when looking at the build output for fake-cli - Is there a need to have the XML documentation files for a bunch of the FAKE libraries included in the fake-cli NuGet package?

e.g.
image

### DIFF
--- a/src/app/fake-cli/fake-cli.fsproj
+++ b/src/app/fake-cli/fake-cli.fsproj
@@ -11,6 +11,7 @@
         <PostInstallScript>$([System.IO.File]::ReadAllText('$(MSBuildProjectDirectory)/debian-postinst.sh'))</PostInstallScript>
         <NoWarn>FS3186</NoWarn>
         <ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>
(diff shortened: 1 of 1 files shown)


Thorium:
Maybe they help for API-docs (https://fake.build/reference/index.html) or user tool-tips?

Numpsy:
The documentation XML files in the individual library nuget packages can be used for intellisense and such, and the doc comments themselves are used by the published documentation, but I didn't think the files in the dotnet global tool would be loaded as the package isn't attached to consuming projects like a library is (naybe I'm wrong there though)

Thorium:
Many people still use Fake not via cli-tool but as "dotnet fsi build.fsx" with independent library dependencies, and I think xml-docs there will help.

xperiandri:
@Numpsy  Do you disable XML docs output only for CLI tool?

Numpsy:
> @Numpsy Do you disable XML docs output only for CLI tool?

That was the intent - e.g. in my local build the Fake.DotNet.Cli NuGet package contains the documentation XML file, but the fake-cli package doesn't - 

image

xperiandri:
@Numpsy let's merge?

xperiandri (approved):

PR #2832, opened 2024-09-24
ResolveDotNetRoots to support resoving multiple .NET version paths

Thorium:
There is still an existing issue with `ResolveDotNetRoot` that it fetches the first .NET version from the computer, which might or might not be the intended version to be used. This happens if the folder doesn't have pre-fixed global.json.

This commit adds `ResolveDotNetRoots` which instead returns array of found version.
Then the SdkAssemblyResolver can work better in a computer where there are multiple .NET versions installed.

### DIFF
--- a/src/app/Fake.Runtime/SdkAssemblyResolver.fs
+++ b/src/app/Fake.Runtime/SdkAssemblyResolver.fs
@@ -341,10 +341,11 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
                 releases
                 |> Seq.map (fun release -> release.Runtime.Version.ToString())
                 |> Seq.distinct
+                |> Seq.toList
 
             if this.LogLevel.PrintVerbose then
                 versions
-                |> Seq.iter (fun version -> Trace.trace $"Resolved runtime version: {version}")
+                |> List.iter (fun version -> Trace.trace $"Resolved runtime version: {version}")
 
             versions
 
--- a/src/app/Fake.Runtime/SdkAssemblyResolver.fs
+++ b/src/app/Fake.Runtime/SdkAssemblyResolver.fs
@@ -149,6 +149,89 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
             |> Option.bind resolveFile
             |> Option.map (fun dotnetRoot -> dotnetRoot.Directory.FullName)
 
+    /// <summary>
+    /// provides the path to the `dotnet` binary running this library, respecting various dotnet <see href="https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_root-dotnet_rootx86%5D">environment variables</see>.
+    /// Also probes the PATH and checks the default installation locations
+    /// </summary>
+    member this.ResolveDotNetRoots() =
+        let isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+        let isMac = RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
+        let isLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+        let isUnix = isLinux || isMac
+
+        let potentialDotnetHostEnvVars =
+            [ "DOTNET_HOST_PATH", id // is a full path to dotnet binary
+              "DOTNET_ROOT", (fun s -> Path.Combine(s, this.DotNetBinaryName)) // needs dotnet binary appended
+              "DOTNET_ROOT(x86)", (fun s -> Path.Combine(s, this.DotNetBinaryName)) ] // needs dotnet binary appended
+
+        let existingEnvVarValue envVarValue =
+            match envVarValue with
+            | null
+            | "" -> None
+            | other -> Some other
+
+        let tryFindFromEnvVar () =
+            potentialDotnetHostEnvVars
+            |> List.choose (fun (envVar, transformer) ->
+                match Environment.GetEnvironmentVariable envVar |> existingEnvVarValue with
+                | Some varValue -> Some(transformer varValue |> FileInfo)
+                | None -> None)
+
+        let PATHSeparator = if isUnix then ':' else ';'
+
+        /// Fully resolve the symlink, returning a fully-resolved FileInfo
+        /// that is not a symlink,
+        /// or returning None if the file or any of its resolved targets
+        /// does not exist.
+        let rec resolveFile (fi: System.IO.FileInfo) : FileInfo option =
+            if not fi.Exists then None
+            elif isNull fi.LinkTarget then Some fi
+            else resolveFile (System.IO.FileInfo fi.LinkTarget)
+
+        let tryFindFromPATH () =
+            Environment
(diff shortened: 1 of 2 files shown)


src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -149,6 +149,89 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
+    member this.ResolveDotNetRoots() =
+        let isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+        let isMac = RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
+        let isLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+        let isUnix = isLinux || isMac
+
+        let potentialDotnetHostEnvVars =
+            [ "DOTNET_HOST_PATH", id // is a full path to dotnet binary
+              "DOTNET_ROOT", (fun s -> Path.Combine(s, this.DotNetBinaryName)) // needs dotnet binary appended
+              "DOTNET_ROOT(x86)", (fun s -> Path.Combine(s, this.DotNetBinaryName)) ] // needs dotnet binary appended
```
xperiandri, line 165:
```suggestion
            [ struct "DOTNET_HOST_PATH", id // is a full path to dotnet binary
              struct "DOTNET_ROOT", (fun s -> Path.Combine(s, this.DotNetBinaryName)) // needs dotnet binary appended
              struct "DOTNET_ROOT(x86)", (fun s -> Path.Combine(s, this.DotNetBinaryName)) ] // needs dotnet binary appended
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -149,6 +149,89 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
+            | null
+            | "" -> None
+            | other -> Some other
+
+        let tryFindFromEnvVar () =
+            potentialDotnetHostEnvVars
+            |> List.choose (fun (envVar, transformer) ->
+                match Environment.GetEnvironmentVariable envVar |> existingEnvVarValue with
+                | Some varValue -> Some(transformer varValue |> FileInfo)
+                | None -> None)
```
xperiandri, line 178:
```suggestion
        let existingEnvVarValue envVarValue =
            match envVarValue with
            | null
            | "" -> ValueNone
            | other -> ValueSome other

        let tryFindFromEnvVar () =
            potentialDotnetHostEnvVars
            |> List.choose (fun struct (envVar, transformer) ->
                Environment.GetEnvironmentVariable envVar
                |> existingEnvVarValue
                |> ValueOption.map (transformer >> FileInfo)
                |> ValueOption.toOption
           )
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -149,6 +149,89 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
+            else resolveFile (System.IO.FileInfo fi.LinkTarget)
+
+        let tryFindFromPATH () =
+            Environment
+                .GetEnvironmentVariable("PATH")
+                .Split(PATHSeparator, StringSplitOptions.RemoveEmptyEntries ||| StringSplitOptions.TrimEntries)
+            |> Array.choose (fun d ->
+                let fi = Path.Combine(d, this.DotNetBinaryName) |> FileInfo
+
+                if fi.Exists then Some fi else None)
```
xperiandri, line 198:
```suggestion
            |> Seq.map (fun d -> Path.Combine(d, this.DotNetBinaryName))
            |> Seq.map FileInfo
            |> Seq.where _.Exists
            |> Seq.toArray
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -217,24 +300,47 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
-        let sdkVersion = DotNet.getVersion versionOptions |> ReleaseVersion
+        let dotnetRoots = this.ResolveDotNetRoots()
+
+        let sdkVersions =
+            if Array.isEmpty dotnetRoots then
+                [ DotNet.getVersion (versionOptions None) |> ReleaseVersion ]
+            else
+                dotnetRoots
+                |> Array.map (fun dotnetRoot -> DotNet.getVersion (versionOptions (Some dotnetRoot)) |> ReleaseVersion)
+                |> Array.toList
```
xperiandri, line 320:
```suggestion
                |> Seq.map (fun dotnetRoot -> DotNet.getVersion (versionOptions (Some dotnetRoot)) |> ReleaseVersion)
                |> Seq.toList
```

src/app/Fake.Runtime/SdkAssemblyResolver.fs
```diff
@@ -217,24 +300,47 @@ type SdkAssemblyResolver(logLevel: Trace.VerboseLevel) =
...
+
             failwithf
-                $"Could not find a suitable .NET 6 runtime version matching SDK version: {sdkVersion.ToString()} (You can also try setting environment variable FAKE_SDK_RESOLVER_CUSTOM_DOTNET_VERSION to e.g. {sdkVersion.Major.ToString()}.0 )"
+                $"Could not find a suitable .NET 6 runtime version matching SDK version: {versions} (You can also try setting environment variable FAKE_SDK_RESOLVER_CUSTOM_DOTNET_VERSION to e.g. {example}.0 )"
         | releases ->
             let versions =
-                releases |> List.map (fun release -> release.Runtime.Version.ToString())
+                releases
+                |> List.map (fun release -> release.Runtime.Version.ToString())
+                |> List.distinct
```
xperiandri, line 343:
```suggestion
                |> Seq.map (fun release -> release.Runtime.Version.ToString())
                |> Seq.distinct
```

xperiandri (approved):
I would suggest a few changes but up to you

src/app/Fake.Runtime/SdkAssemblyResolver.fs
Thorium: This is a Tuple<string, Func<_,_>>, I think it's better to keep as Tuple...

src/app/Fake.Runtime/SdkAssemblyResolver.fs
Thorium: Makes sense, this was just copied from the old code, but I'll update both.

Thorium: We loose the whole benefit of ValueTuple vs Tuple if we still have to assign both via `ValueOption.toOption` so I think this change will make sense if List.choose and List.tryPick etc will support natively ValueTuple, but if not, there is no point, just extra-mapping of types.

src/app/Fake.Runtime/SdkAssemblyResolver.fs
Thorium: choose is preferred over map->map->where exist, by https://fsprojects.github.io/FSharpLint/

PR #2837, opened 2024-09-29
Publish unit test results for pull request builds

Numpsy:
Based on comments in #2818 - publish the unit test results files during CI builds, and then publish a test report using https://github.com/EnricoMi/publish-unit-test-result-action.

This first attempt is based on the action documentation, and on some implementations in other repositories (This is the first time I've used it myself)

Note: It appears that Github won't pick up the new publish-test-results action until the yaml file for it is in the master branch, so I've had to merge a test change into the master branch in my fork to test it - you can see the results for that at https://github.com/Numpsy/FAKE/runs/30818769728 which looks reasonable.

Not sure if it'll need tuning to control where the test results get pushed to - e.g. if they get displayed inline inside pull requests as well as in the build results page.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -393,8 +393,10 @@ let callPaket wd args =
 /// <param name="dllPath">Test assembly to run tests from</param>
 /// <param name="resultsXml">Expecto test results XML file</param>
 let runExpecto workDir dllPath resultsXml =
+    let resultsFile = "testresults" </> resultsXml
+
     let processResult =
-        DotNet.exec (dotnetWorkingDir workDir) (sprintf "%s" dllPath) (sprintf "--nunit-summary %s" resultsXml)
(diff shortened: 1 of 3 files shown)


build.fsx
```diff
@@ -393,8 +393,10 @@ let callPaket wd args =
 /// <param name="dllPath">Test assembly to run tests from</param>
 /// <param name="resultsXml">Expecto test results XML file</param>
 let runExpecto workDir dllPath resultsXml =
+    let resultsFile = "testresults" </> resultsXml
```
Numpsy, line 396: As it stands, the test result xml files get written into the root directory of the repository, but it's easier to upload them in a batch if they all get written into a subdirectory and this looked like the easiest place to change to do that

xperiandri (approved):

PR #2840, opened 2024-10-07
Environment variable documentation update

Thorium:
After struggling a bit with environment variables, docs update.
This PR hopefully saves some time in future.

### DIFF
--- a/docs/guide/core-targets.md
+++ b/docs/guide/core-targets.md
@@ -89,8 +89,9 @@ Now we have the following options:
 
 ## Script with arguments
 
-If you don't do `Target.initEnvironment()` at all, then FAKE will only use environment variables already set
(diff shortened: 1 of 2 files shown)


docs/guide/core-targets.md
```diff
@@ -89,6 +89,11 @@ Now we have the following options:
 
 ## Script with arguments
 
+If you don't do `Target.initEnvironment()` at all, then FAKE will only use environment variables already set
+(e.g. Windows: `set mything=123`) instead of parsing the command-line `-e`.
```
xperiandri, line 93:
```suggestion
If you don't call `Target.initEnvironment()` before you use the `Environment` module, 
then FAKE will only use system and inline environment variables 
(e.g. Windows: `set mything=123`) instead of parsing the command line `-e` params.
```

xperiandri (approved):

PR #2843, opened 2024-10-20
Discover VSTest and MSTest using VSWhere

farlee2121:
Fake.DotNet.Testing.VSTest and Fake.DotNet.Testing.MSTest currently search for VSTest and MSTest respectively using hard-coded lists of possible paths. These paths are simply different versions of Visual Studio.

@robertpi previously did work to update these paths and partially move them to a VSWhere for locating Visual Studio installs.
The work was not merged. #2542 

This PR rebases #2542 and further migrates VSTest discovery to use VSWhere.

This should reduce effort to keep Fake.DotNet.Testing.VSTest and Fake.DotNet.Testing.MSTest up to date as new versions of Visual Studio are released.

This PR will also fix the current issue where Fake.DotNet.Testing.VSTest cannot automatically discover VSTest with recent versions of Visual Studio.

Note that VSWhere [only works for VisualStudio 2017+](https://github.com/vbfox/FoxSharp/tree/master/src/BlackFox.VsWhere). We'll want to re-add hard-coded pre-2017 paths if we want to continue supporting those older versions of Visual Studio

@xperiandri

### DIFF
--- a/src/app/Fake.DotNet.Testing.MSTest/MSTest.fs
+++ b/src/app/Fake.DotNet.Testing.MSTest/MSTest.fs
@@ -8,20 +8,14 @@ module MSTest =
 
     open System
     open System.Text
+    open BlackFox.VsWhere
     open Fake.Core
+    open Fake.IO
     open Fake.Testing.Common
-
-    let internal mstestPaths =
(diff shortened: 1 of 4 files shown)


xperiandri (approved):

xperiandri:
@farlee2121 cool! Should I merge?

PR #2845, opened 2024-12-24
Fix the test result output path for TemplateIntegrationTests

Numpsy:
### Description

I noticed when having a poke at the .NET 8 CI test failures that the 'Fake.DotNet.Cli.IntegrationTests' test results file isn't being published into the CI as the restults file is getting put under the integration tests folder rather than under the root results folder with the other 3 result xml files - so this is an attempt to fix that by using the same root results folder as the other test project runs

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -664,11 +664,10 @@ Target.create "DotNetCoreIntegrationTests" (fun _ ->
         "Fake_Core_IntegrationTests.TestResults.xml")
 
 Target.create "TemplateIntegrationTests" (fun _ ->
-    let targetDir = srcDir </> "test" </> "Fake.DotNet.Cli.IntegrationTests"
 
     runExpecto
-        targetDir
-        "bin/Release/net6.0/Fake.DotNet.Cli.IntegrationTests.dll"
+        root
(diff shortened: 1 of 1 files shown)


build.fsx
```diff
@@ -664,11 +664,10 @@ Target.create "DotNetCoreIntegrationTests" (fun _ ->
         "Fake_Core_IntegrationTests.TestResults.xml")
 
 Target.create "TemplateIntegrationTests" (fun _ ->
-    let targetDir = srcDir </> "test" </> "Fake.DotNet.Cli.IntegrationTests"
 
     runExpecto
-        targetDir
-        "bin/Release/net6.0/Fake.DotNet.Cli.IntegrationTests.dll"
+        root
+        ("src" </> "test" </> "Fake.DotNet.Cli.IntegrationTests" </> "bin" </> "Release" </> "net6.0" </> "Fake.DotNet.Cli.IntegrationTests.dll")
```
Numpsy, line 670: I pulled the change out of a version of the .NET 8 branch, so it picked up the change to avoid hard coded path separators as well

xperiandri (approved):

xperiandri:
@Numpsy should any previous fixes or this be documented in the release notes?

Numpsy:
@xperiandri https://github.com/fsprojects/FAKE/pull/2839 maybe? (this one is just in the tests)

xperiandri:
Yes, not very important for consumers but makes sense to mention

xperiandri (approved):

PR #2846, opened 2024-12-24
Fix sprintf format in TemplateTests::getDebuggingInfo

Numpsy:
When looking at the remaining test failures when running on .NET 8 I noticed that some of the test error messages contained

...Debugging Info: Fake.DotNet.Cli.IntegrationTests.TemplateTests+getDebuggingInfo@58-2 Result: ...

and I think that '+getDebuggingInfo' is because the sprintf in getDebuggingInfo contains 3 %s but only two parameters, so it's actually being treated as a partially applied function by the string interpolcation in isProcessSucceeded rather than a string?

Looks like it's been like that for some time - not sure what the original intent was

### DIFF
--- a/src/test/Fake.DotNet.Cli.IntegrationTests/TemplateTests.fs
+++ b/src/test/Fake.DotNet.Cli.IntegrationTests/TemplateTests.fs
@@ -56,7 +56,7 @@ let inline redirect () =
 
 let getDebuggingInfo () =
     sprintf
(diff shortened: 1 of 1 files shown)


Numpsy:
Example of the existing test output when an error occurs: https://github.com/fsprojects/FAKE/actions/runs/11109335183/job/30864439817?pr=2818#step:9:8381

xperiandri (approved):

PR #2853, opened 2025-01-12
Update the `dotnet-install` script URL in `Fake.DotNet.getGenericDotNetCliInstallerUrl`

Numpsy:
refs https://github.com/fsprojects/FAKE/issues/2848

### Description

The scripts have been updated due to https://github.com/dotnet/core/issues/9671, but according to https://github.com/dotnet/sdk/blob/main/scripts/obtain/dotnet-install-readme the source for the scripts was changed from https://github.com/dotnet/sdk to https://github.com/dotnet/install-scripts some years ago, so it looks like FAKE was already using an old version of the scripts?

### DIFF
--- a/src/app/Fake.DotNet.Cli/DotNet.fs
+++ b/src/app/Fake.DotNet.Cli/DotNet.fs
@@ -136,7 +136,7 @@ module DotNet =
     /// Get .NET Core SDK download uri
     /// </summary>
     let private getGenericDotNetCliInstallerUrl branch installerName =
-        sprintf "https://raw.githubusercontent.com/dotnet/cli/%s/scripts/obtain/%s" branch installerName
+        sprintf "https://raw.githubusercontent.com/dotnet/install-scripts/%s/src/%s" branch installerName
 
     let private getPowershellDotNetCliInstallerUrl branch =
         getGenericDotNetCliInstallerUrl branch "dotnet-install.ps1"
@@ -188,7 +188,7 @@ module DotNet =
         /// Parameter default values.
         static member Default =
             { AlwaysDownload = false
-              Branch = "master"
+              Branch = "main"
               CustomDownloadDir = None }
 
     /// <summary>


src/app/Fake.DotNet.Cli/DotNet.fs
```diff
@@ -136,7 +136,7 @@ module DotNet =
     /// Get .NET Core SDK download uri
     /// </summary>
     let private getGenericDotNetCliInstallerUrl branch installerName =
-        sprintf "https://raw.githubusercontent.com/dotnet/cli/%s/scripts/obtain/%s" branch installerName
+        sprintf "https://raw.githubusercontent.com/dotnet/install-scripts/%s/src/%s" branch installerName
```
Numpsy, line 139: I'm not really certain if this should be pulling the scripts from github or the 'distribution' urls such as https://dot.net/v1/dotnet-install.sh - but thats a bigger change which would mean dropping the configurable branch, so I just tried to keep it minimal for now

richlander: We will be sharing guidance shortly.

This issue should provide some insight: https://github.com/dotnet/install-scripts/issues/558#issuecomment-2575550277

xperiandri:
@Numpsy should I merge?

Numpsy:
> @Numpsy should I merge?

I don't think that the existing URL is right (the script there doesn't contain the distribution domain update), but it might need a call on if 'branch' is really useful or if it should always use ```https://dot.net/v1/``` in future major versions

richlander:
We don't guarantee prod-level stability/response with the GitHub URL.

xperiandri:
Should we already merge this?

Numpsy:
I thought that Microsoft said they were going to turn off the old domains, but that doesn't seem to have happened, so I'm not sure what the current situation is there.

PR #2872, opened 2025-10-19
Update the NuGet package to 6.12.4

Numpsy:
refs https://github.com/fsprojects/FAKE/issues/2871

This will hopefully update various transitive dependencies to newer versions to pick up security related fixes.


Numpsy:
I set if to 6.12 instead of the latest because that still has a .NET 5.0 TFM that can be consumed by the .NET 6 build here, where newer versions have upped that to .NET 8.0, meaning we'd get pushed over to the .NET Standard 2.0 build.
Maybe that's not a big issue, but it's a bigger change, so (one for if the .NET 8 upgrade here gets finished at any point)

PR #2880, opened 2025-10-28
Async API for Process

Thorium:
FAKE is used by many .NET ecosystems via not only .NET application or exe-file, but also via libraries.
Thus in some cases it would be more beneficial to expose more the Async API instead of FAKE do Async.RunSynchronously everywhere (which is OK in console application), because big enterprise build scripts could maybe benefit of this asynchrony.

So this PR is a start, to expose Process.fs methods also by Async. But for backward compatibility the old methods should also be kept.


xperiandri (approved):

Numpsy:
Looks like it needs fantomas running on the changes?

PR #2882, opened 2025-11-01
Try using File.ResolveLinkTarget instead of Mono.Unix.UnixPath.GetRea…

### LINKED ISSUE #2790: FAKE.DotNet.Cli / Mono.Posix / Symlinks
A thought (as yet not investigated in any more detail than this)

Fake currently has a dependency on Mono.Posix.NetStandard to get the 'GetRealPath' function to resolve symlinks:
https://github.com/fsprojects/FAKE/blob/accd2f4dfcaafd33fb4d965fbf84dc4b9c51cdd7/src/app/Fake.DotNet.Cli/DotNet.fs#L839

That caused a problem on ARM CPU macs due to the lack of an ARM build of the native code part of that package.

I tried updating it to the newer Mono.Unix package in #2781  back that failed with other issues.

So, I was reading about a little for ideas and noticed that .NET 6.0 now has a built in 'ResolveLinkTarget' function that does this sort of thing with no external dependencies:
https://learn.microsoft.com/en-us/dotnet/api/system.io.file.resolvelinktarget?view=net-6.0

So - I wonder if it be possible to use that for .NET 6.0 builds and avoid the Mono.Unix dependency?

Numpsy:
…lPath on .NET (Core)

### Description

fixes #2790

This is an attempt at using the built in ```File.ResolveLinkTarget``` in the .NET 6 builds, rather than relying on Mono.Unix.
The function isn't in .NET Standard, so the buitl for that still needs to use the external library.

WIP because it might need some more exception handling.

### DIFF
--- a/src/app/Fake.DotNet.Cli/DotNet.fs
+++ b/src/app/Fake.DotNet.Cli/DotNet.fs
@@ -831,12 +831,32 @@ module DotNet =
 
         let currentDotNetRoot = Environment.environVar "DOTNET_ROOT"
 
+        // resolve potential symbolic link to the real location
+        // On .NET we can use File.ResolveLinkTarget,  on .NET Standard we use Mono.Unix.UnixPath.GetRealPath
+        let resolveLinkTarget linkPath =
+
+            let resolvedPath =
+#if NET
+                // ResolveLinkTarget returns null if the input isn't a link. Just use the input path directly in that case.
+                // @@TODO@@ ResolveLinkTarget might throw if the input doesn't exist - maybe need to handle that.
+                let resolvedTarget = File.ResolveLinkTarget(linkPath, true)
+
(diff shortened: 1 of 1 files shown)


Numpsy:
I'm not sure what the history of the ```FX_NO_POSIX``` define is.

It's defined in the project file at https://github.com/fsprojects/FAKE/blob/74e1e6f000f48754a69eb52f6fcb02e5e086061d/src/app/Fake.DotNet.Cli/Fake.DotNet.Cli.fsproj#L16 but only if the TFM is ```net472```, and the project isn't built as ```net472```, only ```net6.0;netstandard2.0```

Numpsy:
The .NET 6 project will also still reference the Mono.Posix.NETStandard NuGet package, not sure how to conditionalize that in Paket

PR #2884, opened 2025-11-30
Async API for Process (+ Fantomas)

Numpsy:
This just #2880 but with an extra commit to run Fantomas


xperiandri:
Do we need those wrappers that perform sync operations?
I think it should be a caller's responsibility to handle that

Numpsy:
This is just https://github.com/fsprojects/FAKE/pull/2880 but run through Fantomas - should discussions about the idea be over there?

Numpsy:
@Thorium Any comments on the machines comments as these were your changes?

PR #2885, opened 2025-11-30
Fix doc comments in Process.fs

Numpsy:
The Exec functions have a parameter called 'dir' but the doc comments have 'directory', so this just makes them consistent

(Just noticed a warning in Rider when looking at #2880 and though they should be consistent)

### DIFF
--- a/src/app/Fake.Core.Process/Process.fs
+++ b/src/app/Fake.Core.Process/Process.fs
@@ -927,12 +927,12 @@ type Shell private () =
           Args = [] }
 
     /// <summary>
-    /// Runs the given process, waits for it's completion and returns the exit code.
+    /// Runs the given process, waits for its completion and returns the exit code.
     /// </summary>
     ///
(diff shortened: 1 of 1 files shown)


xperiandri:
Are there any relevant comments from Copilot?

Numpsy:
I wasn't looking at anything other than the param names before, but looking at the code now the 'async' function doesn't look very async? (actually it looks the same as the implementation of the not-async version)

xperiandri:
Yep so I think we should get rid of that

Numpsy:
> Yep so I think we should get rid of that

It doesn't seem to be used from anywhere in this repo at least:

https://github.com/Numpsy/FAKE/commit/4c91022cb5912a14a26768bdcfc63f619e813969
https://github.com/Numpsy/FAKE/actions/runs/25744391229

So remove in v8 maybe

xperiandri (approved):

PR #2852, opened 2025-01-09
Further attempts at building with .NET 8

Numpsy:
### Description

As https://github.com/fsprojects/FAKE/pull/2818, but updated on top of the latest changes and with a couple of extra commits to try to get more tests passing - see further comments for details

PR #2886, opened 2025-11-30
Add a direct reference to System.Configuration.ConfigurationManager v…

Numpsy:
…6.0.0 to Fake.DotNet.MSBuild

refs https://github.com/fsprojects/FAKE/issues/2871

This attempts to remove a transient reference to old versions of System.Drawing.Common with known CVEs by directly referencing a newer version of ConfigurationManager

fake-cli is already using v6 (it's included in the fake-cli tool nuget package) which hopefully means it's already had some testing with that version

### DIFF
--- a/src/app/Fake.DotNet.MSBuild/paket.references
+++ b/src/app/Fake.DotNet.MSBuild/paket.references
@@ -4,3 +4,4 @@ FSharp.Core
 NETStandard.Library
 MSBuild.StructuredLogger
 BlackFox.VsWhere
+System.Configuration.ConfigurationManager


Numpsy:
This shouldn't be need after the .NET 8 update as all the transient dependencies are using newer versions anyway, so it's a question of if it's useful for a 6.1.5 before that.

src/app/Fake.DotNet.MSBuild/paket.references
```diff
@@ -4,3 +4,4 @@ FSharp.Core
 NETStandard.Library
 MSBuild.StructuredLogger
 BlackFox.VsWhere
+System.Configuration.ConfigurationManager
```
xperiandri, line 7: I would prefer some order like `System, FSharp, Microsoft, Other`
```suggestion
System.Configuration.ConfigurationManager
FSharp.Core
NETStandard.Library
MSBuild.StructuredLogger
BlackFox.VsWhere
```

xperiandri (changes requested):

Numpsy:
Anyway, this won't be needed if #2852 is finished, so do we want it, or not?

